PR #37, opened 2020-07-16
add pseudo random functions

### LINKED ISSUE #36: add pseudo random functions
the `Random` component currently implements cryptographically secure random functions, these should be moved to the `SecureRandom` namespace, while adding pseudo-random functions under `PseudoRandom` namespace.

i.e

```php

function float(): float {
  return (float)(namespace\int(0, Math\INT53_MAX - 1) / Math\INT53_MAX);
}

function int(int $min = Math\INT53_MIN, int $max = Math\INT53_MAX): int {
  return mt_rand($min, $max);
}

```

hasseneb:
closes #36


src/Psl/Arr/random.php
```diff
@@ -33,5 +33,5 @@ function random(array $values)
     }
 
     /** @psalm-suppress MissingThrowsDocblock */
-    return at($shuffled, Random\int(0, $size - 1));
+    return at($shuffled, SecureRandom\int(0, $size - 1));
```
azjezz: we should use pseudo-random here.

src/Psl/PseudoRandom/float.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Psl\PseudoRandom;
+
+use Psl;
+use Psl\Math;
+
+/**
+ * Returns a cryptographically secure random float in the range from 0.0 to 1.0.
+ *
+ * @throws Psl\Exception\RuntimeException If it was not possible to gather sufficient entropy.
```
azjezz: `PseudoRandom\float`, and `PseudoRandom\int` don't return cryptographically secure randoms, and the exception is never thrown.

azjezz (commented):
LGTM 👍 just a couple of comments.

azjezz (dismissed):

PR #38, opened 2020-07-16
remove type check command

hasseneb:


azjezz (approved):

azjezz:
P.S: @muglug psalter currently breaks PSL by removing things that it shouldn't. i will try to file an issue :)

PR #48, opened 2020-08-18
[Env] initial commit

azjezz:


azjezz:
This is a direct PHP re-implementation of Rusts std::env ( https://doc.rust-lang.org/std/env/index.html ), which probably needs a revision to match the naming used in PSL.

PR #53, opened 2020-09-26
[DataStructure] Introduce the DataStructure component

azjezz:
TODO:
- [x] Queue
- [x] PriorityQueue
- [x] Stack

### DIFF
--- a/src/Psl/DataStructure/PriorityQueue.php
+++ b/src/Psl/DataStructure/PriorityQueue.php
@@ -1,0 +1,139 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\DataStructure;
+
+use Psl;
+use Psl\Arr;
+use Psl\Math;
+
+/**
+ * @psalm-template T
+ *
+ * @implements PriorityQueueInterface<T>
+ */
+final class PriorityQueue implements PriorityQueueInterface
+{
+    /**
+     * @psalm-var array<int, list<T>>
+     */
+    private array $queue = [];
+
+    /**
+     * Adds a node to the queue.
+     *
+     * @psalm-param T $node
+     */
+    public function enqueue($node, int $priority = 0): void
+    {
+        $nodes = $this->queue[$priority] ?? [];
+        $nodes[] = $node;
+
+        $this->queue[$priority] = $nodes;
+    }
+
+    /**
+     * Retrieves, but does not remove, the node at the head of this queue,
+     * or returns null if this queue is empty.
+     *
+     * @psalm-return null|T
+     */
+    public function peek()
+    {
+        if (0 === $this->count()) {
+            return null;
+        }
+
+        // Retrieve the highest priority.
+        $priority = Math\max(Arr\keys($this->queue)) ?? 0;
+
+        // Retrieve the list of nodes with the priority `$priority`.
+        $nodes = Arr\idx($this->queue, $priority, []);
+
+        // Retrieve the first node of the list.
+        return Arr\first($nodes);
+    }
+
+    /**
+     * Retrieves and removes the node at the head of this queue,
+     * or returns null if this queue is empty.
+     *
+     * @psalm-return null|T
+     */
+    public function pull()
+    {
+        if (0 === $this->count()) {
+            return null;
+        }
+
+        /** @psalm-suppress MissingThrowsDocblock - the queue is not empty */
+        return $this->dequeue();
+    }
+
+    /**
+     * Dequeues a node from the queue.
+     *
+     * @psalm-return T
+     *
+     * @throws Psl\Exception\InvariantViolationException If the Queue is invalid.
+     */
+    public function dequeue()
+    {
+        Psl\invariant(0 !== $this->count(), 'Cannot dequeue a node from an empty Queue.');
+
+        /**
+         * Peeking into a non-empty queue always results in a value.
+         *
+         * @psalm-var T $node
+         */
+        $node = $this->peek();
+
+        $this->drop();
+
+        return $node;
+    }
+
+    private function drop(): void
+    {
+        /**
+         * Retrieve the highest priority.
+         *
+         * @var int $priority
+         */
+        $priority = Math\max(Arr\keys($this->queue));
+
+        /**
+         * Retrieve the list of nodes with the priority `$priority`.
+         *
+         * @psalm-suppress MissingThrowsDocblock
+         */
+        $nodes = Arr\at($this->queue, $priority);
+
+        // If the list contained only this node,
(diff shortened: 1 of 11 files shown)


src/Psl/DataStructure/PriorityQueue.php
```diff
@@ -0,0 +1,137 @@
...
+     *
+     * @psalm-param T $node
+     */
+    public function enqueue($node, int $priority = 0): void
+    {
+        $nodes = $this->queue[$priority] ?? [];
+        $nodes[] = $node;
+
+        $this->queue[$priority] = $nodes;
+        $this->queue = Arr\filter(
```
veewee: Not sure why you are filtering the queue in here?
If you add an item, you don't add empty items - so the filter seems a bit out of place.

src/Psl/DataStructure/PriorityQueue.php
```diff
@@ -0,0 +1,137 @@
...
+    {
+        return $this->fetch(true);
+    }
+
+    /**
+     * @psalm-return T
+     *
+     * @throws Psl\Exception\InvariantViolationException If the Queue is invalid.
+     */
+    private function fetch(bool $remove)
```
veewee: maybe better to create a `fetch` function that fetches and a `remove` that removes?
Now you're doing 2 things in 1 function.

src/Psl/DataStructure/QueueInterface.php
```diff
@@ -0,0 +1,55 @@
...
+use Countable;
+
+/**
+ * An interface representing a queue data structure ( FIFO ).
+ *
+ * @template T
+ *
+ * @see https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics)
+ */
+interface QueueInterface extends Countable
```
veewee, line 17: How would you use the queue?
I assume in most cases you want to iterate over the queue.
However, this leaves you with this implementation:

```php
while ($res = $queue->pull() !== null) {
}
```

I find this quite awful :)
Can we make those classes `Iterable`? 
That way you can also use the `Iter\` functions on a queue.


(same remarks for all those interfaces)

src/Psl/DataStructure/QueueInterface.php
```diff
@@ -0,0 +1,55 @@
...
+     */
+    public function peek();
+
+    /**
+     * Retrieves and removes the node at the head of this queue,
+     * or returns null if this queue is empty.
+     *
+     * @psalm-return null|T
+     */
+    public function pull();
```
veewee, line 40: These can return `null|T` - but if you queue `null`, the queue won't really be empty.

veewee (commented):
Looks good! Added some remarks to make it even more awesome!

src/Psl/DataStructure/QueueInterface.php
azjezz: In that case `T` is nullable, assuming `T` is `null|string`, the return type is `null|(null|string)`.

so it is expected to return null if the head of the queue does contain null as a value.

src/Psl/DataStructure/QueueInterface.php
azjezz: I thought about implementing iterator or iterator aggregate, but not sure which would be better?

implementing iterator means that the queue might change within the loop, resulting probably in a corrupted queue, implementing iterator aggregate, means values enqueued within the loop, won't be iterated over.

which one do you think makes more sense?

src/Psl/DataStructure/QueueInterface.php
veewee: Yeah in that case the `dequeue()` method is better, but using that function is a bit more cumbersome since it throws exceptions.

src/Psl/DataStructure/PriorityQueue.php
azjezz: Ah, yep, I missed this before implementing the filter at the `fetch` method :D 

thanks.

src/Psl/DataStructure/PriorityQueue.php
azjezz: makes sense 👍🏼

src/Psl/DataStructure/QueueInterface.php
azjezz: that's why they both exist. you can use whatever behavior you prefer, null or an exception.

personally i prefer `pull`, unless working with nullable values ( in which case i wouldn't know if the queue is empty or if its just a null value, unless i call count() )

azjezz:
ugh, i did `git reset` instead of `git rebase` 🤦🏼‍♂️

src/Psl/DataStructure/QueueInterface.php
veewee: That is a good question. Did not think about that tbh ...
Maybe you can think about the use-cases : would it make sense to be able to queue something whilst consuming the items inside the queue?
For example the stack : if you add something, it will be picked up in the next iteration.


You could also take a look how the SPL variants work and determine if it makes sense:

- https://www.php.net/manual/en/class.splstack.php
- https://www.php.net/manual/en/class.splpriorityqueue.php
- https://www.php.net/manual/en/class.splqueue.php


I do like the idea of being able to add items while iterating, but I don't know if it will be used in the real world.
Places where I used priority queues was e.g. inside event listeners - which are configured before consumed.
That specific implementation did not clear the queue whilst iterating.

On the other hand : the SPL provides `Iterator` and not `IteratorAggregate`.

azjezz: I can't think of a case where I had to push items into a queue while iterating over it either. and I think generally enquing and consuming should be done seperatly, therefor we should implement `IteratorAggregate`

azjezz: Ah, another thing, should we remove items for the queue when iteratoring over them?

azjezz: currently, SPL doesn't ... while consuming a queue or a stack, i would assume that items are removed ... this is quite weird.

https://3v4l.org/WL8os

azjezz: WDYT about adding `empty(): bool` instead? 🤔 

I think when iterating over a  queue, people will have differente expectations ( can I enqueue while consuming? will i be consuming the nodes i pushed? will they be removed from the queue? )

while the following is pretty clear:

```php
while(!$queue->empty()) {
  $node = $queue->dequeue();

  // use the node. 
  if (something) {
    $queue->enqueue(something); // will be consumed in this loop
  }
}
```

azjezz: let's merge as is now, and continue discussing this in #83 for later.

PR #56, opened 2020-09-27
[Encoding] introduce encoding component

azjezz:
- [x] base 64
- [x] base 16 ( hex )

--- 

Changelog:

Features
- added `Psl\Encoding\Base64\encode`
- added `Psl\Encoding\Base64\decode`
- added `Psl\Encoding\Hex\encode`
- added `Psl\Encoding\Hex\decode`

### DIFF
--- a/src/Psl/Encoding/Base64/decode.php
+++ b/src/Psl/Encoding/Base64/decode.php
@@ -1,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Encoding\Base64;
+
+use Psl\Encoding\Exception;
+use Psl\Str;
+use Psl\Type;
+
+use function base64_decode;
+use function preg_match;
+
+/**
+ * Decode a base64-encoded string into raw binary.
+ *
+ * Base64 character set:
+ *  [A-Z]      [a-z]      [0-9]      +     /
+ *  0x41-0x5a, 0x61-0x7a, 0x30-0x39, 0x2b, 0x2f
+ *
+ * @psalm-pure
+ *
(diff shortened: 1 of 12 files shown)


src/Psl/Encoding/Base64/decode.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+function decode(string $base64): string
+{
+    if (!preg_match('%^[a-zA-Z0-9/+]*={0,2}$%', $base64)) {
+        throw new Exception\RangeException(
+            'The given base64 string contains characters outside the base64 range.'
+        );
+    }
+
+    return (string) @base64_decode($base64, true);
```
azjezz: 🤔

veewee: When does it trigger a warning (or something similar)?
I've tried with some utf8 chars, but only got `false` due to invalid range.

azjezz: incorrect padding, but not sure, I would need to add more tests I guess.

azjezz: hm, actually, it returns false without any warning for invalid padding. so I guess we just check the return type, if it's false, we throw `Exception\PaddingException`.

azjezz: or we can check the length before calling base64_decode, as base64 encoded strings length is a multiple of 4 when padded correctly.

veewee: Probably best to do both? Check the length + the return type

PR #57, opened 2020-09-28
[Fun] Add pipe and after combinators

veewee:
- [x] after
- [x] pipe

### DIFF
--- a/src/Psl/Fun/after.php
+++ b/src/Psl/Fun/after.php
@@ -37,5 +37,5 @@ namespace Psl\Fun;
  */
 function after(callable $first, callable $next): callable
 {
-    return fn ($input) => $next($first($input));
+    return static fn ($input) => $next($first($input));
 }
--- a/src/Psl/Fun/after.php
+++ b/src/Psl/Fun/after.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Fun;
+
+/**
+ * Returns a function that calls the next function with the result of the first function.
+ *
+ * Example:
+ *
+ *      $runExpression = Fun\after(
+ *          fn($i) => $i +  1,
+ *          fn(int $i) => $i * 5
+ *      );
+ *      => $runExpression(1) === 10;
+ *      => $runExpression(2) === 15;
+ *      => $runExpression(3) === 20;
+ *
+ *      $greet = Fun\after(
+ *          fn($value) => 'Hello' . $value,
+ *          fn($value) => $value . '!'
+ *      );
+ *      => $greet('World') === 'Hello World!';
+ *      => $greet('Jim') === 'Hello Jim!';
+ *
+ * @template I
+ * @template O
+ * @template R
+ *
+ * @psalm-param callable(I): O $first
+ * @psalm-param callable(O): R $next
+ *
+ * @psalm-return callable(I): R
+ *
+ * @psalm-pure
+ */
+function after(callable $first, callable $next): callable
+{
+    return fn ($input) => $next($first($input));
+}
--- a/src/Psl/Fun/pipe.php
+++ b/src/Psl/Fun/pipe.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Fun;
+
+use function Psl\Iter\reduce;
+
+/**
+ * Performs left-to-right function composition.
+ * Example
+ *      $pipe = Fun\pipe(
+ *          fn($value) => 'Hello' . $value,
+ *          fn($value) => $value . '!',
+ *          fn($value) => '¡' . $value
+ *      );
+ *      => $greet('World') === '¡Hello World!';
+ *      => $greet('Jim') === '¡Hello Jim!';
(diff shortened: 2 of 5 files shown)


src/Psl/Fun/pipe.php
```diff
@@ -0,0 +1,41 @@
...
+ *          fn($value) => 'Hello' . $value,
+ *          fn($value) => $value . '!',
+ *          fn($value) => '¡' . $value
+ *      );
+ *      => $greet('World') === '¡Hello World!';
+ *      => $greet('Jim') === '¡Hello Jim!';
+ *
+ * @psalm-param list<callable(mixed): mixed> $stages
+ *
+ * @psalm-return callable(mixed): mixed
```
azjezz: hm, i don't really like `mixed` here, what do you think about:

```
@template T

@psalm-param callable(T): T ...$stages

@psalm-return callable(T): T
```

src/Psl/Fun/after.php
```diff
@@ -0,0 +1,41 @@
...
+ * @psalm-param callable(I): O $first
+ * @psalm-param callable(O): R $next
+ *
+ * @psalm-return callable(I): R
+ *
+ * @psalm-pure
+ */
+function after(callable $first, callable $next): callable
+{
+    return fn ($input) => $next($first($input));
```
azjezz:
```suggestion
    return static fn ($input) => $next($first($input));
```

src/Psl/Fun/pipe.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @psalm-pure
+ */
+function pipe(callable ...$stages): callable
+{
+    return fn ($input) => reduce(
```
azjezz:
```suggestion
    return static fn ($input) => reduce(
```

src/Psl/Fun/pipe.php
```diff
@@ -0,0 +1,41 @@
...
+        $stages,
+        /**
+         * @template IO of mixed
+         *
+         * @param IO $input
+         * @param callable(IO): IO $next
+         *
+         * @return IO
+         */
+        fn ($input, int $key, callable $next) => $next($input),
```
azjezz:
```suggestion
        static fn ($input, int $key, callable $next) => $next($input),
```

tests/Psl/Fun/AfterTest.php
```diff
@@ -0,0 +1,32 @@
...
+            fn (string $z): string => $z . '!!'
+        );
+
+        self::assertSame('Hello world!!', $x('Hello'));
+    }
+
+    public function testItCombinesAFunctionThatDealWithDifferentTypes(): void
+    {
+        $x = Fun\after(
+            fn (string $x): int => Byte\length($x),
```
azjezz: you can use `Str\length` instead here.

tests/Psl/Fun/PipeTest.php
```diff
@@ -0,0 +1,41 @@
...
+            fn (string $z): string => $z . '!',
+        );
+
+        self::assertSame('Hello world?!', $x('Hello'));
+    }
+
+    public function testItCombinesMultipleFunctionsThatDealWithDifferentTypes(): void
+    {
+        $x = Fun\pipe(
+            fn (string $x): int => Byte\length($x),
```
azjezz: same.

azjezz (changes requested):

src/Psl/Fun/pipe.php
veewee: Makes sense, but will still be "mixed" internally.
The problem here is that you can't really keep track of all these intermediate callable's Input/Output types.
I'll change it nevertheless.

src/Psl/Fun/after.php
veewee: Just out of curiosity:
Since this is not in a class context (so no `$this`), does adding the "static" keyword actually cause any benefit?

azjezz: No, but more to do with coding style, and getting used to `static fn` over `fn`

tests/Psl/Fun/PipeTest.php
```diff
@@ -0,0 +1,41 @@
...
+    {
+        $x = Fun\pipe(
+            fn (string $x): int => Str\length($x),
+            fn (int $y): string => $y . '!'
+        );
+
+        self::assertSame('5!', $x('Hello'));
+    }
+
+    /** @test */
```
azjezz: no need for @test annotation here.

tests/Psl/Fun/PipeTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+declare(strict_types=1);
+
+namespace Psl\Tests\Fun;
+
+use PHPUnit\Framework\TestCase;
+use Psl\Fun;
+use Psl\Str;
+
+class PipeTest extends TestCase
```
azjezz: final. ( same for the other test )

veewee: might be good to enforce this through cs

azjezz (approved):

PR #61, opened 2020-09-29
[Result] Rename Asio namespace

veewee:
This PR renames the confusing `Asio` namespace to `Result` and renames the items in that namespace.
Once we agree on these things, I can extend the functionality.


src/Psl/Result/IResult.php
```diff
@@ -11,12 +11,12 @@
...
  *
- * This is an interface. You get generally `IResultOrExceptionWrapper<T>` by calling `wrap<T>()`, passing in
- * the `callable(): T`, and a `WrappedResult<T>` or `WrappedException<Te>` is returned.
+ * This is an interface. You get generally `IResult<T>` by calling `tryResultFrom<T>()`, passing in
+ * the `callable(): T`, and a `Ok<T>` or `Failure<Te>` is returned.
  *
  * @template T
  */
-interface IResultOrExceptionWrapper
+interface IResult
```
veewee: I've choosen a less restrictive interface name. Maybe it makes sense to implement a custom implementation at some point.
In libraries like F#, there is a result type that is an union of 2 generic types:

https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/results#syntax

Sadly that is currently not possible with psalm nor php.

src/Psl/Result/tryResultFrom.php
```diff
@@ -0,0 +1,27 @@
...
+ * Wrap the given callable result in a `Ok`, or `Failure` if the callable throws
+ * an `Exception`.
+ *
+ * @template     T
+ *
+ * @psalm-param  (callable(): T) $fun
+ *
+ * @psalm-return IResult<T>
+ */
+function tryResultFrom(callable $fun): IResult
```
veewee: Though this was more descriptive than `wrap`

src/Psl/Result/Failure.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Psl\Asio;
+namespace Psl\Result;
```
veewee, line 5: I've choosen the namespace `Result` since `Res` could also be a resource.
I am a bit confused about the abbreviated namespaces : some are shortened, others are not.
Is there any guide?

src/Psl/Result/Ok.php
```diff
@@ -12,9 +12,9 @@
  *
  * @template    T
  *
- * @implements  IResultOrExceptionWrapper<T>
+ * @implements  IResult<T>
  */
-final class WrappedResult implements IResultOrExceptionWrapper
+final class Ok implements IResult
```
veewee: Used the F# name here, couldn't come up with a better name myself.

src/Psl/Result/tryResultFrom.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Result;
+
+use Exception;
```
veewee: Would it make sense to use `Throwable` instead?

src/Psl/Result/tryResultFrom.php
azjezz: this doesn't follow PSL naming convention ( functions/variables should use `snake_case` casing ), and once renamed to `try_result_from`, it's a bit long.

I think `result` is redundant since we are inside the `Result` namespace.

`from` is also redundant, as I don't see us adding `try` in the future that might hold another meaning.

so I would go with either `Result\try`, or keep it `Result\wrap`.

but I think `Result\try` is better.

src/Psl/Result/IResult.php
azjezz: let's renaming to ResultInterface instead, previously PSL used `ISomething` naming instead of `SomethingInterface` ( similar to C#, and Hack ), but I think `SomethingInterface` is more commonly used in the PHP ecosystem.

src/Psl/Result/Ok.php
azjezz: 👍🏼

src/Psl/Result/tryResultFrom.php
azjezz: no, as we are trying to wrap userland exception, Errors are usually thrown by PHP and they should not be expected ( e.g TypeError )

azjezz (changes requested):

src/Psl/Result/Failure.php
azjezz: not really, some are shortened because they are reserved ( or soft reserved ) by PHP ( e.g string, array, iterable, function .. ). usually try going with the full name ( unless there's a known, popular shorter version, e.g: IO instead of InputOutput ), falling back to the short version.

src/Psl/Result/Ok.php
azjezz: What do you think about `Success`?  :thinking:

src/Psl/Result/IResult.php
veewee: Yeah, is better!

src/Psl/Result/Ok.php
veewee: Love it!

src/Psl/Result/tryResultFrom.php
veewee: Sadly `try` is a reserved keyword. Suggesting to use `try_from` which results in:

```php
$result = try_from($myAwesomeCallback);

return $result->proceed($sendHttpSuccess, $failGracefully);
```


Looks good to me.

azjezz: from doesn't make sense in my opinion, let's keep `wrap`, as we are wrapping the result.

azjezz (approved):

azjezz:
Thank you @veewee

PR #65, opened 2020-09-30
[Result] Add proceed and then with helper functions

veewee:
- [x] ResultInterface::proceed() : Makes it possible to unwrap a result in a type-safe way.
- [x] ResultInterface::then() : Makes it possible to map a result into another result.
- [x] rethrow() : Shortcut function for rethrowing the exception inside the proceed method.
- [x] passthrough() : Shortcut function for returning the value inside the result inside the proceed method.


src/Psl/Result/passthrough.php
```diff
@@ -0,0 +1,19 @@
...
+ * It wil return the result value without any manipulation.
+ *
+ * @template T
+ * @psalm-return callable(T): T
+ * @psalm-pure
+ */
+function passthrough(): callable
+{
+    return static fn ($result) => $result;
+}
```
azjezz: hm, I think this better go inside the `Fun` namespace, since it doesn't have anything to do with `Result` actually. ( maybe `Fun\pass()` ? )

src/Psl/Result/rethrow.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @psalm-return callable(Exception): no-return
+ * @psalm-pure
+ */
+function rethrow(): callable
+{
+    return static function (Exception $exception) {
+        throw $exception;
+    };
+}
```
azjezz: same here. i think `Fun` makes a better home for this.

src/Psl/Result/rethrow.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * This method can be used in combination with the `ResultInterface::proceed()` method inside the `$ifFailed` callback.
+ * It will rethrow the exception.
+ *
+ * @psalm-return callable(Exception): no-return
+ * @psalm-pure
+ */
+function rethrow(): callable
+{
+    return static function (Exception $exception) {
```
azjezz:
```suggestion
    return static function (Exception $exception): void {
```

src/Psl/Result/rethrow.php
```diff
@@ -0,0 +1,21 @@
...
+
+namespace Psl\Result;
+
+use Exception;
+
+/**
+ * This method can be used in combination with the `ResultInterface::proceed()` method inside the `$ifFailed` callback.
+ * It will rethrow the exception.
+ *
+ * @psalm-return callable(Exception): no-return
```
azjezz:
```suggestion
 * @psalm-return pure-callable(Exception): no-return
```

src/Psl/Result/Failure.php
```diff
@@ -65,4 +66,32 @@ public function isFailed(): bool
...
+        return $ifFailed($this->exception);
+    }
+
+    /**
+     * Failed results can not be mapped.
+     * This method will return current instance so that the failing result can be dealt with upstream in the codebase.
+     *
+     * @psalm-return Failure<T, Te>
+     */
+    public function map(callable $mapper): Failure
```
azjezz: I think its more appropriate to throw the exception here ( similar to `getResult()` ), WDYT? this also means we avoid the union return type in `ResultInterface::map`

src/Psl/Result/ResultInterface.php
```diff
@@ -55,4 +56,33 @@ public function isSucceeded(): bool;
...
+     * so that you can transform it to anything you want.
+     *
+     * @template R
+     *
+     * @psalm-param callable(T): R $ifSucceeded
+     * @psalm-param callable(Exception): R $ifFailed
+     *
+     * @psalm-return R
+     */
+    public function proceed(callable $ifSucceeded, callable $ifFailed);
```
azjezz:
```suggestion
    public function proceed(callable $if_succeeded, callable $if_failed);
```

all local variables and parameters should use snake case naming.

also, don't forget that parameter names are important now that we have named parameters in PHP 8 :D 

```php

$result->proceed(ifSucceeded: $callable, ifFailed: $other); // inconsistent

$result->proceed(if_succeeded: $callable, if_failed: $other); // ok

$result->proceed(on_success: $callable, on_failure: $other); // ok
```

src/Psl/Result/rethrow.php
azjezz: actually, no.

src/Psl/Result/ResultInterface.php
veewee: Would you prefer the if* or on* version?

azjezz: whatever works for you :)

src/Psl/Result/Failure.php
veewee: By design, this maps only on successfull results.
Imagine you have an importer that creates a generator of 1000 results and you want to iterator over them and do some secondary stuff that results in a second list of 1000 results.

If one fails, you might want to deal with that independently through `proceed` at some point in the flow. Not during the mapping.

So in that case, you just keep the exception and don't run a secondary action.

My preference goes to this approach, but I am open for other ideas ;)

azjezz: My issue is with the return type.

```php
/** @var list<ResultInterface<User>> $list */
$list = ...;
$list = Arr\filter($list, fn($result) => $result->isSuccessful());

// psalm would report that $list is list<ResultInterface<User>|ResultInterface<int>> now
// even though we are sure that list contains only successful results.
$list = Arr\map($list, fn($result) => $result->map(fn($user) => $user->getId()));
```

while it would be much better to end up with 1 type at the end.

veewee: Was thinking about making the function behave more like the `promise.then` method, with 2 callbacks that are able to both result in a failure or success.

I hope it won't result in similar type restrictions.

azjezz: hm, if the goal is to have the return type be a result, then what do you think about:

```php
/**
 * @template Tr
 *
 * @psalm-param (callable(T): Tr) 			$if_succeeded
 * @psalm-param (callable(Exception): Tr) 	$if_failed
 * 
 * @psalm-return ResultInterface<Tr>
 */
public function then(callable $if_succeeded, callable $if_failed): ResultInterface
{
  // on success:
  return wrap(fn() => $if_succeeded($this->value));
  // on failure:
  return wrap(fn() => $if_failed($this->exception));
}
```

azjezz: P.S: promises are planned ( see #66 )

veewee: Exactly what I had in mind  ...🤣 Will try it out soon.

azjezz: 😄

azjezz:
rebase needed ( migrated to github actions )

src/Psl/Result/Failure.php
veewee: As reference:
https://github.com/reactphp/promise#promiseinterfacethen

Not sure if the param names must match as well.

(This might also mean it can also be used with amphp, since it supports react promises as well - will check it out)

azjezz: let's go with `callable $if_succeeded, callable $if_failed`

veewee: Went with the on* alternative since I think people will recognise it from js and/or the promise API.
Made it consistent everywhere.

Tested it with react php (which works perfectly!)

```php
<?php

use function Psl\Result\wrap;
use function React\Promise\resolve;

require 'vendor/autoload.php';

$loop = React\EventLoop\Factory::create();

$loop->futureTick(
    static function () {
        resolve(wrap(fn () => 5000))
            ->done(static function (int $delay) {
                var_dump($delay);
            });
    }
);

$loop->run();
```


Also tried it with amp. Even though it recognises the promise, it is not able to determine te return value. (This is because amp resolves the promise, but does nothing with the newly returned promise.


```php
use function Amp\Promise\adapt;
use function Amp\Promise\wait;
use function Psl\Result\wrap;

require 'vendor/autoload.php';

$resultPromise = wrap(fn () => 5000);
$ampPromise = adapt($resultPromise);
$ampPromise->onResolve(static function ($delay) {
    // This dumps 'null' since amp knows the promise is resolved, but does nothing with the resolved value.
    var_dump($delay); 
});

// also null : which makes sense since ampPromise is a defered, could fix that with amp coroutines
$result = wait($ampPromise); 

```

gonna try to see if I can make a PR to fix that behaviour inside their package.


Besides that, I tried adding the `reactphp/promise::done` method (https://github.com/reactphp/promise#done-vs-then), but think in our case the `proceed` one is better. Adding a done function will also making it work in amphp harder.
The `then` is suffucient for it to work as a promise.

src/Psl/Result/ResultInterface.php
```diff
@@ -55,4 +55,38 @@ public function isSucceeded(): bool;
...
+     *
+     * @link https://github.com/reactphp/promise#promiseinterfacethen
+     *
+     * @template R
+     * @psalm-param callable(T): R $on_success
+     * @psalm-param callable(Exception): R $on_failure
+     *
+     * @psalm-return ResultInterface<R>
+     */
+    public function then(callable $on_success, callable $on_failure): ResultInterface;
```
veewee, line 91: what do you think about nullable callbacks and fallback to `passthrough` and `rethrow`?
I dont mind making it explicit here, but it could be handy if you only want to configure a `$on_success`

azjezz (approved):
LGTM, I'm not sure about the name 'passthrough' but we can change that later if we agreed on something else

src/Psl/Result/Failure.php
veewee: FYI, git it to work with amp as well...
Looks like I switched the value for the exception in the `onResolve` function above.

```php
use function Amp\call;
use function Amp\Promise\adapt;
use function Amp\Promise\wait;
use function Psl\Result\wrap;
use function React\Promise\resolve;

require 'vendor/autoload.php';

$resultPromise = wrap(fn () => 5000);
$ampPromise = adapt(resolve($resultPromise));
$ampPromise->onResolve(static function (?\Exception $exception, $delay) {
    var_dump($delay);
});

$coroutine = call(function () use ($ampPromise) {
    $value = yield $ampPromise;
    return $value + 1;
});

$result = wait($coroutine);
var_dump($result);

// => 5000
// => 5001
```


We should probably document this once we start documenting the result classes.

PR #68, opened 2020-09-30
add phpdoc for several files

Septikwar:


azjezz:
everything here is already typed either via `@psalm` annotations or PHP type hint.

Issue #63, opened 2020-09-29
Intl API


azjezz:
wrap ext-intl

azjezz:
we already have grapheme and datetime, i really dont see value of adding the rest of intl.. at least not now.

PR #74, opened 2020-10-03
[Str] add grarpheme support

### LINKED ISSUE #62: [Str] Grapheme support
Just like `Psl\Str\Byte` component, we want to add `Psl\Str\Grapheme` that wraps grapheme_str* function in a nicer API that is consistent with our `Psl\Str` and `Psl\Str\Byte` components.

azjezz:
closes #62 
refs #63

### DIFF
--- a/src/Psl/Str/Grapheme/length.php
+++ b/src/Psl/Str/Grapheme/length.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Str\Grapheme;
+
+use Psl;
+
+use function grapheme_strlen;
+
+/**
+ * Returns the length of the given string in grapheme units
+ *
+ * @psalm-pure
+ *
+ * @throws Psl\Exception\InvariantViolationException If unable to convert $string to UTF-16,
+ *                                                   or split it into graphemes.
+ */
+function length(string $string): int
+{
+    $length = grapheme_strlen($string);
+
+    Psl\invariant(null !== $length, 'unable to convert $string to UTF-16');
+    Psl\invariant(false !== $length, 'unable to split $string into graphemes');
+
+    return $length;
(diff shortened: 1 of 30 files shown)


src/Psl/Str/Grapheme/length.php
```diff
@@ -0,0 +1,17 @@
...
+use function grapheme_strlen;
+
+/**
+ * Returns the length of the given string in grapheme units
+ *
+ * @psalm-pure
+ */
+function length(string $string): int
+{
+    return (int) grapheme_strlen($string);
```
azjezz: I think there's a case where grapheme_strlen returns `null`, not sure how to handle that.

veewee: The docs state it always returns an int though:
https://www.php.net/manual/en/function.grapheme-strlen.php

(We could keep it - just to be sure :) )

veewee (dismissed):
Besides the conflicts, it Looks OK to me...

src/Psl/Str/Grapheme/length.php
azjezz: I learned to never belive the documentation 😆  

see:

- https://github.com/php/php-src/blob/e73bf01f9d0c23dd5baed4445b37c93ec4e0ab29/ext/intl/php_intl.stub.php#L246
- https://github.com/php/php-src/blob/7af24eaedd310911940643d7a5f412393b973389/ext/intl/grapheme/grapheme_string.c#L84
- https://github.com/php/php-src/blob/7af24eaedd310911940643d7a5f412393b973389/ext/intl/grapheme/grapheme_string.c#L96

not sure how to handle these cases ...

azjezz: I added invariant violation checks.

Issue #30, opened 2020-07-10
Quality Assurance tools integration. 


azjezz:
This is a meta issue to keep track of which QA tools are being used in PSL, the end goal is to have all these tools ( and more hopefully ) integrated with PSL and as part of our CI. ( PRs welcome! )

### Coding Standards

| Tool            | Description                 | Integrated |
| --------------- | --------------------------- | ---------- |
| php-cs-fixer    | PHP Coding Standards Fixer  | ✅         |
| php_codesniffer | 							| ✔️ 		  |
 

There's a lot of rules we have that are currently not enforced by CI:

- [ ] a file can only contain 1 function declaration
- [ ] Psl\Foo\Bar\Baz\qux function must be declared in src/Psl/Foo/Bar/Baz/qux.php
- [x] function name must be in snake_case
- [ ] argument name must be in snake_case
- [ ] constant Psl\Foo\Bar\BAZ_QUX must be declared in src/Psl/Foo/Bar/constants.php
- [x] all classes should be declared either final or abstract ( unless its an exception )
- [ ] interface names should end with Interface suffix
- [ ] abstract classes should have the Abstract prefix
- [ ] traits should end with Trait suffix
- [x] closure should always be declared static, even within functions, when $this is not used.

### Static Analysis

| Tool            | Description                  | Integrated | Coverage  |
| --------------- | ---------------------------- | ---------- | --------- |
| PSalm           | Static Analysis tool.        | ✅         | 100%      |
| PHPStan         | Static Analysis tool.        | ❌         | 0%        |
| SymfonyInsight  | PHP project quality, done right. | ✅      | n/a     |
| php-assumptions |  Tool to detect assumptions  | ❌         | n/a?      |
| phpmd 		  |								| ❌ 		   | n/a?	  |

### Testing

| Tool            | Description                | Integrated | Coverage  |
| --------------- | -------------------------- | ---------- | --------- | 
| PHPUnit         | Unit Testing framework.    | ✅         | 100%     |

### Security

| Tool                     | Description                                                   | Integrated |
| ------------------------ | ------------------------------------------------------------- | ---------- |
| RoaveSecurityAdvisories  | Security advisories as a simple composer exclusion list.      | ✅         | 
| SymfonoySecurity         | PHP security vulnerabilities monitoring.                      | ❌         |
| Psalm ( taint analysis ) | Static Analysis tool.                                         | ✅         |

azjezz:
cc @fmasa :)

[...]

PR #76, opened 2020-10-04
fix consistency in parameter names + some cleanup 

azjezz:


src/Psl/Str/Byte/lowercase.php
```diff
@@ -4,12 +4,14 @@
 
 namespace Psl\Str\Byte;
 
+use function strtolower;
```
veewee, line 7: Can we enforce function imports through cs?

src/Psl/Str/Byte/lowercase.php
```diff
@@ -4,12 +4,14 @@
...
+
 /**
  * Returns the string with all alphabetic characters converted to lowercase.
  *
  * @psalm-pure
  */
-function lowercase(string $lowercase): string
+function lowercase(string $string): string
```
veewee, line 14: Just checked, the reserved keyword `string` is ok as a named param
https://3v4l.org/u0D8K

However, since it is a type - Is this a good name for a variable?

src/Psl/Str/Byte/replace_ci.php
```diff
@@ -12,5 +14,5 @@
  */
 function replace_ci(string $haystack, string $needle, string $replacement): string
```
veewee, line 15: ci confused me with `continuous integration` at first. No big deal though

veewee (commented):
Looks OK!
How are we going to keep this consistent without having to create a cleanup PR every once in a while?

src/Psl/Str/Byte/lowercase.php
azjezz: this is what php-src is using now. i can't think of a a better alternative.

src/Psl/Str/Byte/replace_ci.php
azjezz: ci here ( and in another Str\ functions ), indicates `case insensative`, I think we need to clearify that in the documentation, but i can't see a better name for it ( replace_case_insensative? too long? )

src/Psl/Str/Byte/lowercase.php
azjezz: not sure, this is the first project where I use code sniffer 😛 if you wanna make changes to it, go ahead, as long as we are following PSR 1 + 2 + 12, I don't care about other changes.

azjezz:
not sure how we can enforce this later in CI.

there's a lot of rules we have that are currently not enforced by CI:
- a file can only contain 1 function declaration
- `Psl\Foo\Bar\Baz\qux` function must be declared in `src/Psl/Foo/Bar/Baz/qux.php`
- function name must be in snake_case
- argument name must be in snake_case
- constant `Psl\Foo\Bar\BAZ_QUX` must be declared in `src/Psl/Foo/Bar/constants.php`
- all classes should be declared either final or abstract ( unless its an exception )
- interface names should end with `Interface` suffix
- abstract classes should have the `Abstract` prefix
- traits should end with `Trait` suffix
- closure should always be declared static, even within functions, when $this is not used.
.... and more

src/Psl/Str/Byte/replace_ci.php
veewee: That is too long indeed... ci is ok - just something you need to know. It makes sense in the context.

veewee:
Added the list to #30 
I think there are at least some of them that we can tackle with cs

veewee (dismissed):

PR #82, opened 2020-10-07
run tests on windows

azjezz:


src/Psl/Env/current_exec.php
```diff
@@ -15,5 +17,11 @@
  */
 function current_exec(): string
 {
-    return realpath((string) $_SERVER['SCRIPT_NAME']);
+    $executable = realpath((string) $_SERVER['SCRIPT_NAME']);
```
azjezz, line 20: realpath reads the link, but not on windows.

PR #84, opened 2020-10-07
[Iter] split reduce into multiple functions

### LINKED ISSUE #73: [Iter] split `reduce` into multiple functions
Currently the `Iter\reduce` function acts as a `reduce_with_keys`. Is there a specific reason for this or is it safe to split it into `reduce`, `reduce_keys` and `reduce_with_keys` function?

https://github.com/azjezz/psl/blob/develop/src/Psl/Iter/reduce.php

veewee:
Fixes #73

- [x] Iter\reduce
- [x] Iter\reduce_keys
- [x] Iter\reduce_with_keys

### DIFF
--- a/src/Psl/Iter/reduce_keys.php
+++ b/src/Psl/Iter/reduce_keys.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Psl\Iter;
 
 /**
- * Reduce iterable using a function.
+ * Reduce iterable keys using a function.
  *
  * The reduction function is passed an accumulator value and the current
  * iterator value and returns a new accumulator. The accumulator is initialized
@@ -13,27 +13,28 @@ namespace Psl\Iter;
  *
  * Examples:
  *
- *      Iter\reduce(Iter\range(1, 5), fn($accumulator, $value) => $accumulator + $value, 0)
- *      => 15
- *
- *      Iter\reduce(Iter\range(1, 5), fn($accumulator, $value) => $accumulator * $value, 1)
- *      => 120
+ *      Iter\reduce_keys(
+ *          Iter\range(1, 5),
+ *          static fn(int $accumulator, int $key): int => $accumulator + $key,
+ *          0,
+  *     )
+ *      => 10
  *
  * @psalm-template Tk
  * @psalm-template Tv
  * @psalm-template Ts
  *
  * @psalm-param iterable<Tk, Tv>        $iterable
- * @psalm-param (callable(?Ts, Tv): Ts) $function
+ * @psalm-param (callable(?Ts, Tk): Ts) $function
  * @psalm-param Ts|null                 $initial
  *
  * @psalm-return Ts|null
  */
-function reduce(iterable $iterable, callable $function, $initial = null)
+function reduce_keys(iterable $iterable, callable $function, $initial = null)
 {
     $accumulator = $initial;
-    foreach ($iterable as $v) {
-        $accumulator = $function($accumulator, $v);
+    foreach ($iterable as $k => $v) {
+        $accumulator = $function($accumulator, $k);
     }
 
     return $accumulator;
--- a/src/Psl/Iter/reduce_with_keys.php
+++ b/src/Psl/Iter/reduce_with_keys.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Iter;
+
+/**
+ * Reduce iterable with both values and keys using a function.
+ *
+ * The reduction function is passed an accumulator value and the current
+ * iterator value and returns a new accumulator. The accumulator is initialized
+ * to $initial.
+ *
+ * Examples:
+ *
+ *      Iter\reduce_with_keys(
+ *          Iter\range(1, 5),
+ *          static fn(int $accumulator, int $key, int $value): int => $accumulator + $value,
+ *          0,
+  *     )
+ *      => 15
+ *
+ *      Iter\reduce_with_keys(
+ *          Iter\range(1, 5),
+ *          static fn(int $accumulator, int $key, int $value): int => $accumulator * $value,
+ *          0,
+ *     )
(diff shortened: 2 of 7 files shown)


src/Psl/Iter/reduce.php
```diff
@@ -19,21 +19,21 @@
  *      Iter\reduce(Iter\range(1, 5), fn($accumulator, $value) => $accumulator * $value, 1)
  *      => 120
  *
- * @psalm-template Tk
+ * @psalm-template Tk of array-key
```
azjezz: iterable can have a non array-key keys.

e.g:

```
/**
 * @var iterable<list<int>, Collection\Vector<int>> $my_iterable
 */
$my_iterable = (static function(): iterable {
  yield [1, 2, 3] => new Collection\Vector([1, 2, 3]);
  yield [4, 5, 6] => new Collection\Vector([4, 5, 6]);
})();
```

src/Psl/Iter/reduce_keys.php
```diff
@@ -0,0 +1,37 @@
...
+ * The reduction function is passed an accumulator value and the current
+ * iterator value and returns a new accumulator. The accumulator is initialized
+ * to $initial.
+ *
+ * Examples:
+ *
+ *      Iter\reduce_keys(Iter\range(1, 5), fn($accumulator, $key) => $accumulator + $key, 0)
+ *      => 10
+ *
+ * @psalm-template Tk of array-key
```
azjezz: same here.

src/Psl/Iter/reduce_keys.php
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Reduce iterable keys using a function.
+ *
+ * The reduction function is passed an accumulator value and the current
+ * iterator value and returns a new accumulator. The accumulator is initialized
+ * to $initial.
+ *
+ * Examples:
+ *
+ *      Iter\reduce_keys(Iter\range(1, 5), fn($accumulator, $key) => $accumulator + $key, 0)
```
azjezz:
```suggestion
 *      Iter\reduce_keys(
 *          Iter\range(1, 5),
 *          static fn(int $accumulator, int $key): int => $accumulator + $key,
 *          0,
  *     )
```

src/Psl/Iter/reduce_with_keys.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Reduce iterable with both values and keys using a function.
+ *
+ * The reduction function is passed an accumulator value and the current
+ * iterator value and returns a new accumulator. The accumulator is initialized
+ * to $initial.
+ *
+ * Examples:
+ *
+ *      Iter\reduce_with_keys(Iter\range(1, 5), fn($accumulator, $key, $value) => $accumulator + $value, 0)
```
azjezz:
```suggestion
 *      Iter\reduce_with_keys(
 *          Iter\range(1, 5),
 *          static fn(int $accumulator, int $key, int $value): int => $accumulator + $value,
 *          0,
  *     )
```

src/Psl/Iter/reduce_with_keys.php
```diff
@@ -0,0 +1,40 @@
...
+ *      => 15
+ *
+ *      Iter\reduce_with_keys(Iter\range(1, 5), fn($accumulator, $key, $value) => $accumulator * $value, 1)
```
azjezz:
```suggestion
 *      Iter\reduce_with_keys(
 *          Iter\range(1, 5),
 *          static fn(int $accumulator, int $key, int $value): int => $accumulator * $value,
 *          0,
 *     )
```

tests/Psl/Iter/ReduceKeysTest.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Tests\Iter;
+
+use PHPUnit\Framework\TestCase;
+use Psl\Iter;
+
+class ReduceKeysTest extends TestCase
```
azjezz:
```suggestion
final class ReduceKeysTest extends TestCase
```

I will probably do a PR tomorrow to switch all test cases to final. it's annoying to see them in the suggestions when typing `Foo extends |` 😅

azjezz (commented):
LGTM 🎉 

just minor cs/typing comments, and then we can merge this

src/Psl/Iter/reduce.php
veewee: I did not know this was possible. On the other hand :  why?! :)

azjezz (approved):

drupol:
Sorry for jumping in here, I have a small question. 

Why do you provide 3 different functions when this could be made in one single function ?

You would by default pass the `$key` AND the `$value` from the iterator to the function, then, the user decide what to do with it?

Wouldn't it be better to have one single function that does this? Wouldn't it be less confusing?

Thanks

azjezz:
We usually provide multiple functions rather than 1 function when they operate either on keys, values, or both ( e.g: `Iter\map`, and `Iter\filter` ), this is done in order to have consistency between these functions.

If `Iter\map` and `Iter\filter` operate on values, having `Iter\reduce` operate on both is inconsistent.

drupol:
Fair enough, thanks!

Keep up the good work :-)

veewee:
@drupol  : by splitting the function up into smaller functions, you can make sure the implementation keeps clean and consice without any unused parameters. I'dd say that is the biggest advantage.

drupol:
> @drupol : by splitting the function up into smaller functions, you can make sure the implementation keeps clean and consice without any unused parameters. I'dd say that is the biggest advantage.

Yes that's true... but it also, in a minor way, duplicate part of the code... and this is something that I'm trying to avoid at all cost in my own development.

veewee:
Agree with that on a general development level. However, in this package the easy of end-user use > the little duplication that is added IMO.

PR #85, opened 2020-10-09
[Fun] Add when function

veewee:
- [x] when

Example

```php
$greet = Fun\when(
    static fn(string $name): bool => $name === 'Jos',
    static fn(string $name): string => 'Bonjour ' . $name . '!',
    static fn(string $name): string => 'Hello ' . $name . '!'
);

$greet('Jos')
=> Str('Bonjour Jos!');

$greet('World')
=> Str('Hello World!');
```

### DIFF
--- a/src/Psl/Fun/when.php
+++ b/src/Psl/Fun/when.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Fun;
+
+/**
+ * Creates a callback that can run a left or right function based on a condition.
+ *
+ * Example:
+ *
+ *      $greet = Fun\when(
+ *          static fn(string $name): bool => $name === 'Jos',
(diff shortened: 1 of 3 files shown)


src/Psl/Fun/when.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ *      $greet('World')
+ *      => Str('Hello World!');
+ *
+ *      $greet('Jos')
+ *      => Str('Bonjour Jos!');
+ *
+ * @template Input
+ * @template Output
+ *
```
azjezz: `Ti` / `To`

usually, we keep templates short and prefix them with `T` ( e.g `Tk` instead of `TKey` / `Tv` instead of `TValue` )

azjezz (dismissed):
LGTM.

azjezz (approved):
Thank you!

PR #86, opened 2020-10-09
[Iter] Improve Iterable factories

veewee:
- [x] `Iterable::from`
- [x] `times(10, $factory)`
- [x] Removed internal `lazy_iterator` in favour of new `Iterable::from` method
- [x] Fixed `Fun\when` static function
- [x] Renamed `Fun\passthrough` to `Fun\identity` (which makes more sense in the iter functions)


src/Psl/Iter/drop_while.php
```diff
@@ -28,7 +25,7 @@
  */
 function drop_while(iterable $iterable, callable $predicate): Iterator
 {
-    return Internal\lazy_iterator(static function () use ($iterable, $predicate) {
+    return Iterator::from(static function () use ($iterable, $predicate) {
```
azjezz: please add missing return type ( there's couple others that are missing return type as well )

```suggestion
    return Iterator::from(static function () use ($iterable, $predicate): Generator {
```

src/Psl/Iter/times.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Examples:
+ *
+ *     Iter\times(5, Fun\Identity())
+ *     => Iter(1, 2, 3, 4, 5)
+ *
+ * @psalm-template T
+ *
+ * @psalm-param int $number How many times should the factory be executed?
+ * @psalm-param (callable(int): T) $factory
```
azjezz:
```suggestion
 * @psalm-param int                        $number How many times should the factory be executed?
 * @psalm-param (callable(int): T)    $factory
```

tests/Psl/Iter/IteratorTest.php
```diff
@@ -18,6 +18,13 @@ public function testCreateFromGenerator(): void
         static::assertCount(3, $iterator);
     }
 
+    public function testCreateFromFactory(): void
+    {
+        $iterator = Iter\Iterator::from((fn () => yield from [1, 2, 3]));
```
azjezz, line 23:
```suggestion
        $iterator = Iter\Iterator::from(static fn () => yield from [1, 2, 3]);
```

tests/Psl/Iter/TimesTest.php
```diff
@@ -0,0 +1,27 @@
...
+declare(strict_types=1);
+
+namespace Psl\Tests\Iter;
+
+use PHPUnit\Framework\TestCase;
+use Psl\Exception\InvariantViolationException;
+use Psl\Fun;
+use Psl\Iter;
+
+class TimesTest extends TestCase
```
azjezz:
```suggestion
final class TimesTest extends TestCase
```

tests/Psl/Fun/IdentityTest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Tests\Fun;
+
+use PHPUnit\Framework\TestCase;
+use Psl\Fun;
+
+class IdentityTest extends TestCase
```
azjezz:
```suggestion
final class IdentityTest extends TestCase
```

src/Psl/Iter/times.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @psalm-return Iterator<int, T>
+ *
+ * @throws Psl\Exception\InvariantViolationException If $number < 1
+ */
+function times(int $number, callable $factory): Iterator
+{
+    Psl\invariant($number >= 1, 'The number of times you want to iterate must be at least 1.');
```
azjezz:
```suggestion
    Psl\invariant(null === $multiplier || $multiplier >= 0, 'Expected a non-negative multiplier');
```

src/Psl/Iter/times.php
```diff
@@ -0,0 +1,36 @@
...
+ * @psalm-template T
+ *
+ * @psalm-param int $number How many times should the factory be executed?
+ * @psalm-param (callable(int): T) $factory
+ *
+ * @psalm-return Iterator<int, T>
+ *
+ * @throws Psl\Exception\InvariantViolationException If $number < 1
+ */
+function times(int $number, callable $factory): Iterator
```
azjezz: I think for the sake of consistency with `Str\repeat` and `Iter\repeat`, the `$number` argument must be last.

`times` doesn't really hold for me here, as it can be intercepted as `Iter\repeat`, what do you think about `reproduce` instead ? 

```suggestion
function reproduce(callable $factory, ?int $number = null): Iterator
```

( when a null argument is given, this will produce `Math\INFINITY` items, similar to `Iter\repeat` )

azjezz (commented):
Great additions 👍🏼 

some nitpicking as usual 😛

src/Psl/Iter/times.php
azjezz: GitHub editor is not the best for this 😆 but you get the idea.

src/Psl/Iter/times.php
veewee: It isn't really multiplying but incrementing ... Or do you want me to make $numer multiplyable? Because that sounds more like a case for `map()`.

azjezz: I just copy-pasted this from Iter\repeat 😅 we can keep it `$number`

src/Psl/Iter/times.php
veewee: Sounds good! The INFINITY thing is nice! makes sense

azjezz (approved):

azjezz:
Thanks 🎉

PR #87, opened 2020-10-15
[QA] Add PHP-CS-Fixer 

veewee:
Automatically check some CS rules as described in #30 

FYI : the config and the fixes are in 2 separate commits. I can change those independently if you wish to change a setting.

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -23,3 +23,6 @@ jobs:
 
       - name: "checking coding standards ( codesniffer )"
         run: "composer cs:check"
+
+      - name: "checking coding standards ( php-cs-fixer )"
+        run: "composer cs-fixer:check"
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -1,0 +1,52 @@
+<?php
+
+return PhpCsFixer\Config::create()
+    ->setFinder(
+        \Symfony\Component\Finder\Finder::create()
+            ->in([
+                __DIR__ . '/src',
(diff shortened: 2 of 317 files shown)


veewee:
The latest psalm version gives this issue:

```
ERROR: RedundantCondition - src/Psl/Str/convert_encoding.php:27:5 - Type string for $result is never false (see https://psalm.dev/122)
    Psl\invariant(false !== $result, 'Unable to convert $string from $from_encoding to $to_encoding.');


ERROR: RedundantCondition - src/Psl/Str/convert_encoding.php:27:19 - string can never contain false (see https://psalm.dev/122)
    Psl\invariant(false !== $result, 'Unable to convert $string from $from_encoding to $to_encoding.');


```

Haven't touched the file. Is it safe to remove the invariant check? Or shall I add an ignore for that rule?
Looks like it always returns a string: https://www.php.net/manual/en/function.mb-convert-encoding.php

azjezz:
okay; you can go ahead and remove that check.

src/Psl/Exception/InvariantViolationException.php
```diff
@@ -4,6 +4,6 @@
 
 namespace Psl\Exception;
 
-final class InvariantViolationException extends RuntimeException
+class InvariantViolationException extends RuntimeException
```
azjezz: why was final removed from this class?

veewee: Manual mistake, re-added it.

azjezz (approved):

Issue #60, opened 2020-09-29
XML API


veewee:

veewee:
@azjezz 
Knowing that I implemented pretty much everything XML related in https://github.com/veewee/xml instead now., how do you want to continue with this issue?

Since I know you don't really like the DOM part, I am willing to move it over to the php-standard-library organisation and maintain it in there as well, if that might be a solution for you. I think the other components turned out quite good, but feel free to propose improvements.

azjezz:
not sure really, i assigned it to you since you are much more familiar with it.

i think we want to have something to do with `DOM` in here, but we don't want to also dive deep into the XML rabbit hole.

so we can have a replacement for `DOMDocument`, but not the other parts or xml ( we can refer people to `veewee/xml` if they need more advanced features ).

WDYT?

veewee:
The dom extension is the worst to wrap, since it's api is so broad. That is why I went with a more flexible design in veewee/Xml: leaking parts of the dom objects to the end user whilst keeping in control about the more error-prone parts.

I dont really know how to simplify it for this package's standards / point of view.

For this package a xml_encode and xml_decode function could be nice. But I already wrote that for veewee/xml and it's quite some logic to get everything parsed safely.

So i'dd just recommend using that one instead :p

azjezz:
If you think it would be too much to wrap DOM here, i think we can close the issue then, and agree to delegate things to `veewee/xml`, we don't have to dig deep into html/xml stuff, for html ( #140 ), we just need `encode`/`decode` functions ( same as PHPs `htmlspecialchars` )

veewee:
Ok. Closing this one for now.

PR #88, opened 2020-10-16
[XML] Basic error handling

veewee:
This PR provides some functions to safely handle libxml errors whilst using dom, simplexml, xmlreader, ...
It can be used inside for building an XML component (#60)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,10 +13,12 @@
         "php": "^7.4 | ^8.0",
         "ext-bcmath": "*",
         "ext-json": "*",
+        "ext-libxml": "*",
         "ext-mbstring": "*",
         "ext-sodium": "*"
     },
     "require-dev": {
+        "ext-simplexml": "*",
         "phpunit/phpunit": "^9.0",
         "php-coveralls/php-coveralls": "^2.2",
         "friendsofphp/php-cs-fixer": "^2.16",
--- a/src/Psl/Internal/Loader.php
+++ b/src/Psl/Internal/Loader.php
@@ -336,6 +336,10 @@ final class Loader
         'Psl\Hash\equals',
         'Psl\Hash\Hmac\hash',
         'Psl\Hash\Hmac\algorithms',
+        'Psl\Xml\Internal\detect_errors',
+        'Psl\Xml\Internal\issue_collection_from_xml_errors',
+        'Psl\Xml\Internal\issue_from_xml_error',
+        'Psl\Xml\Internal\issue_level_from_xml_error',
     ];
 
     public const INTERFACES = [
@@ -353,6 +357,7 @@ final class Loader
         'Psl\Observer\SubjectInterface',
         'Psl\Observer\ObserverInterface',
         'Psl\Result\ResultInterface',
+        'Psl\Xml\Exception\ExceptionInterface',
     ];
 
     public const TRAITS = [
@@ -393,6 +398,10 @@ final class Loader
         'Psl\Json\Exception\DecodeException',
         'Psl\Json\Exception\EncodeException',
         'Psl\Hash\Context',
+        'Psl\Xml\Issue\Issue',
+        'Psl\Xml\Issue\IssueCollection',
+        'Psl\Xml\Issue\Level',
+        'Psl\Xml\Exception\XmlException',
     ];
 
     private const TYPE_CONSTANTS = 1;
--- a/src/Psl/Internal/Loader.php
+++ b/src/Psl/Internal/Loader.php
@@ -401,7 +401,7 @@ final class Loader
         'Psl\Xml\Issue\Issue',
         'Psl\Xml\Issue\IssueCollection',
         'Psl\Xml\Issue\Level',
-        'Psl\Xml\Exception\XmlException',
+        'Psl\Xml\Exception\RuntimeException',
     ];
 
     private const TYPE_CONSTANTS = 1;
--- a/src/Psl/Xml/Exception/XmlException.php
+++ b/src/Psl/Xml/Exception/XmlException.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Exception;
+
+use Exception;
+use Psl\Exception\RuntimeException;
+use Psl\Xml\Issue\IssueCollection;
+
+class XmlException extends RuntimeException implements ExceptionInterface
+{
+    private function __construct(string $message, Exception $previous = null)
+    {
+        parent::__construct(
+            $message,
+            (int) ($previous ? $previous->getCode() : 0),
+            $previous
+        );
+    }
+
+    public static function combineExceptionWithIssues(Exception $exception, IssueCollection $errors): self
+    {
+        return new self(
+            $exception->getMessage() . PHP_EOL . $errors->toString(),
+            $exception
+        );
+    }
+}
--- a/src/Psl/Xml/Internal/detect_errors.php
+++ b/src/Psl/Xml/Internal/detect_errors.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Internal;
+
+use LibXMLError;
+use Psl\Exception\InvariantViolationException;
+use Psl\Result;
+use Psl\Result\ResultInterface;
+use Psl\Xml\Issue\IssueCollection;
+
+use function libxml_clear_errors;
+use function libxml_get_errors;
+use function libxml_use_internal_errors;
+
+/**
+ * You can wrap this detector around a function that interacts with xml, dom, simplexml, ...
+ * It wil return a result that contains the result or an exception if an error was encountered.
+ * Besides the result, you will receive a list of XML errors.
+ * No XML warnings will be displayed in the meantime.
+ *
+ * @internal
+ *
+ * @template Tr
+ *
+ * @psalm-param callable(): Tr $run
+ * @psalm-return array{ResultInterface<Tr>, IssueCollection}
+ *
+ * @throws InvariantViolationException
+ */
+function detect_errors(callable $run): array
+{
+    $previousErrorReporting = libxml_use_internal_errors(true);
+    libxml_clear_errors();
+
+    $result = Result\wrap($run);
+
+    /** @var list<LibXMLError> $errors */
+    $errors = libxml_get_errors();
+    libxml_clear_errors();
+    libxml_use_internal_errors($previousErrorReporting);
+
+    return [$result, issue_collection_from_xml_errors($errors)];
+}
--- a/src/Psl/Xml/Internal/issue_from_xml_error.php
+++ b/src/Psl/Xml/Internal/issue_from_xml_error.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Internal;
+
+use LibXMLError;
+use Psl\Xml\Issue\Issue;
+
+/**
+ * @internal
+ */
+function issue_from_xml_error(LibXMLError $error): ?Issue
+{
+    $level = issue_level_from_xml_error($error);
+    if ($level === null) {
+        return null;
+    }
+
+    return new Issue(
+        $level,
+        $error->code,
+        $error->column,
+        $error->message,
+        $error->file,
+        $error->line,
+    );
+}
--- a/src/Psl/Xml/Internal/issue_level_from_xml_error.php
+++ b/src/Psl/Xml/Internal/issue_level_from_xml_error.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Internal;
+
+use LibXMLError;
+use Psl\Xml\Issue\Level;
+
+/**
+ * @internal
+ */
+function issue_level_from_xml_error(LibXMLError $error): ?Level
+{
+    switch ($error->level) {
+        case LIBXML_ERR_ERROR:
+            return Level::error();
+        case LIBXML_ERR_FATAL:
+            return Level::fatal();
+        case LIBXML_ERR_WARNING:
+            return Level::warning();
+    }
+
+    return null;
+}
--- a/src/Psl/Xml/Issue/Issue.php
+++ b/src/Psl/Xml/Issue/Issue.php
@@ -6,6 +6,9 @@ namespace Psl\Xml\Issue;
 
 use Psl\Str;
 
+/**
+ * @psam-immutable
+ */
 final class Issue
 {
     private Level $level;
--- a/src/Psl/Xml/Issue/Issue.php
+++ b/src/Psl/Xml/Issue/Issue.php
@@ -1,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Issue;
+
+use Psl\Str;
+
+final class Issue
+{
+    private Level $level;
+    private int $code;
+    private int $column;
+    private string $message;
+    private string $file;
+    private int $line;
+
+    public function __construct(
+        Level $level,
+        int $code,
+        int $column,
+        string $message,
+        string $file,
+        int $line
+    ) {
+        $this->level = $level;
+        $this->code = $code;
+        $this->column = $column;
+        $this->message = $message;
+        $this->file = $file;
+        $this->line = $line;
+    }
+
+    public function level(): Level
+    {
+        return $this->level;
+    }
+
+    public function code(): int
+    {
+        return $this->code;
+    }
+
+    public function column(): int
+    {
+        return $this->column;
+    }
+
+    public function message(): string
+    {
+        return $this->message;
+    }
+
+    public function file(): string
+    {
+        return $this->file;
+    }
+
+    public function line(): int
+    {
+        return $this->line;
+    }
+
+    /**
+     * @psalm-suppress MissingThrowsDocblock
+     */
+    public function toString(): string
+    {
+        return Str\format(
+            '[%s] %s: %s (%s) on line %s,%s',
+            Str\uppercase($this->level->toString()),
+            $this->file,
+            $this->message,
+            $this->code,
+            $this->line,
+            $this->column
+        );
+    }
+}
--- a/src/Psl/Xml/Issue/IssueCollection.php
+++ b/src/Psl/Xml/Issue/IssueCollection.php
@@ -11,6 +11,11 @@ use Psl\Iter\Iterator;
 use Psl\Math;
 use Psl\Str;
 
+/**
+ * @template-implements IteratorAggregate<int, Issue>
+ *
+ * @psalm-immutable
+ */
 final class IssueCollection implements Countable, IteratorAggregate
 {
     /**
--- a/src/Psl/Xml/Issue/IssueCollection.php
+++ b/src/Psl/Xml/Issue/IssueCollection.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Issue;
+
+use Countable;
+use IteratorAggregate;
+use Psl\Arr;
+use Psl\Iter\Iterator;
+use Psl\Math;
+use Psl\Str;
+
+final class IssueCollection implements Countable, IteratorAggregate
+{
+    /**
+     * @var list<Issue>
+     */
+    private array $issues;
+
+    public function __construct(Issue ...$errors)
+    {
+        $this->issues = $errors;
+    }
+
+    /**
+     * @return Iterator<int, Issue>
+     */
+    public function getIterator(): Iterator
+    {
+        return Iterator::create($this->issues);
+    }
+
+    public function count(): int
+    {
+        return count($this->issues);
+    }
+
+    /**
+     * @param (callable(Issue): bool) $filter
+     */
+    public function filter(callable $filter): self
+    {
+        return new self(...Arr\filter($this->issues, $filter));
+    }
+
+    public function getHighestLevel(): ?Level
+    {
+        $issue = Math\max_by($this->issues, static fn(Issue $issue): int => $issue->level()->value());
+
+        return $issue ? $issue->level() : null;
+    }
+
+    public function toString(): string
+    {
+        $values = Arr\values(Arr\map($this->issues, static fn (Issue $error): string => $error->toString()));
+
+        return Str\join($values, PHP_EOL);
+    }
+}
--- a/src/Psl/Xml/Issue/Level.php
+++ b/src/Psl/Xml/Issue/Level.php
@@ -1,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Issue;
+
+/**
+ * @psalm-immutable
+ */
+final class Level
+{
+    private const LEVEL_WARNING = 1;
+    private const LEVEL_ERROR = 2;
+    private const LEVEL_FATAL = 3;
+
+    private int $value;
+
+    private function __construct(int $value)
+    {
+        $this->value = $value;
+    }
+
+    /**
+     * @psalm-pure
+     */
+    public static function error(): self
+    {
+        return new self(self::LEVEL_ERROR);
+    }
+
+    /**
+     * @psalm-pure
+     */
+    public static function fatal(): self
+    {
+        return new self(self::LEVEL_FATAL);
+    }
+
+    /**
+     * @psalm-pure
+     */
+    public static function warning(): self
+    {
+        return new self(self::LEVEL_WARNING);
+    }
+
+    public function value(): int
+    {
+        return $this->value;
+    }
+
+    /**
+     * @return 'warning'|'fatal'|'error'
+     */
+    public function toString(): string
+    {
+        if ($this->isWarning()) {
+            return 'warning';
+        }
+
+        if ($this->isError()) {
+            return 'error';
+        }
+
+        return 'fatal';
+    }
+
+    public function matches(Level $level): bool
+    {
+        return $level->value() === $this->value;
+    }
+
+    public function isError(): bool
+    {
+        return $this->matches(self::error());
+    }
+
+    public function isFatal(): bool
+    {
+        return $this->matches(self::fatal());
+    }
+
+    public function isWarning(): bool
+    {
+        return $this->matches(self::warning());
+    }
+}
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -16,7 +16,7 @@ jobs:
           php-version: "7.4"
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
-          extensions: bcmath, mbstring, intl, sodium, json
+          extensions: bcmath, mbstring, intl, sodium, json, libxml, simplexml
 
       - name: "installing dependencies"
         run: "composer update --no-interaction --no-progress"
--- a/.github/workflows/security-analysis.yml
+++ b/.github/workflows/security-analysis.yml
@@ -16,7 +16,7 @@ jobs:
           php-version: "7.4"
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
-          extensions: bcmath, mbstring, intl, sodium, json
+          extensions: bcmath, mbstring, intl, sodium, json, libxml, simplexml
 
       - name: "installing dependencies"
         run: "composer update --no-interaction --no-progress"
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -16,7 +16,7 @@ jobs:
           php-version: "7.4"
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
-          extensions: bcmath, mbstring, intl, sodium, json
+          extensions: bcmath, mbstring, intl, sodium, json, libxml, simplexml
 
       - name: "installing dependencies"
         run: "composer update --no-interaction --no-progress"
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -28,7 +28,7 @@ jobs:
           php-version: "${{ matrix.php-version }}"
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
-          extensions: bcmath, mbstring, intl, sodium, json
+          extensions: bcmath, mbstring, intl, sodium, json, libxml, simplexml
 
       - name: "caching dependencies"
         uses: "actions/cache@v2"
--- a/src/Psl/Xml/Exception/ExceptionInterface.php
+++ b/src/Psl/Xml/Exception/ExceptionInterface.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Exception;
+
+use Psl\Exception\ExceptionInterface as PslExceptionInterface;
+
+interface ExceptionInterface extends PslExceptionInterface
+{
+}
--- a/src/Psl/Xml/Exception/RuntimeException.php
+++ b/src/Psl/Xml/Exception/RuntimeException.php
@@ -5,10 +5,10 @@ declare(strict_types=1);
 namespace Psl\Xml\Exception;
 
 use Exception;
-use Psl\Exception\RuntimeException;
+use Psl\Exception\RuntimeException as PslRuntimeException;
 use Psl\Xml\Issue\IssueCollection;
 
-class XmlException extends RuntimeException implements ExceptionInterface
+final class RuntimeException extends PslRuntimeException implements ExceptionInterface
 {
     private function __construct(string $message, Exception $previous = null)
     {
--- a/src/Psl/Xml/Internal/issue_collection_from_xml_errors.php
+++ b/src/Psl/Xml/Internal/issue_collection_from_xml_errors.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Internal;
+
+use LibXMLError;
+use Psl\Arr;
+use Psl\Xml\Issue\Issue;
+use Psl\Xml\Issue\IssueCollection;
+
+/**
+ * @internal
+ *
+ * @psalm-param list<LibXMLError> $errors
+ */
+function issue_collection_from_xml_errors(array $errors): IssueCollection
+{
+    return new IssueCollection(
+        ...Arr\filter_nulls(Arr\map(
+            $errors,
+            static fn(LibXMLError $error): ?Issue => issue_from_xml_error($error)
+        ))
+    );
+}
--- a/tests/Psl/Xml/Exception/XmlExceptionTest.php
+++ b/tests/Psl/Xml/Exception/XmlExceptionTest.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Tests\Xml\Exception;
+
+use Exception;
+use PHPUnit\Framework\TestCase;
+use Psl\Tests\Xml\Issue\UseIssueTrait;
+use Psl\Xml\Exception\ExceptionInterface;
+use Psl\Xml\Exception\XmlException;
+use Psl\Xml\Issue\IssueCollection;
+use Psl\Xml\Issue\Level;
+
+final class XmlExceptionTest extends TestCase
+{
+    use UseIssueTrait;
+
+    public function testItCanThrowAnExceptionContainingXmlErrors(): void
+    {
+        $exception = new Exception('nonono');
+        $issues = new IssueCollection(
+            $this->createIssue(Level::fatal()),
+            $this->createIssue(Level::warning())
+        );
+
+        $this->expectException(XmlException::class);
+        $this->expectException(ExceptionInterface::class);
+        $this->expectExceptionMessage($exception->getMessage() . PHP_EOL . $issues->toString());
+
+        throw XmlException::combineExceptionWithIssues($exception, $issues);
+    }
+}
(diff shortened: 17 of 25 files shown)


src/Psl/Xml/detect_errors.php
```diff
@@ -0,0 +1,43 @@
...
+    $previousErrorReporting = libxml_use_internal_errors(true);
+    libxml_clear_errors();
+    $result = $run();
+
+    /** @var list<LibXMLError> $errors */
+    $errors = libxml_get_errors();
+    libxml_clear_errors();
+    libxml_use_internal_errors($previousErrorReporting);
+
+    if ($errors) {
```
veewee: Not sure if this is the best way. For example : whilst validating XSD schema's, it is possible to receive warnings.
Should those be 'blocking'? It's not that it results in a completely unusable XML format.

How would we be able to deal with this? I was thinking about either

- Adding a parameter to set an error level
- return a tupple so that you can have both a result and/or an exception - similar to [go's error handling](https://blog.golang.org/error-handling-and-go)

WDYT?

src/Psl/Xml/format_error.php
```diff
@@ -0,0 +1,26 @@
...
+
+use LibXMLError;
+use Psl\Str;
+
+/**
+ * This function can be used to format a LibXMLError.
+ *
+ * @psalm-suppress MissingThrowsDocblock
+ */
+function format_error(LibXMLError $error): string
```
veewee: I decided to add all these format functions in here.
This is currently used inside the exception, but could also be used to e.g. build an XML linter.
Don't think any of these functions should go into the internal namespace.

src/Psl/Xml/format_errors.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * This function can be used to format multiple LibXMLErrors.
+ *
+ * @psalm-param list<LibXMLError> $errors
+ */
+function format_errors(array $errors): string
+{
+    return Str\join(
+        Arr\values(
```
veewee: Any clue why psalm doesn't know that the map actually returns a list?
I had to add this `values` hack since I couldn't find it out... The types in `Arr\map` seem valid.

src/Psl/Xml/format_error.php
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Psl\Xml;
+
+use LibXMLError;
+use Psl\Str;
+
+/**
+ * This function can be used to format a LibXMLError.
+ *
+ * @psalm-suppress MissingThrowsDocblock
```
azjezz: we should move this above the function that causes the exception.

src/Psl/Xml/format_error.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+function format_error(LibXMLError $error): string
+{
+    return sprintf(
```
azjezz:
```suggestion
    return Str\format(
```

we should avoid using php std functions if we have an alternative 😉

src/Psl/Xml/Exception/XmlErrorsException.php
```diff
@@ -0,0 +1,42 @@
...
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Exception;
+
+use LibXMLError;
+use Psl\Exception\RuntimeException;
+use Psl\Xml;
+
+final class XmlErrorsException extends RuntimeException
```
azjezz: we need `Psl\Xml\Exception\ExceptionInterface`, and the name here should probably change 🤔 

`XmlErrorsException` doesn't really give out anything.

src/Psl/Xml/detect_errors.php
azjezz: we should be as strict as possible here, i think throwing an exception is the right thing to do, however, throwing an exception that is compsed of multiple errors seems wrong, we should throw only for the first error.

`$error = Arr\first($errors)` and we throw that.

we can add a error level parameter, i don't like the API but i think it would be okay since this should be in the internal namespace, where we can have strictness configured in the public api.

e.g:

$parser = new Xml\Parser(strict: true); // true => all, false => ignore warnings.

azjezz (changes requested):
I don't think any of these function should be in `Psl\Xml` namespace, but `Psl\Xml\Internal`,

they are useful as building blocks for a better API, but they still refernce the php xml API ( e.g `format_error(LibXMLError $error)` ).

by using `Psl\Xml`, the end-user should not have any prior knowledge of php libxml/simplexml, we should take care of errors, and always throw exceptions.

src/Psl/Xml/format_errors.php
azjezz: `Arr\map` is `Arr\map<Tk as array-key, Tv, Ts>(array<Tk, Tv>, (function(Tk): Ts)): array<Tk, Ts>`, psalm can't know from this signature that the returned array is in the same order, all it knows is that the key type is the same, and since `list<T>` is `array<int, T>` ( but ordered ), the return type would be `array<int, T>`, again, psalm can't verify that it's ordered.

src/Psl/Xml/detect_errors.php
veewee: I uploaded an alternative solution which is IMO the way to go:

The `$run` method can either succeed or fail, so is a result.
Next to that result : you get a list of errors, which is the context on why something failed or like warnings when the run method succeeded.

An Example can be seen in phpunit : if you add invalid xml tag, phpunit will display e.g.

```
  Warning - The configuration file did not pass validation!
  The following problems have been detected:

  Line 29:
  - Element 'someinvaliddtag': This element is not expected.

  Test results may not be as expected.
```

However, it still works as expected.


Only returning the first error is not OK, since E.g. parsing an xsd may result in multiple issues. One might be more blocking than another one.


By doing it the new way, the implementing function is in full charge on how to deal with exceptions.
For example:

- load xml 
  - The load function will throw an exception : cannot load, the errors collection will explain why.
  - The load function can return a document, but with some notices to display (like phpunit)
- validate xml
  - The implementation can determine which level is blocking, the detect function just returns everything.
  - The validation bool result could be changed based on errors - if it makes sense

src/Psl/Xml/format_error.php
veewee: That doesnt work in psalm sadly

veewee:
Hi @azjezz 

I'm a bit short on time but wanted to share a WIP of an alternative approach. (no tests were changed, types are broken + exception class remained untouched)

This one wraps the error in a PSL class, making it publicly usefull + adds an error collection which can be used to filter.
The format functions are moved inside those classes, making it more straight forward.

The `detect_errors` function now returns a tupple of `[Result<T>, XmlErrorCollection]`, making the implementing function in control of what happens. (see comment on the code snippet.)

What do you think about this approach comparing to the original one?

azjezz:
I think error handling by collecting errors and return a collection of errors is not always the right thing to do, it is nice if you are building a linter, but personally i would prefer throwing an exception for the first issue.

what we can do is have an error handler similar to PHP-parser ( https://github.com/nikic/PHP-Parser/tree/master/lib/PhpParser/ErrorHandler ) that can be configured in the parser, WDYT?

veewee:
> I think error handling by collecting errors and return a collection of errors is not always the right thing to do, it is nice if you are building a linter, but personally i would prefer throwing an exception for the first issue.

That is exactly the point : you don't know what they are building.
The function `detect_errors` is there to detect the errors whilst running an XML command, but still keep the outcome of the function.
The idea behind it is that an implementing function is in charge on how to handle errors.


(See examples below)

> what we can do is have an error handler similar to PHP-parser ( https://github.com/nikic/PHP-Parser/tree/master/lib/PhpParser/ErrorHandler ) that can be configured in the parser, WDYT?

I don't think this is possible, since libxml just returns all errors when being asked for. It is not parsing and throwing errors whilst parsing.

Besides that : if you get the complete list of errors at once, what is the point in displaying only the first one? Imagine you have 20 errors, that means you need to execute the code 20 times in order to resolve all errors... In that case it makes more sense to display them all (like a phpcs or php-cs-fixer result)


The thing I had in mind is that the method that implements the function is responsible for the error handling.
Example code (for domdocument):

```php
function load_xml_string(string $xml): \DOMDocument
{
    [$result, $errors] = detect_errors(static function() use ($xml): DOMDocument {
        $dom = new DOMDocument();
        $result = $dom->loadXML($xml);

        Psl\invariant(
            $result,
            'Could not load XML string'
        );
        
        return $dom;
    });

    return $result->proceed(
        identity(),
        static function (\Exception  $exception) use ($errors): void {
            throw SomeXmlException::wrapErrors($exception, $errors);
        }
    );
}

function validate_xsd(string $xml_string, string $xsd): XmlErrorCollection
{
    $xml = load_xml_string($xml_string);
    [$result, $errors] = detect_errors(static function() use ($xml, $xsd): bool {
        return $xml->schemaValidate($xsd);
    });
    
    return $errors;
}

echo validate_xsd($xml, $xsd)->ignoreWarnings()->toString();
```

This means that the function can be used for an internal XML api to detect errors and that the other psl xml component will deal with the errors based on the case.
I did not put it in internal, since it can still be used for other libraries to base xml error handling on.


Another thing I had in mind is builden an XML reader that returns xml strings based on callbacks.
That could look like this:

```php
$reader = new XmlReader::forFile($xml_file, [
    fn (NodeSequence $tree) => $tree()->current()->name() == 'x' && $tree()->hasParentWithName('y');
]);
/** @var $xml_items string[] */
$xml_items = $reader->produce();
```

In this case, the Produce function will use the `detect_errors` function and will only block (as-in throw exception) on fatals and errors.


I hope this makes more sense to you, since I think that the `detect_errors` function is a quite good solution right now :)

Feel free to suggest another approach  if you still think different on the topic.

azjezz:
Well, i get your point, but we also want to wrap `DOMDocument` and any other related API in PSL, meaning that if someone would use PSL to build per-say a XSD validator, they would rely completely on PSL, rather than relying on PSL only to detect errors.

e.g:

```php
$xml = Xml\load($xml_string);
$violations = $xml->validate($xsd);
if ($violations->count()) {
  foreach($violations as $violation) {
    echo $violation->getMessage();
  }
}

```

> I hope this makes more sense to you, since I think that the detect_errors function is a quite good solution right now :)

I agree; i just think that it should be internal. not something exposed in the public API.

veewee:
Ok, point taken. Will move the function to internal, but keep it more or less as-is.
As discussed before : I am still not sure if it is a good idea to wrap everything inside dom or simplexml in a custom API.
There is a lot in there and some of it is quite complex. We'll see where we land it!

Propably gonna open up some suggestion issues in which we can discuss stuff before implementing on that topic.

veewee:
Scruntinizer failure is not related to PR.

src/Psl/Xml/Error/XmlErrorCollection.php
```diff
@@ -0,0 +1,73 @@
...
+use Countable;
+use IteratorAggregate;
+use Psl\Arr;
+use Psl\Iter;
+use Psl\Iter\Iterator;
+use Psl\Str;
+
+use const Psl\Xml\ERR_WARNING;
+
+final class XmlErrorCollection implements Countable, IteratorAggregate
```
azjezz:
```suggestion
final class ErrorCollection implements Countable, IteratorAggregate
```

we don't have to repeat ourselves :) Xml is already in the FQCN.

src/Psl/Xml/Error/XmlErrorCollection.php
```diff
@@ -0,0 +1,73 @@
...
+{
+    /**
+     * @var list<XmlError>
+     */
+    private array $errors;
+
+    public function __construct(XmlError ...$errors)
```
azjezz:
```suggestion
    public function __construct(array $errors, bool $ignore_warnings = false)
```

src/Psl/Xml/Error/XmlErrorCollection.php
```diff
@@ -0,0 +1,73 @@
...
+
+    public function ignoreWarnings(): self
+    {
+        return new self(
+            ...Arr\filter(
+                $this->errors,
+                static fn (XmlError $error): bool => $error->level() !== ERR_WARNING
+            )
+        );
+    }
```
azjezz: I think this should be moved to the constructor instead. ( see above )

src/Psl/Xml/Exception/XmlException.php
```diff
@@ -0,0 +1,29 @@
...
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Exception;
+
+use Exception;
+use Psl\Exception\RuntimeException;
+use Psl\Xml\Error\XmlErrorCollection;
+
+class XmlException extends RuntimeException
```
azjezz: we should add `Psl\Xml\Exception\ExceptionInterface` here ( as we do with every other component )

src/Psl/Xml/Error/XmlError.php
```diff
@@ -0,0 +1,113 @@
...
+namespace Psl\Xml\Error;
+
+use LibXMLError;
+use Psl\Str;
+
+use const Psl\Xml\ERR_ERROR;
+use const Psl\Xml\ERR_FATAL;
+use const Psl\Xml\ERR_WARNING;
+
+final class XmlError
```
azjezz:
```suggestion
final class Error
```

src/Psl/Xml/Error/XmlError.php
```diff
@@ -0,0 +1,113 @@
...
+{
+    private int $level;
+    private int $code;
+    private int $column;
+    private string $message;
+    private string $file;
+    private int $line;
```
azjezz: we should we make these public + add `@psalm-readonly` ? ( also, the class should marked immutable )

src/Psl/Xml/constants.php
```diff
@@ -0,0 +1,15 @@
...
+
+use const LIBXML_ERR_ERROR;
+use const LIBXML_ERR_FATAL;
+use const LIBXML_ERR_NONE;
+use const LIBXML_ERR_WARNING;
+
+const ERR_ERROR = LIBXML_ERR_ERROR;
+const ERR_FATAL = LIBXML_ERR_FATAL;
+const ERR_NONE = LIBXML_ERR_NONE;
+const ERR_WARNING = LIBXML_ERR_WARNING;
```
azjezz: I don't think there's a point of declaring these constants.

should we move them to `Xml\Error` class instead? 

however, having `Error::LEVEL_ERROR` doesn't make sense, if it's an error, why would it have a warning level instead of a error level? shouldn't that be in a different subclass of a parent class that repesent all of `Warning`, `Error`, and `FatalError`?

i think another idea would be to rename `Error`/`ErrorCollection` to `Issue`/`IssueCollection`, as `Issue::warning`/`Issue::error`/`issue::fatal` make more sense.

WDYT?

azjezz (commented):
sorry for that late review :)

composer.json
```diff
@@ -13,10 +13,12 @@
         "php": "^7.4 | ^8.0",
         "ext-bcmath": "*",
         "ext-json": "*",
+        "ext-libxml": "*",
         "ext-mbstring": "*",
         "ext-sodium": "*"
     },
     "require-dev": {
+        "ext-simplexml": "*",
```
azjezz, line 21: can you please add these extensions to all GitHub actions? the seem to be added by default, but I don't want the build to fail if configurations were to be changed in the future ( better safe than sorry 😛 )

src/Psl/Xml/Error/XmlErrorCollection.php
veewee: I don't agree : The collection is being filled by the detect function and handled by the implementing function.
So you don't know if you want to ignore warnings or not.

For XSD, you want to show them.
For loading documents, you don't really care about them - since they are non blocking.

I'dd like to keep it as a separate method if you don't mind.

src/Psl/Xml/Error/XmlError.php
veewee: If the end-user isn't using psalm, the properties can be changed.
So I'dd say it makes more sense to keep it as-is so that it cannot be abused by design.

(Otherwise we might just stick with regular LibXmlErrors - since they have public props)

WDYT?

src/Psl/Xml/constants.php
veewee: That's possible ... and results in a cleaner API. will do!

src/Psl/Xml/Error/XmlErrorCollection.php
azjezz: hm, i don't really like this method, what about `filter(...$levels): self`? that makes more sense if you want to seperate levels.

```
$all = ...;
$fatal_errors = $all->filter('fatal');
$errors = $all->filter('error');
$warnings = $all->filter('warnings');

$important = $all->filter('fatal', 'error');
```

src/Psl/Xml/constants.php
azjezz: P.S: `none` shouldn't be included, if it's 'none', it's shouldn't be part of the `IssueCollection` :)

src/Psl/Xml/Error/XmlError.php
azjezz: It's times like these I'm tempted to use https://github.com/Roave/you-are-using-it-wrong/ 😛 but want to keep it dependency-free. 

Anyway, let's stick with the getters for now.

veewee: Then you can only use it with latest PHP version ... which doesnt bother me personal but wouldnt be optimal for a package like this :)

src/Psl/Xml/Issue/Level.php
```diff
@@ -0,0 +1,126 @@
...
+        return new self(self::LEVEL_WARNING);
+    }
+
+    /**
+     * @psalm-pure
+     */
+    public static function none(): self
+    {
+        return new self(self::LEVEL_NONE);
+    }
```
azjezz: should not be considered an "issue"

src/Psl/Xml/Issue/Level.php
```diff
@@ -0,0 +1,126 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Issue;
+
+use Psl\Exception\InvariantViolationException;
+
+use function in_array;
+use function Psl\invariant;
```
azjezz:
```suggestion
use function Psl;
use function Psl\Arr;
```

src/Psl/Xml/Issue/Level.php
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * @psalm-pure
+     * @throws InvariantViolationException
+     */
+    public static function fromLevelCode(int $value): self
+    {
+        invariant(
+            in_array($value, [self::LEVEL_ERROR, self::LEVEL_FATAL, self::LEVEL_NONE, self::LEVEL_WARNING], true),
+            'Received invalid XML issue level: ' . (string) $value
+        );
```
azjezz: Psl\invariant message will be formatted using Str\foramt ( sprintf ) with the rest of the argument )
```suggestion
        Psl\invariant(
            Arr\contains([self::LEVEL_ERROR, self::LEVEL_FATAL, self::LEVEL_NONE, self::LEVEL_WARNING], $value),
            'Received invalid XML issue level "%d".',
            $value
        );
```

src/Psl/Xml/Issue/Level.php
veewee: Decided to keep none since the collection can return Level::none() as highest level if it is empty. Makes sense from a level value object POV right?

src/Psl/Xml/Issue/IssueCollection.php
```diff
@@ -0,0 +1,78 @@
...
+     * @param (callable(Issue): bool) $filter
+     */
+    public function filter(callable $filter): self
+    {
+        return new self(
+            ...Arr\filter(
+                $this->issues,
+                $filter
+            )
+        );
```
azjezz: can be in one line

src/Psl/Xml/Issue/IssueCollection.php
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * @psalm-suppress InvalidNullableReturnType
+     */
+    public function getHighestLevel(): Level
+    {
+        /** @psalm-suppress NullableReturnStatement */
```
veewee: Looks like the reduce typing doesn't work that well...
It thinks that the first argument passed to reducer can be null, but clearly it never is.
Is this something we need to fix in the reduce function? Or do I need to add something to the reducer's function docblocks in order to let psalm know it is never null?

src/Psl/Xml/Issue/IssueCollection.php
```diff
@@ -0,0 +1,78 @@
...
+        return Iter\reduce(
+            $this->issues,
+            static fn (?Level $highest, Issue $error): Level
+                => $highest && $highest->value() >= $error->level()->value() ? $highest : $error->level(),
```
azjezz: Math\max_by :)

src/Psl/Xml/Issue/IssueCollection.php
```diff
@@ -0,0 +1,78 @@
...
+
+    public function count(): int
+    {
+        return count($this->issues);
+    }
+
+    /**
+     * @param (callable(Issue): bool) $filter
+     */
+    public function filter(callable $filter): self
```
veewee, line 47: Decided to go with a regular filter callback to give maximum flexibility

src/Psl/Xml/Issue/Level.php
```diff
@@ -0,0 +1,126 @@
...
+
+use const LIBXML_ERR_ERROR;
+use const LIBXML_ERR_FATAL;
+use const LIBXML_ERR_NONE;
+use const LIBXML_ERR_WARNING;
+
+/**
+ * @psalm-immutable
+ */
+final class Level
```
veewee, line 10: Made more sense to convert the level logic into it's own VO.

src/Psl/Xml/Issue/IssueCollection.php
```diff
@@ -0,0 +1,78 @@
...
+    {
+        return Str\join(
+            Arr\values(
+                Arr\map(
+                    $this->issues,
+                    static fn (Issue $error): string => $error->toString()
+                )
+            ),
+            PHP_EOL
+        );
```
azjezz:
```suggestion
        $issues = Arr\values(Arr\map($this->issues, static fn (Issue $issue): string => $issue->toString()));
        
        return Str\join($issues, PHP_EOL);
```

src/Psl/Xml/Issue/Level.php
```diff
@@ -0,0 +1,126 @@
...
+     */
+    public static function fromLevelCode(int $value): self
+    {
+        invariant(
+            in_array($value, [self::LEVEL_ERROR, self::LEVEL_FATAL, self::LEVEL_NONE, self::LEVEL_WARNING], true),
+            'Received invalid XML issue level: ' . (string) $value
+        );
+
+        return new self($value);
+    }
```
azjezz: should be internal ( also in which case it will not throw invariant violation as it's in a controlled environment )

src/Psl/Xml/Issue/Issue.php
```diff
@@ -0,0 +1,96 @@
...
+        return new self(
+            Level::fromLevelCode($error->level),
+            $error->code,
+            $error->column,
+            $error->message,
+            $error->file,
+            $error->line,
+        );
+    }
+
```
azjezz: should be internal as well, again, we don't want to expose anything in the public API that reference the extensions used.

azjezz (changes requested):
more nitpicking 😛

src/Psl/Xml/Issue/Level.php
azjezz: why not null? indicating there are no issues to pick the highest from :)

src/Psl/Xml/Issue/IssueCollection.php
azjezz: hm, not sure, but we should use Math\max_by here.

src/Psl/Xml/Issue/IssueCollection.php
azjezz: even better :D

src/Psl/Xml/Issue/Level.php
azjezz: agree!

src/Psl/Xml/Issue/Issue.php
veewee: I am really not sure what you are expecting from me here :)
It's just a named constructor on an *internal* class and it needs to be placed somewhere ...

src/Psl/Xml/Issue/IssueCollection.php
veewee: will check it out!

src/Psl/Xml/Issue/Level.php
veewee: ... confused ... :)

What do you mean with should be internal?
It will always throw if the issue level is not know right?

src/Psl/Xml/Issue/Level.php
veewee: Isn't it strange to get `null` back from `$issues->getHighestLevel()`?

src/Psl/Xml/Issue/Issue.php
azjezz: > an internal class

the class is not internal tho :p

src/Psl/Xml/Issue/Level.php
azjezz: no, not really, if this is moved to `Psl\Xml\Internal\issue_level_from_xml(int $level): Level` we can be sure that we are always passing a valid $level so we don't need to check if it's one of XML errors.

src/Psl/Xml/Issue/Level.php
azjezz: throw instead? we have `count()` so users should probably call that before starting to check the errors and the highest level anyway.

src/Psl/Xml/Issue/Issue.php
azjezz: the function can be moved to `Psl\Xml\Internal\issue_from_xml_error` instead. and we keep this class exposed in the public API.

src/Psl/Xml/Issue/Level.php
veewee: you would most likely use it as

```php
if ($issues->getHighestLevel() > Level::warning()->value()) // ...
```

Would you prefer to

```php
try {
   if ($issues->getHighestLevel() > Level::warning()->value()) // ...
} catch () {
// ...
}
```

For no real reason? 


(Agreeing that you could use count() first ... but to make it dummy proof - returning 0 seems ok-ish ..)

src/Psl/Xml/Issue/Issue.php
veewee: ah ok that way, will try it later :)

src/Psl/Xml/Issue/Level.php
azjezz: Like: :) 
```php
if (0 === $issues->count()) {
  return;
}

if ($issues->getHighestLevel() > Level::warning()->value()) {

}
```

another thing we can do is ... not return `IssuesCollecttion` unless we do actually have issues, otherwise null? that seems sensable, so that `IssueCollection` is always on a valid state that contains issues.

( throws in the constructor if no issues are passed ... )

azjezz: but that might bring an issue with the filter method throwing ... 

also to note, that return null works with `if ($issues->getHighestLevel() > Level::warning()->value())` since null will be cast to 0 :p

veewee: In that case you should do

```php
if ($issues->getHighestLevel()?->value() > Level::warning()->value())
```

Since the issue is more something that is used internally, I really don't think adding an empty state is a bad idea.
Just for convenience. The ideas above all have there downsides:

- If you check for count 0, psalm still thinks it can throw
- if you only allow non empty, you have an issue filterting
- to be compatible with libxm, which has a none error level - otherwise the internal level detector might fail as well.

src/Psl/Xml/Internal/detect_errors.php
```diff
@@ -0,0 +1,55 @@
...
+    libxml_use_internal_errors($previousErrorReporting);
+
+    return [
+        $result,
+        new IssueCollection(
+            ...map(
+                $errors,
+                static fn(LibXMLError $error): Issue => issue_from_xml_error($error)
+            )
+        )
```
azjezz: issue_collection_from_xml_errors? might be useful somewhere else 🤔

src/Psl/Xml/Internal/issue_from_xml_error.php
```diff
@@ -0,0 +1,23 @@
...
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Internal;
+
+use LibXMLError;
+use Psl\Xml\Issue\Issue;
+
+/**
+ * @psalm-internal \Psl\Xml
```
azjezz:
```suggestion
 * @psalm-internal
 *
 * @psalm-pure
```

src/Psl/Xml/Internal/issue_level_from_xml_error.php
```diff
@@ -0,0 +1,17 @@
...
+use Psl\Xml\Issue\Level;
+
+/**
+ * @psalm-internal \Psl\Xml
+ * @psalm-suppress MissingThrowsDocblock
+ */
+function issue_level_from_xml_error(LibXMLError $error): Level
+{
+    return Level::fromLevelCode($error->level);
+}
```
azjezz:
```suggestion
/**
 * @psalm-internal
 *
 * @psalm-pure
 */
function issue_level_from_xml_error(LibXMLError $error): ?Level
{
    switch($error->level) {
        case LIBXML_ERR_ERROR:
            return Level::error();
        case LIBXML_ERR_FATAL:
            return Level::fatal();
        case LIBXML_ERR_WARNING:
            return Level::warning();
    }
    
    return null;
}
```

src/Psl/Xml/Internal/issue_from_xml_error.php
```diff
@@ -0,0 +1,23 @@
...
+function issue_from_xml_error(LibXMLError $error): Issue
+{
+    return new Issue(
+        issue_level_from_xml_error($error),
+        $error->code,
+        $error->column,
+        $error->message,
+        $error->file,
+        $error->line,
+    );
```
azjezz:
```suggestion
function issue_from_xml_error(LibXMLError $error): ?Issue
{
    $level = issue_level_from_xml_error($error);
    if (null === $level) {
        return null;
    }
 
    return new Issue(
        $level,
        $error->code,
        $error->column,
        $error->message,
        $error->file,
        $error->line,
    );
```

src/Psl/Xml/Internal/detect_errors.php
```diff
@@ -0,0 +1,55 @@
...
+use LibXMLError;
+use Psl\Exception\InvariantViolationException;
+use Psl\Result\ResultInterface;
+use Psl\Xml\Issue\Issue;
+use Psl\Xml\Issue\IssueCollection;
+
+use function libxml_clear_errors;
+use function libxml_get_errors;
+use function libxml_use_internal_errors;
+use function Psl\Arr\map;
```
azjezz:
```suggestion
use Psl\Arr;
```

src/Psl/Xml/Internal/detect_errors.php
azjezz:
```suggestion
        new IssueCollection(
            ...Arr\filter_nulls(Arr\map(
                $errors,
                static fn(LibXMLError $error): ?Issue => issue_from_xml_error($error)
            ))
        )
```

src/Psl/Xml/Issue/Level.php
```diff
@@ -0,0 +1,128 @@
...
+    public static function fromLevelCode(int $value): self
+    {
+        Psl\invariant(
+            in_array($value, [self::LEVEL_ERROR, self::LEVEL_FATAL, self::LEVEL_NONE, self::LEVEL_WARNING], true),
+            'Received invalid XML issue level: %s',
+            (string) $value
+        );
+
+        return new self($value);
+    }
```
azjezz: should be removed after the suggested changes.

src/Psl/Xml/Issue/IssueCollection.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        return new self(...Arr\filter($this->issues, $filter));
+    }
+
+    public function getHighestLevel(): Level
+    {
+        $issue = Math\max_by($this->issues, static fn(Issue $issue): int => $issue->level()->value());
+
+        return $issue ? $issue->level() : Level::none();
+    }
```
azjezz:
```suggestion
    public function getHighestLevel(): ?Level
    {
        $issue = Math\max_by($this->issues, static fn(Issue $issue): int => $issue->level()->value());

        return $issue ? $issue->level() : null;
    }
```

azjezz: I still believe this should return null or throw, have an instance of `Issue` that doesn't repesent any issue, just as a way of saying "fine" is not really a good thing.

src/Psl/Xml/Issue/Level.php
```diff
@@ -0,0 +1,128 @@
...
+use const LIBXML_ERR_WARNING;
+
+/**
+ * @psalm-immutable
+ */
+final class Level
+{
+    private const LEVEL_ERROR = LIBXML_ERR_ERROR;
+    private const LEVEL_FATAL = LIBXML_ERR_FATAL;
+    private const LEVEL_NONE = LIBXML_ERR_NONE;
```
azjezz: should be removed after the suggested changes.

src/Psl/Xml/Issue/Level.php
azjezz: > to be compatible with libxm, which has a none error level - otherwise the internal level detector might fail as well.

we don't want compatible with libxml, or any other PHP Core extension, as most of them have broken APIs already. that's the point of this library :)

> If you check for count 0, psalm still thinks it can throw

`/** @psalm-suppress MissingThrowsDocblock - collection is not empty*/` does the job in this case, we already do this tons of times within PSL, because it's valid.

azjezz (changes requested):

src/Psl/Xml/Internal/issue_from_xml_error.php
veewee: Doesnt that mean it cannot be used outside `Psl\Xml\Internal`?

azjezz: yes, but not outside `Psl`

veewee: https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-internal

Psalm needs me to provide a namespace. I'dd say `Psl\Xml` is ok, since other parts of psl don't have anything to do with xml?

azjezz: not sure how we do this everywhere else, but we don't usually supply the namespace.

azjezz: let's go with @internal for consistency ( https://github.com/azjezz/psl/blob/develop/src/Psl/Internal/Loader.php#L25 )

src/Psl/Xml/Internal/issue_level_from_xml_error.php
veewee: I gave up on the none ... you win ... :)

src/Psl/Xml/Issue/Level.php
veewee: We still need the values. Hardcoded them as numbers but checked if they matched the once from libxml inside the testsuite. That makes it libxml-independant.

src/Psl/Xml/Exception/XmlException.php
```diff
@@ -0,0 +1,29 @@
...
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Exception;
+
+use Exception;
+use Psl\Exception\RuntimeException;
+use Psl\Xml\Issue\IssueCollection;
+
+class XmlException extends RuntimeException implements ExceptionInterface
```
azjezz:
```suggestion
final class RuntimeException extends RuntimeException implements ExceptionInterface
```

src/Psl/Xml/Issue/Issue.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Issue;
+
+use Psl\Str;
+
+final class Issue
```
azjezz, line 12:
```suggestion
/**
 * @psam-immutable
 */
final class Issue
```

src/Psl/Xml/Issue/IssueCollection.php
```diff
@@ -0,0 +1,60 @@
...
+namespace Psl\Xml\Issue;
+
+use Countable;
+use IteratorAggregate;
+use Psl\Arr;
+use Psl\Iter\Iterator;
+use Psl\Math;
+use Psl\Str;
+
+final class IssueCollection implements Countable, IteratorAggregate
```
azjezz, line 19:
```suggestion
/**
 * @psalm-implements IteratorAggregate<int, Issue>
 *
 * @psalm-immutable
 */
final class IssueCollection implements Countable, IteratorAggregate
```

src/Psl/Internal/Loader.php
```diff
@@ -393,6 +398,10 @@ final class Loader
         'Psl\Json\Exception\DecodeException',
         'Psl\Json\Exception\EncodeException',
         'Psl\Hash\Context',
+        'Psl\Xml\Issue\Issue',
+        'Psl\Xml\Issue\IssueCollection',
+        'Psl\Xml\Issue\Level',
+        'Psl\Xml\Exception\XmlException',
```
azjezz:
```suggestion
        'Psl\Xml\Exception\RuntimeException',
```

azjezz (dismissed):
Perfect! I have added some suggestions, I will apply them in a patch here before merging.

Thank you! 🎉

azjezz (approved):
Thank you again for your patience 😅

veewee:
> Thank you again for your patience 😅

No problem. The result is quite satisfactory!
Can't wait to continue working on the next XML component

PR #90, opened 2020-10-17
[Hash] introduce the Hash API

### LINKED ISSUE #89: Hash API

azjezz:
closes #89

### DIFF
--- a/src/Psl/Hash/Context.php
+++ b/src/Psl/Hash/Context.php
@@ -1,0 +1,103 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Hash;
+
+use HashContext;
+use Psl;
+use Psl\Arr;
+use Psl\Str;
+
+use function hash_final;
+use function hash_init;
+use function hash_update;
+
+use const HASH_HMAC;
+
+/**
+ * Incremental hashing context.
+ *
+ * Example:
+ *
+ *      Hash\Context::forAlgorithm('md5')
+ *          ->update('The quick brown fox ')
+ *          ->update('jumped over the lazy dog.')
+ *          ->finalize()
+ *      => Str("5c6ffbdd40d9556b73a21e63c3e0e904")
+ *
+ * @psalm-immutable
+ */
+final class Context
+{
+    private HashContext $internalContext;
+
+    private function __construct(HashContext $internal_context)
+    {
+        $this->internalContext = $internal_context;
+    }
+
+    /**
+     * Initialize an incremental hashing context.
+     *
+     * @throws Psl\Exception\InvariantViolationException If the given algorithm is unsupported.
+     *
+     * @psalm-pure
+     */
+    public static function forAlgorithm(string $algorithm): Context
+    {
+        Psl\invariant(
+            Arr\contains(algorithms(), $algorithm),
+            'Expected a valid hashing algorithm, "%s" given.',
+            $algorithm,
+        );
+        $internal_context = hash_init($algorithm);
+
+        return new self($internal_context);
+    }
+
+    /**
+     * Initialize an incremental HMAC hashing context.
+     *
+     * @throws Psl\Exception\InvariantViolationException If the given algorithm is unsupported.
+     *
+     * @psalm-pure
+     */
+    public static function hmac(string $algorithm, string $key): Context
+    {
+        Psl\invariant(
+            Arr\contains(Hmac\algorithms(), $algorithm),
+            'Expected a hashing algorithms suitable for HMAC, "%s" given.',
+            $algorithm
+        );
+
+        Psl\invariant(!Str\is_empty($key), 'Expected a non-empty shared secret key.');
+
+        $internal_context = hash_init($algorithm, HASH_HMAC, $key);
+
+        return new self($internal_context);
+    }
+
+    /**
+     * Pump data into an active hashing context.
+     */
+    public function update(string $data): Context
+    {
+        /** @var HashContext $internal_context */
+        $internal_context = hash_copy($this->internalContext);
+        hash_update($internal_context, $data);
+
+        return new self($internal_context);
+    }
+
+    /**
+     * Finalize an incremental hash and return resulting digest.
+     */
+    public function finalize(): string
+    {
+        /** @var HashContext $internal_context */
+        $internal_context = hash_copy($this->internalContext);
+
+        return hash_final($internal_context, false);
+    }
+}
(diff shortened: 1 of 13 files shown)


src/Psl/Hash/Context.php
```diff
@@ -0,0 +1,72 @@
...
+    /**
+     * Finalize an incremental hash and return resulting digest.
+     */
+    public function finalize(): string
+    {
+        /** @var HashContext $internal_context */
+        $internal_context = hash_copy($this->internalContext);
+
+        return hash_final($internal_context, false);
+    }
```
azjezz, line 102: @veewee what do you think about this?

Finalizing the context here doesn't free the internal hash context, since this object is immutable, it makes sense from the type system perspective, you can be sure that passing this object around will never result in any modifications, however, but I'm not really sure 🤔 with that method name, it might be expected that the context is now freed from the memory.

do you think we should free the internal context at this stage? should we rename this method?

src/Psl/Hash/Hmac/hash.php
```diff
@@ -0,0 +1,24 @@
...
+use function hash_hmac;
+
+/**
+ * Generate a keyed hash value using the HMAC method.
+ *
+ * @throws Psl\Exception\InvariantViolationException If the given algorithm is unsupported.
+ *
+ * @psalm-pure
+ */
+function hash(string $data, string $algorithm, string $key): string
```
azjezz, line 17: The reason I flipped the arguments here is to "match" the signature of `Password\hash` ( even tho $algorithm has a default value there ):

consistent:
```php
$hash = Password\hash($password, $algorithm);
$hash = Hash\hash($some_data, $algorithm);
```

inconsistent:

```php
$hash = Password\hash($password, $algorithm);
$hash = Hash\hash($algorithm, $some_data);
```

src/Psl/Hash/Context.php
```diff
@@ -0,0 +1,72 @@
...
+        return new self($internal_context);
+    }
+
+    /**
+     * Pump data into an active hashing context.
+     */
+    public function update(string $data): Context
+    {
+        /** @var HashContext $internal_context */
+        $internal_context = hash_copy($this->internalContext);
```
veewee, line 87: You are creating an immutable context class with this approach.
But I am wondering : does this come at a big memory cost if the data inside the internal context is quite big?

src/Psl/Hash/Context.php
veewee: Not sure ...

Since you copy the internal context first, the final call will always succeed.
I don't think you should free the context, because that will result in an unusable Context class.

Because it is immutable, you can finalize all intermediate context states as well.
From that point of view, it would make sense not to nuke the internal context resource either.

On the other hand, there is also the memory question from above that comes to mind.


I am not really sure what the best approach is at this moment.
Maybe we can make a descision based on the anser to the memory question?

src/Psl/Hash/Hmac/Context.php
```diff
@@ -0,0 +1,72 @@
...
+    {
+        $this->internalContext = $internal_context;
+    }
+
+    /**
+     * Initialize an incremental hashing context.
+     *
+     * @throws Psl\Exception\InvariantViolationException If the given algorithm is unsupported.
+     */
+    public static function create(string $algorithm, string $key): Context
```
veewee: Would it make sense to create 1 hashing context with 2 named constructors?

- One for regular hashes: `Context::forAlgorithm(...)`
- One for hmac hashes: `Context::hmac()`

The rest seems similar.

veewee: Maybe best to switch algorithm and key in here as well for consistency?

src/Psl/Hash/Hmac/hash.php
veewee: I'dd keep it consistent. Those are things that can be prefilled by the IDE as well...
What bothers me with those hashing functions is what algorithms are available.

Sure, one can look it up - but would it make sense to create shortcut functions per algorithm?
Isn't that list more or less fixed?

veewee (commented):
Looks nice. I don't have a lot of experience with those incremental hashing functions, but tried answering your questions.

src/Psl/Hash/Hmac/hash.php
azjezz: not really, PHP std stopped doing that function-per-algo thing a while ago, every few months PHP will add support for a couple of algorithms, this can get a bit messy, there's `Hash\algorithmes` function, we can add documentation of the currently available algorithms there i guess.

src/Psl/Hash/Context.php
azjezz: if you are leaving a reference to the `Context` class somewhere, yes ... but if the object is destructed, the internal context will be freed. meaning no issues.

i assume its mostly used as follow:
```php
$context = Context::create('sha1');
foreach($block = $reader->read($block_size)) {
  // old $context is destructedas the new one overrides it.
  $context = $context->update($block);
}

$hash = $context->finalize();
// there only 1 reference to `Context` left ( $context )
```

src/Psl/Hash/Hmac/Context.php
azjezz: not sure 🤔 it made sense to have it in a separate `Hmac` namespace with other hmac functions

src/Psl/Hash/Hmac/hash.php
azjezz: full list: https://www.php.net/manual/en/function.hash-algos.php

veewee: ok thats a big list - makes sense not to include it

src/Psl/Hash/Context.php
veewee: Assuming the context will most likely only live inside a function that returns the hash string, I wouldn't care too much about freeing up the resource.

src/Psl/Hash/Hmac/Context.php
azjezz: `Hmac\Context` has been merged with `Hash\Context` as suggested 👍🏼

src/Psl/Hash/Context.php
azjezz: exactly my thought.

I have updated `hash` functions to use context as well ( avoding `hash`/`hash_hmac` builtin functions ).

veewee (approved):
Looks good! Nice job!

PR #95, opened 2020-10-30
[Experimental][Xml] Reader

veewee:
This PR Contains some example interfaces for providing an XML reader.
There are some limitations to php's `XMLReader` regarding to what input it can handle.

I don't really like the `ReaderResourceInterface` that much, but don't have a good alternative besides named constructors directly on the reader. 

The PR also contains the `stop_on_first_issue` function I shared earlier so that we can discuss it as well.

### DIFF
--- a/src/Psl/Xml/Internal/stop_on_first_issue.php
+++ b/src/Psl/Xml/Internal/stop_on_first_issue.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Xml\Internal;
+
+use Generator;
+use Psl\Arr;
+use Psl\Xml\Exception\RuntimeException;
+
+use function libxml_clear_errors;
+use function libxml_get_errors;
+use function libxml_use_internal_errors;
+
+/**
+ * Additional logic needs to be placed on top of the detect_errors function if you want to lazily detect errors.
+ * This method can be used while reading an \XMLReader and returning a Generator.
+ * After every tick, this function will validate if issues were detected.
+ *
+ * @internal
+ *
+ * @template Tr
+ *
+ * @psalm-param (callable(): bool) $tick
+ * @psalm-param (callable(): ?Tr) $run
+ * @psalm-return Generator<int, Tr, null, mixed>
+ *
+ * @throws RuntimeException
+ */
+function stop_on_first_issue(callable $tick, callable $run): Generator
+{
+    $previousErrorReporting = libxml_use_internal_errors(true);
+    libxml_clear_errors();
+
+    $detectIssues = static function (): void {
+        $issues = issue_collection_from_xml_errors(Arr\values(libxml_get_errors()));
+        if ($issues->count()) {
+            throw RuntimeException::fromIssues('Detected issues during the parsing of the XML Stream', $issues);
+        }
+    };
+
+    try {
+        while ($tick()) {
+            $result = $run();
+            if ($result !== null) {
+                yield $result;
+            }
+
+            $detectIssues();
+        }
+
+        $detectIssues();
+    } finally {
+        libxml_clear_errors();
+        libxml_use_internal_errors($previousErrorReporting);
+    }
+}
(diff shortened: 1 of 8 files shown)


src/Psl/Xml/Internal/stop_on_first_issue.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @template Tr
+ *
+ * @psalm-param callable(): bool $tick
+ * @psalm-param callable(): Tr|null $read
+ * @psalm-return Generator<int, Tr, mixed, null>
+ *
+ * @throws RuntimeException
+ */
+function stop_on_first_issue(callable $tick, callable $run): Generator
```
veewee, line 30: Example usage:

```php

<?php
$reader = new XMLReader();
$reader->open('users.xml');

$iterator = stop_on_first_issue(
    static fn(): bool => $reader->read(),
    static fn(): ?string => $reader->name === 'user' ? $reader->readOuterXml() : null,
);

// This will not report any issues and trigger warnings, because of the Generator
// If you replace the yield with e.g. an array, it will work
foreach ($iterator as $xml) {
    echo $xml;
}
```

src/Psl/Xml/Reader/ReaderResourceInterface.php
```diff
@@ -0,0 +1,56 @@
...
+ * @see https://www.php.net/manual/en/xmlreader.xml.php
+ */
+interface ReaderResourceInterface
+{
+    public static function fromXml(string $xml): self;
+
+    /**
+     * It is important this function checks if a file exists.
+     * Otherwise it is possible to read a remote url, which is not optimal.
+     * Mainly because remote url's require additional stream contexts etc.
```
veewee, line 30: see https://www.php.net/manual/en/function.libxml-set-streams-context.php

src/Psl/Xml/Reader/ReaderResourceInterface.php
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Therefore, it is not possible to use a IO ReadHandle
+ * unless we do some bad stuff like streaming it to a temp file and read that file.
+ *
+ * This resource interface could be removed if we added these named constructors on the reader interface.
+ * Both are not optimal .. but neither is life!
+ *
+ * @see https://www.php.net/manual/en/xmlreader.open.php
+ * @see https://www.php.net/manual/en/xmlreader.xml.php
+ */
```
veewee, line 22: FYI : The same goes for DOMDocument and simplexml : they can only load a complete XML string or a URI.
So maybe better to move this resource one namespace up

src/Psl/Xml/Reader/ReaderResourceInterface.php
```diff
@@ -0,0 +1,56 @@
...
+     * @throws FileNotFoundException
+     */
+    public static function fromFile(string $file): self;
+
+    /**
+     * Downside : This would leak the XMLReader from libxml... Which sucks
+     * Other option are
+     *
+     *  - configure(XMLReader) which would change an existing resource in a mutable way - not optimal
+     *  - getUri(): ?string & getXml(): ?string -> not optimal either ...
```
veewee, line 41: Probably better if we will be using this resource for other sections of the xml namespace as well. 
Nevertheless, a strange API ...

src/Psl/Xml/Internal/stop_on_first_issue.php
```diff
@@ -0,0 +1,57 @@
...
+        while ($tick()) {
+            $result = $run();
+            if ($result !== null) {
+                yield $result;
+            }
+
+            $detectIssues();
+        }
+
+        $detectIssues();
```
veewee, line 52: In case initial $tick failed already.

azjezz:
needs rebase

azjezz:
"rebased" 😆

veewee:
Closing this one for now.
Implemented an improved version of the reader in a separate XML package.

https://github.com/veewee/xml/tree/master/src/Xml/Reader

PR #101, opened 2020-11-07
 [Type] replace `Type\arr` by `Type\{mutable_}map` and `Type\{mutable_}vector`

### LINKED ISSUE #100: [Type] Missing way of declaring map/vector types
Currently there is no type for specifying e.g. `Type\list(Type\string())`.
Example use case:

```
$result = Json\typed($json, Type\vector(Type\string())->toArray();
```

azjezz:
closes #100


veewee:
Was thinking : does it make sense to remove the array type altogether? In some situations in which you dont have that much control over de source data, you might be ok with having a very loose regular `mixed` array.

tests/Psl/Json/TypedTest.php
```diff
@@ -5,28 +5,44 @@
...
+    public function testTypedVector(): void
+    {
+        /** @var MapInterface $actual */
+        $actual = Json\typed(
+            '["php", "std", "stdlib", "utility", "psl"]',
+            Type\vector(Type\string())
+        );
+
+        static::assertInstanceOf(VectorInterface::class, $actual);
+        static::assertSame(['php', 'std', 'stdlib', 'utility', 'psl'], $actual->toArray());
```
azjezz: cc @veewee 😉

azjezz:
> Was thinking : does it make sense to remove the array type altogether? In some situations in which you dont have that much control over de source data, you might be ok with having a very loose regular mixed array.

that is just `Type\map(Type\array_key(), Type\mixed())->coerce($var)->toArray()`

azjezz:
but i think we can add `Type\untyped_array` if you are up for a PR 😄

veewee:
Sure, Will see what I can do. Hopefully at the end of the week :)

src/Psl/Type/Internal/MutableMapType.php
```diff
@@ -0,0 +1,154 @@
...
+                /** @psalm-var Tk $k */
+                $k = $key_type->coerce($k);
+                /** @psalm-var Tv $v */
+                $v = $value_type->coerce($v);
+
+                $entries[] = [$k, $v];
+            }
+
+            /** @psalm-var Iter\Iterator<Tk, Tv> $iterator */
+            $iterator = Iter\from_entries($entries);
```
veewee, line 87: Would it make sense not to use entries but build the map up from the foreach directly? This would allow for pure / immutable type classes (since `from_entries($iterable)` is not pure)

(same goes for the other usages of entries)

veewee (approved):
Looks good! Added a small remark  / question.
Gonna try to open up a PR for maing the type layer pure as well. Still got some issues there.

src/Psl/Type/Internal/MutableMapType.php
azjezz: it woudln't make any differencce here, since coerce accepts any iterable, and does a foreach on it, meaning it can never be pure. as that iterable we are doing foreach on isn't really immutable.

azjezz:
>  Gonna try to open up a PR for maing the type layer pure as well. Still got some issues there.

not sure if that's possible, but we will see, i'm merging this as-is for now.

veewee:
> Gonna try to open up a PR for maing the type layer pure as well. Still got some issues there.

> not sure if that's possible, but we will see, i'm merging this as-is for now.

Put some more time in it and it turns out it is not possible ...
As long as we cannot declare immutable iterables, this means it won't be possible unless we change everything to a regular array, which doesn't seem optimal to me in this case.

PR #102, opened 2020-11-08
[Str] add `after` and `before` helpers

### LINKED ISSUE #75: [Str] add `after` and `before` helpers
add `after{_ci|_last{_ci}}` and `before{_ci|_last{_ci}}` helper string functions.

We should implement these functions for `Psl\Str`, `Psl\Str\Byte`, and `Psl\Str\Grapheme` ( refs #74 ), which totals to a new 24 functions.

`Psl\Str` implementations would have an extra argument `?string $encoding = null`.

```php
function after(string $haystack, string $needle, bool $include_needle = false, int $offset = 0): string;

function after_ci(string $haystack, string $needle, bool $include_needle = false, int $offset = 0): string;

function after_last(string $haystack, string $needle, bool $include_needle = false, int $offset = 0): string;

function after_last_ci(string $haystack, string $needle, bool $include_needle = false, int $offset = 0): string;

function before(string $haystack, string $needle, bool $include_needle = false, int $offset = 0): string;

function before_ci(string $haystack, string $needle, bool $include_needle = false, int $offset = 0): string;

function before_last(string $haystack, string $needle, bool $include_needle = false, int $offset = 0): string;

function before_last_ci(string $haystack, string $needle, bool $include_needle = false, int $offset = 0): string;
```

azjezz:
closes #75

TODO:
- [x] implement `before{_ci|_last{_ci}}` functions `Str\`
- [x] implement `after{before{_ci|_last{_ci}}` functions for `Str\`
- [x] implement `search_last_ci` function for `Str\Byte\`
- [x] implement `before{_ci|_last{_ci}}` functions `Str\Byte\`
- [x] implement `after{before{_ci|_last{_ci}}` functions for `Str\Byte\`
- [x] implement `before{_ci|_last{_ci}}` functions `Str\Grapheme\`
- [x] implement `after{before{_ci|_last{_ci}}` functions for `Str\Grapheme\`
- [x] write tests ( keep coverage at 100% )
- [ ] write documentation


src/Psl/Str/Byte/after.php
```diff
@@ -0,0 +1,26 @@
...
+namespace Psl\Str\Byte;
+
+use Psl;
+
+/**
+ * @throws Psl\Exception\InvariantViolationException If the $offset is out-of-bounds.
+ *
+ * @psalm-pure
+ */
+function after(string $haystack, string $needle, bool $include_needle = false, int $offset = 0): string
```
veewee: Not sure about the `include_needle`. Wont you expect it to be always false because you are searching for a string *after* another string?
What would be the use-case?

Maybe better if you create variants like `after_with_needle` instead of providing a boolean option that makes the function behave differently?

azjezz: i agree with you, i have remove the `$include_needle` argument, and instead of returning the $hystack when $needle is not found, now null is returned.

i don't think `after_with_needle` is needed. as it would make things a bit more complicated, and it's the same as:

```php
$result = Str\after($haystack, $needle);
$result = null === $result ? $needle.$result : null;
```

PR #111, opened 2021-01-18
[Math] clamp function

veewee:
This PR adds a `Math\clamp($num, $min, $max)` function that limits a number to a given range.

### DIFF
--- a/src/Psl/Math/clamp.php
+++ b/src/Psl/Math/clamp.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Math;
+
+use Psl;
+
+/**
+ * Returns a number whose value is limited to the given range.
+ *
+ * @template T of float|int
+ *
+ * @param T $number
+ * @param T $min
+ * @param T $max
+ *
+ * @return T
+ * @throws Psl\Exception\InvariantViolationException If min is bigger than max
+ *
+ * @psalm-pure
+ */
+function clamp($number, $min, $max)
+{
+    Psl\invariant($min <= $max, 'Expected $min to be lower or equal to $max.');
+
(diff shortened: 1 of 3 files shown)


src/Psl/Math/clamp.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Returns a number whose value is limited to the given range.
+ *
+ * @template TNum of float|int
+ *
+ * @param TNum $number
+ * @param TNum $min
+ * @param TNum $max
+ *
+ * @return TNum
```
azjezz:
```suggestion
 * @template T of float|int
 *
 * @param T $number
 * @param T $min
 * @param T $max
 *
 * @return T
```

src/Psl/Math/clamp.php
```diff
@@ -0,0 +1,36 @@
...
+ * @param T $max
+ *
+ * @return T
+ * @throws Psl\Exception\InvariantViolationException If min is bigger than max
+ *
+ * @psalm-pure
+ */
+function clamp($number, $min, $max)
+{
+    Psl\invariant($min <= $max, 'Expected %s to be lower or equal to %s.', $min, $max);
```
azjezz: we should avoid having parameters added to the exception message ( as they can be dynamic and you don't really want dynamic messages in logs as it would be hard to group them together ) 

ref: https://github.com/azjezz/psl/blob/1.0.x/src/Psl/Math/base_convert.php#L47

```suggestion
    Psl\invariant($min <= $max, 'Expected $min to be lower or equal to $max.');
```

azjezz (dismissed):

azjezz (dismissed):

azjezz (approved):

PR #112, opened 2021-01-18
[Iter][Arr] flat_map function

veewee:
This function introduces a `flat_map` function to both `Arr` and `Iter`.


src/Psl/Arr/flat_map.php
```diff
@@ -0,0 +1,35 @@
...
+
+use Psl\Type;
+
+/**
+ * @psalm-template Tk of array-key
+ * @psalm-template Tv
+ * @psalm-template T
+ *
+ * @psalm-param iterable<Tk, Tv>                $iterable Iterable to be mapped over
+ * @psalm-param (callable(Tv): (T|iterable<T>)) $mapper
```
azjezz:
```suggestion
 * @psalm-param (callable(Tv): iterable<T>) $mapper
```

The issue is if T itself is an array or an iterable, we can't really determine if the return value is a single entry or a list.

src/Psl/Arr/flat_map.php
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @psalm-return list<T>
+ */
+function flat_map(iterable $iterable, callable $mapper): array
+{
+    $flattened = [];
+    foreach ($iterable as $value) {
+        $result = $mapper($value);
+        if (Type\is_iterable($result)) {
```
azjezz: what if `T` is `array<string>` ? :)

src/Psl/Iter/flat_map.php
```diff
@@ -0,0 +1,34 @@
...
+function flat_map(iterable $iterable, callable $mapper): iterable
+{
+    return Iterator::from(static function () use ($iterable, $mapper) {
+        foreach ($iterable as $value) {
+            $result = $mapper($value);
+            if (Type\is_iterable($result)) {
+                /** @var T $item */
+                foreach ($result as $item) {
+                    yield $item;
+                }
```
azjezz: same as above.

src/Psl/Arr/flat_map.php
veewee: I wanted it to be close to the JS version of flat map, but they have separate syntax for arrays and objects.
A use-case for flatmaps could be this as well:

```php
flat_map(
   [
       fn () => $node1,
       fn () => [$node2, $node3],
   ],
   fn ($builder) => $builder()
)
```

Imagine the initial list of items are composable builders that you'dd also want to use seperately. In that case you might not want to force them all to return iterables. But you do want a node list as result.

src/Psl/Arr/flat_map.php
veewee: then it would be flattened into `list<string>`, which kinda makes sense ...
If T were a map, it will be converted to vec and flattened as well...

Not sure how we should be able to deal with that?

azjezz: according to the function signature, it shouldn't.

here's a case where this will fail hard:

```hack
flat_map<T>(list<T> $list, (function(): T|iterable<T>) $callable): list<T>;

/** @var list<list<int>> $list */
$list = [$users_ids, $another_list_of_users_ids];
$callable = function(list<int> $user_ids): list<User>|list<list<User>> {
  /** @return list<User> */
  return $database->findAllUsers($user_ids);
}

/** list<list<User>> */
$user_groups = flat_map($list, $callable);

foreach($user_groups as $group) {
  foreach($group as $user) { // ERROR: cannot call foreach on `User`

  }
}
```

veewee: Isn't that what you want to avoid with the "flat" part of `flat_map` anyways?
You want to load 2 groups of users, but want a flat list of users as output from the function.
Otherwise you might as well just use regular map?

azjezz: yes, but according to psalm ( which is correct according to the signature ), the function returns `list<list<User>>`, not `list<User>`.

veewee: In that specific case, the `list<list<User>>` annotation on $callable is wrong.
See: https://psalm.dev/r/75ad777d4a

It knows that in that case the output is `list<User>`

Or do I understand your snippet incorrectly?

azjezz: somehow this doesn't feel right to me.
what is the difference between this and `Arr\flatten(Arr\map($user_ids, $mapper))`?

AFAIK, the only diffenet is that flatten takes `list<list<T>>` and returns `list<T>`, while `flat_map` takes `list<list<T>>|list<T>` and returns `list<T>`, right?

veewee: `Arr\flatten` just doesnt work in this case since it merges based on key. See #113
The difference between map and flat_map is as followed:

- map converts one input item into one output item
- flatmap converts one input item into zero or more output items per input item. The result is the combination of all the results.


So in types, the difference is in the mapper return signature:

```hack
map(list<a>, (a): b): list<b>
flat_map(list<a>, (a): b|list<b>): list<b>
```

The input will always be a flat list.
However, the mapper inside flat_map can return another list that needs to be flattened into the output of `list<b>`.


I think the thing you are worried about is the `b|list<b>` return of the mapper for flat_map? It might make sense to return a `list<b>` instead. That way the typing and implementation is more straight forward. Would that solve the issue you mention or am I missing something?

azjezz: yes, let's avoid having union types here. return `iterable<T>` 👍

azjezz (approved):

PR #115, opened 2021-01-29
[Iter] Make reducers consistent

veewee:
Whilst using the `reduce_with_keys` function, I noticed it has a different signature then the regular `reduce` function.
I removed the default null value to make it consistent with the changes from #98.

### DIFF
--- a/src/Psl/Iter/reduce_keys.php
+++ b/src/Psl/Iter/reduce_keys.php
@@ -24,13 +24,13 @@ namespace Psl\Iter;
  * @psalm-template Tv
  * @psalm-template Ts
  *
- * @psalm-param iterable<Tk, Tv>        $iterable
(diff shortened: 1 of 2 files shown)


azjezz:
Sadly, this is now a BC break 😭

we are going to have to deprecate passing null, and remove it in 2.x, however, I dislike the way PHP triggers deprecations, so we are going to have to work on an `Error` component 😄

veewee:
Yeah, that's correct but a bit sucky :)

The big issue from static analysis pov is that psalm always forces you to keep null in mind, even though you pass an initial value. Resulting in false positives.

I can fix it with stub files, so it's not that blocking though.

azjezz:
considering that both these functions are now deprecated, and the replacement already requires the `$initial`, this can be closed as they will be removed in 2.0.

PR #116, opened 2021-02-01
[Type] add non-empty-string type

azjezz:


src/Psl/Type/Internal/NonEmptyStringType.php
```diff
@@ -0,0 +1,72 @@
...
+    public function coerce($value): string
+    {
+        if (Type\is_string($value) && !Str\is_empty($value)) {
+            return $value;
+        }
+
+        if (Type\is_int($value)) {
+            $str = (string) $value;
+            if (!Str\is_empty($str)) {
+                /** @var non-empty-string $str */
```
azjezz, line 35: see https://github.com/vimeo/psalm/issues/5141

PR #119, opened 2021-02-06
[Type] add shape

azjezz:
This patch adds a new `Type\shape` function to verify object-like-arrays at runtime, as well as a psalm plugin (`Psl\Internal\Psalm\Plugin` ) to help with typing, as we can't really type this function statically.

currently, `vendor/bin/psalm-plugin` is unable to discover this plugin as the package type is not `psalm-plugin`, however, people can enable it manually as mentioned in the README. ( see  https://github.com/vimeo/psalm/pull/5165 )

```php
try {
  // without plugin: array<string, string>
  // with plugin: array{'username': string, 'password': string}
  $credentials = Type\shape(
    'username' => Type\string(),
    'password' => Type\string(),
  )->assert($_REQUEST['credentials']);

  // safe
  $username = $credentials['username'};
  $password = $credentials['password'];

  // ...

} catch(Type\Exception\AssertException $e) {
   // $_REQUEST['credentials'] is not array{'username': string, 'password': string}
}
```


README.md
```diff
@@ -50,6 +50,23 @@ Supported installation method is via [composer](https://getcomposer.org):
...
+```xml
+<psalm>
+   ...
+   <plugins>
+      ...
+      <pluginClass class="Psl\Internal\Psalm\Plugin"/>
+   </plugins>
+</psalm>
+
+```
```
azjezz, line 68: if https://github.com/vimeo/psalm/pull/5165 gets merged, we can recommend using `vendor/bin/psalm-plugin`, as currently it is unable to discover the plugin class.

azjezz:
@veewee do you have any remarks before merging?

veewee:
In general, it looks good. Gonna take a deeper look after the weekend to make sure I understand everything.

README.md
azjezz: also, we would probably wanna keep this for sometime as the change will only make it to future psalm versions.

src/Psl/Internal/Psalm/EventHandler/ShapeFunctionReturnTypeProvider.php
```diff
@@ -0,0 +1,85 @@
...
+
+declare(strict_types=1);
+
+namespace Psl\Internal\Psalm\EventHandler;
+
+use Psalm\Plugin\EventHandler\Event\FunctionReturnTypeProviderEvent;
+use Psalm\Plugin\EventHandler\FunctionReturnTypeProviderInterface;
+use Psalm\Type;
+
+final class ShapeFunctionReturnTypeProvider implements FunctionReturnTypeProviderInterface
```
veewee, line 11: I am wondering : why do we need to add a psalm plugin?
Psalm knows about object arrays, but If I understand it correctly, you add support for edge cases with this provider?
That's not something you can do with the regular type?

Or does it has to do with the extend of ` @extends Type\Type<array<Tk, Tv>>` on shape type level?

src/Psl/Type/Internal/ShapeType.php
```diff
@@ -0,0 +1,133 @@
...
+
+        throw AssertException::withValue($value, $this->toString(), $this->getTrace());
+    }
+
+    /**
+     * Returns a string representation of the shape.
+     */
+    public function toString(): string
+    {
+        $nodes = [];
```
veewee, line 122: You could use `map` to avoid this intermediate variable building up
and use the format function to build the node itself.

Or you could just keep this :)

README.md
veewee: So for now, if you want the types to complete for the shape type, you need to manually register the plugin?
That is indeed not optimal. But I see that the PR is merged already so just a small bump on the road.

It might be better not to have to provide a plugin though ... If that were possible.

src/Psl/Internal/Psalm/EventHandler/ShapeFunctionReturnTypeProvider.php
azjezz: here's a case where it makes different with/without plugin:

```
$composer = Json\decode(file_get_contents('composer.json'), Type\shape(
  'name' => Type\string(),
  'description' => Type\string(),
  'require' => Type\map(Type\string(), Type\string()),
  'require-dev' => Type\map(Type\string(), Type\string()),
  'scripts' => Type\map(Type\string(), Type\union(Type\string(), Type\vector(Type\string()),
);

$scripts = $composer['scripts'];

foreach($scripts as $script) {
  if (Type\is_string($script)) {
    $commands = [$script];
  } else {
    $commands = $script->toArray();
  }

  //
}
```

without the plugin, psalm would think `$composer` is now `array<string, string|MapInterface<string, string>|MapInterface<string, string|VectorInterface<string>>`, meaning with `$scripts = $composer['scripts']`, psalm thinks `$scripts` is `string|MapInterface<string, string>|MapInterface<string, string|VectorInterface<string>`, however, if you have the plugin enabled, the type of `$composer` would be `array{name: string, description: string, require: MapInterface<string, string>, require-dev: MapInterface<string, string>, scripts: MapInterface<string, string|VectorInterface<string>>`, so when you do  `$scripts = $composer['scripts']`, psalm will know that `$scripts` is now `MapInterface<string, string|VectorInterface<string>>`

veewee: Allright, makes sense!

veewee (approved):

Issue #120, opened 2021-02-07
refactoring Arr and Iter


azjezz:
`Arr` and `Iter` components are currently confusing, with some functions being pure, and other not, some returning lists, dicts, or just a value/key, or something else.

The idea would be:

1. `Iter` would contain generic array functions that operate on all iterables, and usually don't return an iterable ( e.g: Iter\first, Iter\count, Iter\contains, Iter\contains_key .. etc )
2. a new component `Vec` should contain every function that returns a vec ( i.e list<T> ) from both `Arr` and `Iter`.
3. a new component `Dict` should contain every function that returns a dict ( i.e array<Tk, Tv> ) from both `Arr` and `Iter`
4. deprecate `Arr` component ( all functions will be move to either `Iter`, `Vec`, or `Dict`  )

This also means that we lazy loaded iterators, however, i don't see that as an issue, most functions will still accept iterables, so you can provide a generate, and wrap the result in a generator your self.

However, this doesn't solve the `pure` issue, where functions are not considered pure, but after #119, we can extend the psalm plugin to make it so `Vec\sort($array)` considered pure, while `Vec\sort($db->getUsers())` ( where `$db->getUsers()` returns a mutable traversal ) considered impure.

azjezz:
deprecation will take place in 1.2, and the removal of old functions will happen in 2.0.

azjezz:
done in #241

PR #121, opened 2021-02-07
[Vec] Introduce vec component

azjezz:
related to #120 

TODO:

- [x] add basic Vec functions
- [x] deprecate duplicates in Iter\ and Arr\
- [x] write tests

changelog:

Features:
- added `Vec\chunk`
- added `Vec\chunk_with_keys`
- added `Vec\concat`
- added `Vec\fill`
- added `Vec\filter_nulls`
- added `Vec\flat_map`
- added `Vec\keys`
- added `Vec\values`
- added `Vec\partition`
- added `Vec\range`
- added `Vec\reductions`
- added `Vec\reproduce`
- added `Vec\reverse`
- added `Vec\shuffle`
- added `Vec\sort`
- added `Vec\sort_by` 
- added `Vec\values`
- added `Vec\zip`

Deprecation:
- deprecated `Iter\to_array` ( use `Vec\values` instead )
- deprecated `Iter\values` ( use `Vec\values` instead )
- deprecated `Arr\concat` ( use `Vec\concat` instead )
- deprecated `Arr\count` ( use `Iter\count` instead )
- deprecated `Arr\filter_nulls` ( use `Vec\filter_nulls` instead )
- deprecated `Arr\flat_map` ( use `Vec\flat_map` instead )
- deprecated `Arr\keys` ( use `Vec\keys` instead )
- deprecated `Arr\shuffle` ( use `Vec\shuffle` instead )
- deprecated `Arr\sort` ( use `Vec\sort` instead )
- deprecated `Arr\sort_by` ( use `Vec\sort_by` instead )
- deprecated `Arr\values` ( use `Vec\values` instead )
- deprecated `Iter\chunk` ( use `Vec\chunk` instead )
- deprecated `Iter\chunk_with_keys` ( use `Vec\chunk_with_keys` instead )
- deprecated `Iter\filter_nulls` ( use `Vec\filter_nulls` instead )
- deprecated `Iter\flat_map` ( use `Vec\flat_map` instead )
- deprecated `Iter\keys` ( use `Vec\keys` instead )
- deprecated `Iter\range` ( use `Vec\range` instead )
- deprecated `Iter\reductions` ( use `Vec\reductions` instead )
- deprecated `Iter\repeat` ( use `Vec\fill` instead )
- deprecated `Iter\reproduce` ( use `Vec\reproduce` instead )
- deprecated `Iter\reverse` ( use `Vec\reverse` instead )
- deprecated `Iter\zip` ( use `Vec\zip` instead )


src/Psl/Arr/random.php
```diff
@@ -21,12 +23,13 @@
  */
 function random(array $values)
```
veewee, line 24: Is it the goal to remove the `Arr\` namespace at some point?
Because this one, and some others, doesn't have an alternative version.
Maybe better to create/use a Iter\random that can operate on both vector and dict?

Now that I am typing this, I notice this PR adds vec only. So that move is probably a next stage :)

src/Psl/Arr/values.php
```diff
@@ -16,6 +18,10 @@
  * @psalm-return   list<T>
  *
  * @psalm-pure
+ *
+ * @deprecated since 1.2, use Vec\values instead.
+ *
+ * @see Vec\values()
```
veewee, line 24: Is there a specific need for Vec\values or can we enforce vec to always be a `list` type instead of `array<int, ...>` ?

src/Psl/Iter/reverse.php
```diff
@@ -19,12 +19,17 @@
...
+ *
+ * @see Vec\reverse()
  */
 function reverse(iterable $iterable): Iterator
 {
     return Iterator::from(static function () use ($iterable): Generator {
-        $values = to_array($iterable);
-        $size   = Arr\count($values);
+        $values = Vec\values($iterable);
+        $size   = namespace\count($values);
```
veewee, line 31: Any specific reason to add the explicit namespace keyword here? Doesn't it fall back to the count in namespace anyways?

src/Psl/Vec/chunk_with_keys.php
```diff
@@ -0,0 +1,44 @@
...
+ * @template Tv
+ *
+ * @param iterable<Tk, Tv> $iterable The iterable to chunk
+ * @param int $size The size of each chunk
+ *
+ * @return list<array<Tk, Tv>>
+ *
+ * @throws Psl\Exception\InvariantViolationException If $size is negative.
+ */
+function chunk_with_keys(iterable $iterable, int $size): array
```
veewee, line 27: Should this be a vec? The result indeed is a list, but it operates on a dict input?

veewee (commented):
Looks good. Added some small questions / remarks!

src/Psl/Arr/random.php
azjezz: exactly
- add vec\
- add dict\
- completely deprecate arr\ in favor of iter\ that can operate on all iterables
- add feature in the psalm plugin that allows Iter\count($array) to be called in a pure context, since it's not mutating anything

src/Psl/Vec/chunk_with_keys.php
azjezz: we don't care about what we operate on, we care about the return type.

If return type is vec or pair of vecs, it should be in Vec\

src/Psl/Iter/reverse.php
azjezz: so that if Iter\count wasn't loaded ( assuming someone is trying to require_once the files themselves instead of using the loader ), it will fail rather than falling back to the global count.

generally, we want to avoid this weird behavior in PHP were it falls back to using another function declaration when one is not found, so we should use either `Component\func`, `namespace\func`, or `\func` for global functions ( we don't enforce this currently, but we probably should ).

src/Psl/Arr/values.php
azjezz: the return type here is what matters, you would use Vec\values when you want to create a *Vec* from *values* of an iterable.

azjezz: same for Vec\keys where you want to create a Vec from the keys of an iterable.

src/Psl/Iter/chunk.php
```diff
@@ -26,27 +27,15 @@
  * @psalm-return Iterator<int, list<T>>
  *
  * @throws Psl\Exception\InvariantViolationException If $size is negative.
+ *
+ * @deprecated since 1.2, use Vec\chunk instead.
```
ragboyjr, line 31: Curious, why is the iterable version deprecated? When processing a large data stream, it be can be useful to have a lazy chunk fn that doesn't consume the entire data set into memory.

PR #127, opened 2021-02-16
[IO] Introduce IO component

azjezz:


src/Psl/IO/Reader.php
```diff
@@ -0,0 +1,142 @@
...
+
+    private bool    $eof = false;
+    private string  $buffer = '';
+
+    public function __construct(ReadHandle $handle)
+    {
+        $this->handle = $handle;
+    }
+
+    public function getHandle(): ReadHandle
```
veewee, line 23: Maybe better not to provide the `ReadHandle` again from the reader?
You already have the handle whilst constructing if you really want to do something with it in-between reading.

src/Psl/IO/input_handle.php
```diff
@@ -0,0 +1,22 @@
...
+
+use function php_sapi_name;
+
+/**
+ * Return the input handle for the current request.
+ *
+ * In CLI mode, this is likely STDIN; for HTTP requests, it may contain the
+ * POST data, if any.
+ */
+function input_handle(): ReadHandle
```
veewee, line 17: Really like these functions! :)

src/Psl/Internal/suppress.php
```diff
@@ -0,0 +1,25 @@
...
+use function error_reporting;
+
+/**
+ * @psalm-template T
+ *
+ * @psalm-param (callable(): T) $fun
+ *
+ * @psalm-return T
+ */
+function suppress(callable $fun)
```
veewee, line 18: thanks for the warnings PHP ! :p

src/Psl/IO/Reader.php
```diff
@@ -0,0 +1,142 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\IO;
+
+use Psl;
+use Psl\Str;
+
+final class Reader implements ReadHandle
```
veewee, line 11: Are you also planning to add a writer?

src/Psl/IO/Internal/open.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\IO\Internal;
+
+use Psl;
+use Psl\Internal;
+
+function open(string $uri, string $mode): ResourceHandle
```
veewee, line 15: Since this one is internal, will we (besides stdin, stderr, stdout) also provide shortcut functions for any other wrapper out there and what with custom wrappers?

https://www.php.net/manual/en/wrappers.php

src/Psl/IO/CloseSeekReadWriteHandle.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\IO;
+
+interface CloseSeekReadWriteHandle extends CloseReadWriteHandle, CloseSeekReadHandle, CloseSeekWriteHandle, SeekReadWriteHandle
```
veewee: Can you tell me what extra value these combined interfaces provide?
Is it something you will actually typehint on. I'dd assume if I used this in the code, I'dd typehint for `ResourceHandle` most of the time?

veewee (commented):
Looks promising. The interface system for stream handlers are far more superiour than PSR-7 streams! At the moment, I am not sure about all those combined interfaces, but they don't hurt anyone either ...

src/Psl/Internal/suppress.php
```diff
@@ -0,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Internal;
```
veewee: Any reason why you want to internalize this one?
It looks like a function I could use in some of the packages I write as well.

Together with a shortcut function for:

```php
$result = some_php_function_that_could_return_false();
if ($result === false) {
    $err = error_get_last();

    Psl\invariant_violation('error(%d): %s.', $err['type'], $err['message']);
}
```

src/Psl/IO/CloseSeekReadWriteHandle.php
azjezz: most of the time you want to type hint for the feature you want to be enabled in the handle, STDOUT and others cannot be closed, if you type hint against `ResourceHandle`, then you will have access to ->close(), which doesn't actually work in that case.

src/Psl/IO/Reader.php
azjezz: I think not, if we would have a writer, it would probably only contain `writeLine` beside what already exists in `WriteHandle`, and that is pretty simple to do yourself ( write($text."\n"); )
unless you have other things in mind that we can add in a writer?

src/Psl/IO/Internal/open.php
azjezz: honestly, i don't think we should support wrappers.

instead of creating wrappers, people should implement custom handles.

src/Psl/IO/CloseSeekReadWriteHandle.php
veewee: That is true, but the internal `open()` function always returns a `ResourceHandle`. So even if you typehint it as a `WriteHandle`, you still have access to the `close()` function.

src/Psl/IO/Reader.php
veewee: Some other possible additions to a writer that could make it all easier to use:

- `writeLine`
- `copy(SeekableReadHandle)` as-in `stream_copy_to_stream()`.
- `writeALot(iterable)`
- ???

For Stdout you could add some terminal specific things, but that should most likely go in it's own specific writer.

We can always add this stuff if we need it somewhere!

src/Psl/IO/CloseSeekReadWriteHandle.php
azjezz: `ResourceHandle` is internal, and you shouldn't type hint against it, so is `open`.

e.g:

```
my_function(ReadHandle $handle): void {
  // do something with it
}

$stdout = IO\output_handle();
my_function($stdout); // error: expected `ReadHandle` got `WriteHandle`

$stdin = IO\input_handle();
my_function($stdin); // ok
``` 

i know at runtime this will not trigger an error ( at least for the type ), but i'm not sure how to handle that.

IMO, we can handle a decorator for each interface, where the decorator takes `ResourceHandle` and only implement functions for said interface, so that at runtime, you will also get a type error.

e.g:

```
/**
 * @internal
 */
final class WriteHandleDecorator implements WriteHandle {
  public function __construct(private ResourceHandle $handle) {}

  // implement write() here
}
```

so we can have `IO\input_handle` and friends return decorator instances that will ensure no misuse happens.

WDYT?

src/Psl/IO/Reader.php
azjezz: > copy(SeekableReadHandle)

i think that can be a function in its own `IO\copy(ReadHandle $source, WriteHandle $destination, ?int $chunk_size = null, ?float $timeout_seconds = null)`

> writeALot(iterable)

this is a one liner :) 
`Iter\apply($iterable, fn(string $content) => $handle->write($content));`

> terminal specific things

we don't really want that, we want `IO` to contain general purpose IO things, as the same handles will be used for files and network sockets.

src/Psl/IO/CloseSeekReadWriteHandle.php
veewee: ow, missed the fact that you marked `ResourceHandle` as internal. However that doesn't really matter, since it all is publicly accessible.

The big question is : do you want to put in the extra effort for runtime safety or do you trust the users to have static analyzers in place that will tell them they are doing it wrong. 

For me personally, it doesnt matter since psalm will tell me I'm wrong. But I don't the majority of projects use that yet.

I do like the decorators approach: that would also give an actual use-case for the interfaces you added.

azjezz: i think we should put the effort, since even psalm will agree with this:

```
function f(ReadHandle $handle): void {
  if ($handle instanceof SeekHandle) {
    $handle->seek(0);
  }

  $data = $handle->read();
}
```

if you pass `IO\input_handle` to `f`, it shouldn't be able to seek through it, but currently, since the return value is `ResourceHandle`, it can, and both runtime and psalm will agree with it ( even tho that seek will through an exception ).

src/Psl/IO/error_handle.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Return the output handle for the current request.
+ *
+ * This should generally be used for sending data to clients. In CLI mode, this
+ * is usually the process STDOUT.
+ */
+function error_handle(): ?WriteHandle
+{
+    if (php_sapi_name() === "cli") {
```
BackEndTea: Wouldn't `PHP_SAPI === 'cli'` be better here? (Same for other `php_sapi_name()` calls)

src/Psl/Internal/suppress.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * @psalm-param (callable(): T) $fun
+ *
+ * @psalm-return T
+ */
+function suppress(callable $fun)
+{
+    $level = error_reporting(0);
+
+    $result = $fun();
```
BackEndTea: This should probably be wrapped in a `try` `finally`. Otherwise we may end up having `error_reporting` on 0 for the rest of the process.

e.g.

```php
$level = error_reporting(0);

try {
    $result = $fun();
} finally {
    error_reporting($level);
}

return $result;
```

src/Psl/IO/Reader.php
```diff
@@ -0,0 +1,142 @@
...
+                return null;
+            }
+            $buf .= $chunk;
+        } while (!Str\Byte\contains($chunk, $suffix));
+
+        $idx = Str\Byte\search($buf, $suffix);
+        Psl\invariant($idx !== null, 'Should not have exited loop without suffix');
+        $this->buffer = Str\Byte\slice($buf, $idx + $suffix_len);
+        return Str\Byte\slice($buf, 0, $idx);
+    }
```
azjezz, line 107: i just spent 4 hours thinking this method was doing something wrong, turned out i was wrong.

*but* made it 17 lines long instead of 21 in the process, so i guess that's a plus.

src/Psl/IO/CloseSeekReadWriteHandle.php
azjezz: hm, figured i can do:

```php
function input_handle(): ReadHandle
{
    if (PHP_SAPI === "cli") {
        $handle = Internal\open('php://stdin', 'rb');
    } else {
        $handle = Internal\open('php://input', 'rb');
    }

    return new class($handle) implements ReadHandle {
        private Internal\ResourceHandle $handle;

        public function __construct(
            Internal\ResourceHandle $handle
        ) {
            $this->handle = $handle;
        }

        /**
         * An immediate, unordered read.
         *
         * @param int|null $max_bytes the maximum number of bytes to read.
         *
         * up to `$max_bytes` may be allocated in a buffer; large values may lead
         * to unnecessarily hitting the request memory limit.
         *
         * @returns string the read data on success, or an empty string if the end of file is reached.
         *
         * @throws Exception\BlockingException if there is no more data available to read.
         * @throws InvariantViolationException If $max_bytes is 0.
         */
        public function read(?int $max_bytes = null): string
        {
            return $this->handle->read($max_bytes);
        }
    };
}
```

is there any downside to this that didn't cross my mind?

veewee: The biggest issue imo is reusability: you now always have to repeat that method and it's docblock if you want to create a read-only handle.

src/Psl/Internal/suppress.php
azjezz: because this is a "hack" around a behavior that is only caused by PHP builtin functions, PSL functions don't trigger any errors, only exceptions, if you are doing this as well, you are calling php builtin functions, which PSL is trying to replace, so instead of providing a way for people to work around PHP builtin functions, we should provide then with a complete *sane* replacement, that's why i'm against exposing things like this in the public API;

veewee: Agree with that vision.
Nothing is really internal though and it could be nice to not have to copy this kind of stuff in my own work.

However : If I could use psl for everything, I might not need to use it internally anymore.

src/Psl/IO/Reader.php
veewee: Looks good. Do you really need that 'should not happen' check at the end or is it to satisfy psalm's `null|string`?

src/Psl/IO/Reader.php
azjezz: added writer :)

src/Psl/IO/Internal/ResourceHandle.php
```diff
@@ -0,0 +1,189 @@
...
+     * @throws Psl\IO\Exception\BlockingException       If the handle is a socket or similar, and the read would block.
+     * @throws Psl\IO\Exception\RuntimeException        If an error occurred during the operation.
+     * @throws InvariantViolationException              If $max_bytes is 0.
+     */
+    public function read(?int $max_bytes = null): string
+    {
+        return $this->box(function () use ($max_bytes) {
+            Psl\invariant($max_bytes === null || $max_bytes > 0, '$max_bytes must be null, or > 0');
+            $metadata = stream_get_meta_data($this->resource);
+            if ($metadata['blocking']) {
```
veewee: How does non-blocking functionality behaves exactly? 
In non-blocking frameworks like amphp, you have a promise or coroutine  for those kind of actions.
Now you return similar results as the regular f* resource functions.

Does the non-blocking part matters much here?
Or is it just writing to the resource whilst running the next lines of code?

azjezz: blocking mode ensures that the stream returns right away when calling `fread` for example, instead of waiting for data to be available ( e.g: socket ). ( see https://www.php.net/manual/en/function.stream-set-blocking.php )

veewee: Alrighty, makes sense then. 
The documentation on php.net is quite confusing though :)

src/Psl/IO/Reader.php
azjezz: I think it's better to keep this, `Reader` is buffered, this is useful if you want to access the un-buffered handle directly.

src/Psl/IO/CloseSeekReadWriteHandle.php
azjezz: fixed this, added 2 internal decorators.

PR #133, opened 2021-02-17
Point phpunit to local xsd

BackEndTea:
This will make sure the xsd is never out of date with the current phpunit version. It also gives auto completion/validation in phpstorm right away, without having to fetch it.


azjezz (approved):

azjezz:
Thank you @BackEndTea!

BackEndTea:
Thank you @azjezz !

PR #135, opened 2021-02-19
[Regex] Introduce Regex component

### LINKED ISSUE #29: Regex API

azjezz:
closes #29 

This PR doesn't provide a complete replacement for `preg_*` functions, as i don't think we can really type everything correctly.

therefor, here we only add functions that have a known return type, such as `replace`, `split`, and `matches`.

---

changelog:

Features:

- added `Psl\Regex\matches` function
- added `Psl\Regex\replace` function
- added `Psl\Regex\replace_every` function ( for consistency with `Psl\Str\replace_every`, `Psl\Str\Byte\replace_every`, .. etc )
- added `Psl\Regex\split` function

### DIFF
--- a/src/Psl/Regex/matches.php
+++ b/src/Psl/Regex/matches.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Regex;
+
+use function preg_match;
+
+/**
+ * Determine if $subject matches the given $pattern.
+ *
+ * @param non-empty-string  $pattern    The pattern to match against.
+ *
+ * @throws Exception\InvalidPatternException If $pattern is invalid.
+ * @throws Exception\RuntimeException        If an internal error accord.
+ *
+ * @psalm-pure
+ */
+function matches(string $subject, string $pattern, int $offset = 0): bool
+{
+    $_ = [];
+    $result = Internal\call_preg(
+        'preg_match',
+        static fn() => preg_match($pattern, $subject, $_, 0, $offset),
+    );
+
+    return $result === 1;
+}
--- a/src/Psl/Regex/replace.php
+++ b/src/Psl/Regex/replace.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Regex;
+
+use Psl\Type;
+
+use function preg_replace;
+
+/**
+ * Returns the '$haystack' string with all occurrences of `$pattern` replaced by
+ * `$replacement`.
+ *
+ * @param non-empty-string  $pattern    The pattern to search for.
+ *
+ * @throws Exception\InvalidPatternException If $pattern is invalid.
+ * @throws Exception\RuntimeException In case of an unexpected error.
+ *
+ * @psalm-pure
+ */
+function replace(string $haystack, string $pattern, string $replacement): string
+{
+    $result = Internal\call_preg(
+        'preg_replace',
+        static fn() => preg_replace($pattern, $replacement, $haystack),
+    );
+
+    // @codeCoverageIgnoreStart
+    try {
+        /**
+         * @psalm-suppress ImpureFunctionCall - see #130
+         * @psalm-suppress ImpureMethodCall -see #130
+         */
+        return Type\string()->assert($result);
+    } catch (Type\Exception\AssertException $e) {
+        throw new Exception\RuntimeException('Unexpected error', 0, $e);
+    }
+    // @codeCoverageIgnoreEnd
+}
--- a/src/Psl/Result/Failure.php
+++ b/src/Psl/Result/Failure.php
@@ -18,6 +18,8 @@ final class Failure implements ResultInterface
 {
     /**
      * @psalm-var Te
+     *
+     * @readonly
      */
     private Exception $exception;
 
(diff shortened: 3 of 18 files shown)


src/Psl/Regex/matches.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Determine if $subject matches the given $pattern.
+ *
+ * @param non-empty-string  $pattern    The pattern to match against.
+ *
+ * @throws Exception\InvalidPatternException If $pattern is invalid.
+ *
+ * @psalm-pure
+ */
+function matches(string $subject, string $pattern, int $offset = 0): bool
```
BackEndTea, line 19: Maybe we could introduce a separate function to get the matches from a pattern?

src/Psl/Regex/replace.php
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @psalm-pure
+ */
+function replace(string $haystack, string $pattern, string $replacement): string
+{
+    $result = Internal\suppress(
+        static fn() => preg_replace($pattern, $replacement, $haystack)
+    );
+
+    if ($result->isFailed()) {
```
veewee: The code looks good, but just wondering:

WDYT about using the power of the result classes ?
It would safe you some psalm suppresses and might make the flow more visual. 
Now it's a result, then an if with error handling and finally a coercion within a try.

IMO, this code makes the flow a bit more understandable.

```php
function replace(string $haystack, string $pattern, string $replacement): string
{
    return Internal\suppress(
        static fn() => preg_replace($pattern, $replacement, $haystack)
    )->then(
        static fn ($result): string => Type\string()->assert($result),
        static function (Exception $previous) {
            throw new Exception\InvalidPatternException(Str\replace(
                $previous->getMessage(),
                'preg_replace(): ',
                ''
            ), 0, $previous);
        } 
    )->proceed(
        Fun\identity(),
        static function (Exception $error) {
            if ($error instanceof  Type\Exception\AssertException) {
                throw new Exception\RuntimeException('Unexpected error', 0, $e);
            }
            
            throw $error;
        }
    );
}

````

File: src/Psl/Regex/matches.php (hunk 1 above)
Comment (veewee) [reply src/Psl/Regex/matches.php:19]: Indeed, most of the time you want to know the matching patterns as well.

We could use 2 functions like:

- `matches(string $subject, string $pattern, int $offset=0) : bool`
- `matching(string $subject, string $pattern, TypeInterface $shape, int $offset=0) : dict`

That way, the shape validates the result and makes sure it is type-safe?
If the pattern does not match, you'll get an exception that it cannot collect matching items from a non matching regex.

File: src/Psl/Result/Failure.php [hunk 3]
```diff
@@ -18,6 +18,8 @@ final class Failure implements ResultInterface
 {
     /**
      * @psalm-var Te
+     *
+     * @readonly
```
veewee, line 22: Is this possible?
I thought it was not, since the exception is mutable?

Cool :)

src/Psl/Regex/replace.php
azjezz: will not save us from suppression ( `then` and `proceed` are not pure ) :/

src/Psl/Regex/matches.php
azjezz: planning on it, just not the goal of this PR, not sure if i would use `TypeInterface` to force the type, or find another solution, but for now, this PR adds only basic things

src/Psl/Regex/replace.php
azjezz: simplified the code here a bit.

tests/Psl/Regex/ReplaceEveryTest.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * @dataProvider provideData
+     */
+    public function testReplaceEvery(string $expected, string $subject, array $replacements): void
+    {
+        static::assertSame($expected, Regex\replace_every($subject, $replacements));
+    }
+
+    public function provideData(): iterable
```
BackEndTea, line 20: This misses a test case where nothing can be replaced

tests/Psl/Regex/ReplaceTest.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * @dataProvider provideData
+     */
+    public function testReplace(string $expected, string $subject, string $pattern, string $replacement): void
+    {
+        static::assertSame($expected, Regex\replace($subject, $pattern, $replacement));
+    }
+
+    public function provideData(): iterable
```
BackEndTea, line 20: This also misses a test where nothing can be replaced

tests/Psl/Regex/SplitTest.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @dataProvider provideData
+     */
+    public function testSplit(array $expected, string $subject, string $pattern, ?int $limit = null): void
+    {
+        static::assertSame($expected, Regex\split($subject, $pattern, $limit));
+    }
+
+    public function provideData(): iterable
```
BackEndTea, line 20: Misses a test where there is nothing to split

BackEndTea (dismissed):
Few minor details, besides that looks good to me

PR #138, opened 2021-02-21
Introduce Class, Trait, and Interface components

### LINKED ISSUE #59: Objects API
The `Obj` component should contain functions that are related to objects, or OOP in general.

examples ( not sure about the naming yet ):

- `Psl\Obj\contains_method($object, $method): bool`
- `Psl\Obj\contains_property($object, $property): bool`
- `Psl\Obj\class_exists($classname): bool` ( always triggers the autoloader )
- `pure Psl\Obj\class_included($classname): bool` ( never triggers the autoloader )
...

azjezz:
closes #59 

changelog:

- dropped support for PHP 7.4
- Introduced `Class` component
- Introduced `Interface` component
- Introduced `Trait` component

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -1,6 +1,6 @@
 name: "coding standards"
 
-on: 
+on:
   pull_request: ~
   push: ~
 
@@ -15,7 +15,7 @@ jobs:
       - name: "installing PHP"
         uses: "shivammathur/setup-php@v2"
         with:
-          php-version: "7.4"
+          php-version: "8.0"
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
           extensions: bcmath, mbstring, intl, sodium, json
--- a/.github/workflows/documentation-check.yml
+++ b/.github/workflows/documentation-check.yml
@@ -1,6 +1,6 @@
 name: "documentation check"
 
-on: 
+on:
   pull_request: ~
   push: ~
 
@@ -15,7 +15,7 @@ jobs:
       - name: "installing PHP"
         uses: "shivammathur/setup-php@v2"
         with:
-          php-version: "7.4"
+          php-version: "8.0"
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
           extensions: bcmath, mbstring, intl, sodium, json
--- a/.github/workflows/security-analysis.yml
+++ b/.github/workflows/security-analysis.yml
@@ -15,7 +15,7 @@ jobs:
       - name: "installing PHP"
         uses: "shivammathur/setup-php@v2"
         with:
-          php-version: "7.4"
+          php-version: "8.0"
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
           extensions: bcmath, mbstring, intl, sodium, json
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -17,7 +17,7 @@ jobs:
       - name: "installing PHP"
         uses: "shivammathur/setup-php@v2"
         with:
-          php-version: "7.4"
+          php-version: "8.0"
           ini-values: memory_limit=-1
           tools: composer:v2, cs2pr
           extensions: bcmath, mbstring, intl, sodium, json
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -1,6 +1,6 @@
 name: "unit tests"
 
-on: 
+on:
   pull_request: ~
   push: ~
 
@@ -13,7 +13,6 @@ jobs:
     strategy:
       matrix:
         php-version:
-          - "7.4"
           - "8.0"
         operating-system:
           - "macos-latest"
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 
 # php-cs-fixer cache
 /.php_cs.cache
+/.php-cs-fixer.cache
 
 # phpunit cache
 /tools/phpunit/.phpunit.result.cache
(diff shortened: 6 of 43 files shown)


src/Psl/ObjectOriented/class_alias.php
```diff
@@ -0,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\ObjectOriented;
```
azjezz: `Object` and `Class` are reserved, `Obj` and `Cls` are "weird", and `OO` is weirder.

i think `ObjectOriented` makes the most sense.

BackEndTea: Projects Like PhpParser solve this by adding a `_`, e.g. `Object_`. But i think this works fine too

src/Psl/ObjectOriented/class_alias.php
```diff
@@ -0,0 +1,43 @@
...
+ * @psalm-assert-if-true class-string<T> $alias
+ *
+ * @return bool Returns false if unable to create an alias.
+ */
+function class_alias(string $class_name, string $alias): bool
+{
+    $previous = error_reporting(0);
+
+    try {
+        return php_class_alias($class_name, $alias);
```
BackEndTea: Why not use the supress function here?

src/Psl/ObjectOriented/class_exists.php
```diff
@@ -0,0 +1,25 @@
...
+ *
+ *      ObjectOriented\class_exists('Psl\Collection\Vector');
+ *      => Bool(true)
+ *
+ *      ObjectOriented\class_exists('Psl\Collection\Foo');
+ *      => Bool(false)
+ *
+ * @psalm-assert-if-true class-string $class_name
+ */
+function class_exists(string $class_name): bool
```
BackEndTea: How about adding a `class_is_loaded()` function, which does a `class_exists`, but with the 2nd param being false?

azjezz: yea, probably a good idea, not sure about the name `class_included`? `class_loaded`?

src/Psl/ObjectOriented/class_alias.php
azjezz: because i forgot that i added it 🤦‍♂️

src/Psl/ObjectOriented/class_exists.php
veewee: I'dd prefer `class_loaded`, since included is a bit strange in php context.

src/Psl/ObjectOriented/get_class.php
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @param T $object
+ *
+ * @return class-string<T>
+ *
+ * @psalm-pure
+ */
+function get_class(object $object): string
+{
+    return php_get_class($object);
```
veewee: What is the end-user's benifit of using these kind of functions instead of the internal php function?
Is it just for the sake of providing everything?

I've seen these functions in other places in PSL as well, but in my code I avoided these and used the php internal once instead.

src/Psl/ObjectOriented/class_alias.php
veewee: Couldn't come up with a better name either...

src/Psl/ObjectOriented/get_class.php
azjezz: better typed. guarantee to have the same behavior as other user land code, and no type juggling.

In this case specifically, no benefits, but to be consistent and provide everything.

src/Psl/ObjectOriented/class_alias.php
azjezz: `Class` could be used as a namespace in PHP 8.0, i think i'm pulling a @Ocramius card here and requiring PHP ^8.0 in PSL 1.7 🤔

veewee: I can remember the pain of having classes named after a type whilst upgrading from php 5 to 7 and would not recommend that. Even though you can in php 8.

My IDE will autocomplete the namespace, so I don't mind that it is named object-oriented. That much.

But it's up to you :)

azjezz: `Class` is not a type + `Class` is going to be used for a namespace here so your IDEA will never import it if you are typing an argument type or a return type

azjezz: > My IDE will autocomplete the namespace, so I don't mind that it is named object-oriented. That much.

personally i don't import symbols, but rather namespaces most of the time.

so it's:

```php
use Psl\Class;

if (Class\exists($foo)) {
 ...
}
```

rather than:

```php
use function Psl\Class\exists;

if (exists($foo)) {
 ...
}
```

and here `Class\exists($foo)` reads better than `ObjectOriented\class_exists($foo)` ( and definitly better than `exists($foo)` ).

Ocramius:
Instead of introducing a breakage just for being able to use `Class\`, may
I suggest `ClassName\`? 😛

azjezz:
ugh, now i'm not sure if either `Class`, `Object`, or `ClassName` make sense since the namespace also contains functions that deal with traits and interfaces ( interface_exists, trait_exists .. etc )

Ocramius:
`exists()` could be the function name 😛

azjezz:
`Psl\Class\exists`, `Psl\Interface\exists`, `Psl\Trait\exits`? seems like an overkill to have a separate component for just one function.

maybe `Psl\Runtime\class_exists`? ( as in "check if the class exists at *runtime* )? ( the component would definitely contain other functions related to PHP runtime )

Ocramius:
What's "runtime"?

azjezz:
the time when things run :p

Ocramius:
Oh, that's how a German would read it, not an English speaker :P 

Anyway, I would not call it `Runtime`, as `Runtime` would be more about things like `env()`, running processes, etc. Don't see a big problem in having isolated namespaces. Don't have any better suggestions, and `ObjectOriented`, as wonky as it sounds, is still better than `Runtime` at this point.

azjezz:
update: PHP 7.4 support is dropped, now split into 3 components: Class, Trait, and Interface

PR #143, opened 2021-02-22
[Shell] Introduce shell component

### LINKED ISSUE #142: Shell component
Introduce a shell component containing a replacement for the following PHPs builtin shell functions:

- escapeshellarg ( `Psl\Shell\escape_argument` )
- escapeshellcmd ( `Psl\Shell\escape_command` )
- exec, shell_exec ( `Psl\Shell\execute` )

`Psl\Shell\execute` must always return a string, containing the full output, if the command returned a non `0` exist code, an exception should be thrown.

The thrown exception should contain all information that the end user might need ( e.g: `$exception->getCommand()`, `$exception->getExitCode()`, `$exception->getOutput()` .. etc )

azjezz:
closes #142 

---

changelog:

- added `Psl\Shell\execute` function
- added `Psl\Shell\escape_argument` function
- added `Psl\Shell\escape_command` function

### DIFF
--- a/src/Psl/Shell/execute.php
+++ b/src/Psl/Shell/execute.php
@@ -1,0 +1,89 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Shell;
+
+use Psl\Dict;
+use Psl\Env;
+use Psl\Str;
+use Psl\Vec;
+
+use function fclose;
+use function is_dir;
+use function is_resource;
+use function proc_close;
+use function proc_open;
+use function stream_get_contents;
+
+/**
+ * Execute an external program.
+ *
+ * @param string                $command            The command to execute.
+ * @param list<string>          $arguments          The command arguments listed as separate entries.
+ * @param string                $working_directory  The initial working directory for the command.
+ *                                                  This must be an absolute directory path, or null
+ *                                                  if you want to use the default value (the working dir of the
+ *                                                  current directory )
+ * @param array<string, string> $environment        An array with the environment variables for the command that
+ *                                                  will be run.
+ * @param bool                  $escape_arguments   If set to true ( default ), all $arguments will be escaped using
+ *                                                  `Shell\escape_argument`.
+ *
+ * @psalm-taint-sink shell $command
+ *
+ * @throws Exception\FailedExecutionException   In case the command resulted in an exit code other than 0.
+ * @throws Exception\PossibleAttackException    In case the command being run is suspicious ( e.g: contains NULL byte ).
+ * @throws Exception\RuntimeException           In case $working_directory doesn't exist, or unable to create a new
+ *                                              process.
+ */
+function execute(
+    string $command,
+    array $arguments = [],
+    ?string $working_directory = null,
+    array $environment = [],
+    bool $escape_arguments = true
+): string {
+    if ($escape_arguments) {
+        $arguments = Vec\map($arguments, static fn(string $argument): string => escape_argument($argument));
+    }
+
+    $commandline = Str\join([$command, ...$arguments], ' ');
+
+    /** @psalm-suppress MissingThrowsDocblock - safe ( $offset is within-of-bounds ) */
+    if (Str\contains($commandline, "\0")) {
+        throw new Exception\PossibleAttackException('NULL byte detected.');
+    }
+
+    $descriptor = [
+        1 => ['pipe', 'w'],
+        2 => ['pipe', 'w'],
+    ];
+
+    $environment = Dict\merge(Env\get_vars(), $environment);
+    $working_directory = $working_directory ?? Env\current_dir();
+    if (!is_dir($working_directory)) {
+        throw new Exception\RuntimeException('$working_directory does not exist.');
+    }
+
+    $process = proc_open($commandline, $descriptor, $pipes, $working_directory, $environment);
+    // @codeCoverageIgnoreStart
+    // not sure how to replicate this, but it can happen \_o.o_/
+    if (!is_resource($process)) {
+        throw new Exception\RuntimeException('Failed to open a new process.');
+    }
+    // @codeCoverageIgnoreEnd
+
+    $stdout_content = stream_get_contents($pipes[1]);
+    $stderr_content = stream_get_contents($pipes[2]);
+
+    fclose($pipes[1]);
+    fclose($pipes[2]);
+
+    $code = proc_close($process);
+    if ($code !== 0) {
+        throw new Exception\FailedExecutionException($commandline, $stdout_content, $stderr_content, $code);
+    }
+
+    return $stdout_content;
+}
--- a/src/Psl/Internal/Loader.php
+++ b/src/Psl/Internal/Loader.php
@@ -456,6 +456,9 @@ final class Loader
         'Psl\Encoding\Base64\decode',
         'Psl\Encoding\Hex\encode',
         'Psl\Encoding\Hex\decode',
+        'Psl\Shell\escape_command',
+        'Psl\Shell\escape_argument',
+        'Psl\Shell\execute',
     ];
 
     public const INTERFACES = [
@@ -477,6 +480,7 @@ final class Loader
         'Psl\Type\TypeInterface',
         'Psl\Type\Exception\ExceptionInterface',
         'Psl\Regex\Exception\ExceptionInterface',
+        'Psl\Shell\Exception\ExceptionInterface',
     ];
 
     public const TRAITS = [];
@@ -525,6 +529,9 @@ final class Loader
         'Psl\Encoding\Exception\IncorrectPaddingException',
         'Psl\Encoding\Exception\RangeException',
         'Psl\Regex\Exception\InvalidPatternException',
+        'Psl\Shell\Exception\FailedExecutionException',
+        'Psl\Shell\Exception\RuntimeException',
+        'Psl\Shell\Exception\PossibleAttackException',
     ];
 
     private const TYPE_CONSTANTS = 1;
--- a/src/Psl/Shell/Exception/ExceptionInterface.php
+++ b/src/Psl/Shell/Exception/ExceptionInterface.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Shell\Exception;
+
+use Psl\Exception;
+
+interface ExceptionInterface extends Exception\ExceptionInterface
+{
+}
--- a/src/Psl/Shell/Exception/FailedExecutionException.php
+++ b/src/Psl/Shell/Exception/FailedExecutionException.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Shell\Exception;
+
+use Psl\Str;
+
+final class FailedExecutionException extends RuntimeException implements ExceptionInterface
+{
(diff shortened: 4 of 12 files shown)


src/Psl/Shell/execute.php
```diff
@@ -0,0 +1,89 @@
...
+ * @throws Exception\RuntimeException           In case $working_directory doesn't exist, or unable to create a new
+ *                                              process.
+ */
+function execute(
+    string $command,
+    array $arguments = [],
+    ?string $working_directory = null,
+    array $environment = [],
+    bool $escape_arguments = true
+): string {
```
azjezz, line 46: `shell_exec` and `exec` are not powerful enough, `proc_*` is a bit complex, i think this command is a nice middle ground where you can use a costume working directory, specify environment variables, and have more control ( e.g: `exec` will redirect error output of the command being ran to PHPs own process STDERR, which is not what you want most likely, but using `Shell\execute`, you get stderr output in the exception in case of an error ).

However, `Shell\execute` is still not as powerful as `proc_*` ( e.g: cannot write to process input ), so i guess if people want more control, they can either use proc* or `symfony/process` package.



WDYT @veewee?

src/Psl/Shell/execute.php
```diff
@@ -0,0 +1,89 @@
...
+    if ($escape_arguments) {
+        $arguments = Vec\map($arguments, static fn(string $argument): string => escape_argument($argument));
+    }
+
+    $commandline = Str\join([$command, ...$arguments], ' ');
+
+    /** @psalm-suppress MissingThrowsDocblock - safe ( $offset is within-of-bounds ) */
+    if (Str\contains($commandline, "\0")) {
```
azjezz, line 54: should we keep this? `exec` also results in an error in case the command contains a NULL byte.

azjezz:
so far, I'm satisfied with the API, i don't think we want to add anything more complex yet ( at least no until we have #127 ).

src/Psl/Shell/execute.php
```diff
@@ -0,0 +1,89 @@
...
+    array $arguments = [],
+    ?string $working_directory = null,
+    array $environment = [],
+    bool $escape_arguments = true
+): string {
+    if ($escape_arguments) {
+        $arguments = Vec\map($arguments, static fn(string $argument): string => escape_argument($argument));
+    }
+
+    $commandline = Str\join([$command, ...$arguments], ' ');
```
veewee, line 51: What about escaping the command itself?

src/Psl/Shell/execute.php
veewee: For sake of security, it seems ok to keep this logic.
The escape_arguments version replaces it with '?'. Which would return an exception on the cli most likely?

src/Psl/Shell/execute.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /** @psalm-suppress MissingThrowsDocblock - safe ( $offset is within-of-bounds ) */
+    if (Str\contains($commandline, "\0")) {
+        throw new Exception\PossibleAttackException('NULL byte detected.');
+    }
+
+    $descriptor = [
+        1 => ['pipe', 'w'],
```
veewee, line 59: Is it a good idea to support an stdin resource stream from the execute() api?
Currently you only pipe the stdout and stderr.

src/Psl/Shell/execute.php
```diff
@@ -0,0 +1,89 @@
...
+
+    $process = proc_open($commandline, $descriptor, $pipes, $working_directory, $environment);
+    // @codeCoverageIgnoreStart
+    // not sure how to replicate this, but it can happen \_o.o_/
+    if (!is_resource($process)) {
+        throw new Exception\RuntimeException('Failed to open a new process.');
+    }
+    // @codeCoverageIgnoreEnd
+
+    $stdout_content = stream_get_contents($pipes[1]);
```
veewee, line 77: Here you are fetching stdout and stderr. However, there are tools that report to both stdout and stderr.
For example take "composer run".
It reports the command it is executing to stdout, but the results to stdin.
If the run fails, it reports back to stdout.

By using `stream_get_contents`, you are blocking process execution untill the stream is closed (or eof).
This means that there is no way to combine stdout and stderr back to the actual output you see when you run the command on the screen.

So what would be better? Maybe wrap the output in a result and keep the target output linked to the actually outputted content? Something like:

```php
$output = [];
$output[] = ["STDERR", 'Starting command composer run mytask'];
$output[] = ["STDOUT", 'the actual output of the command'];
$output[] = ["STDERR", 'Execution task mytask failed : some context'];
```


You can then wait for the process exection like this:

```php
do {
    $status = proc_get_status($process);
    // Check if content is added to stdout / stderr and link it to the custom output class.
} while ($status && $status['running']);
```

(In that case, the proc_close exit code is not trustworthy, you'll need to fetch the exit code like this)

```php
do {}  while($status && $status['running']);
$exitCode = $status['exitcode'] ?? -1;
proc_close($process);
```


That would give you a chronological order of messages, filterable on what stream it was added.

WDYT?

src/Psl/Shell/execute.php
```diff
@@ -0,0 +1,89 @@
...
+
+    fclose($pipes[1]);
+    fclose($pipes[2]);
+
+    $code = proc_close($process);
+    if ($code !== 0) {
+        throw new Exception\FailedExecutionException($commandline, $stdout_content, $stderr_content, $code);
+    }
+
+    return $stdout_content;
```
veewee, line 88: stdout content could also be usefull.

src/Psl/Shell/execute.php
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @psalm-taint-sink shell $command
+ *
+ * @throws Exception\FailedExecutionException   In case the command resulted in an exit code other than 0.
+ * @throws Exception\PossibleAttackException    In case the command being run is suspicious ( e.g: contains NULL byte ).
+ * @throws Exception\RuntimeException           In case $working_directory doesn't exist, or unable to create a new
+ *                                              process.
+ */
+function execute(
+    string $command,
```
veewee, line 41: The command is actually also a cli argument.
In PHP, they always keep this separated. But what is the purpose of keeping those 2 things separated?
They are converted to a string command anyways.

src/Psl/Shell/execute.php
veewee: Nice work! For most of the cases, this is sufficient. `proc_` surely is the way to go!
Maybe we can also provide a stdin ?ReadHandle param for writing to the stdin at some point that streams to `$pipes[0]`.

Can't wait to drop most of my `symfony/process` dependencies!

src/Psl/Shell/execute.php
azjezz: because in most cases, you already know the command beforehand, but likely not the arguments:

```php
$tmp = Env\temp_dir() . '/bar';

Shell\execute('foo', ['-f', $temp]);

return File\read($temp);
```

This also allows us to escape the arguments individually.

veewee: You already have that choice though: You can either use the flag `escape_arguments` or you don't.
Even though you know the command in advance, you can still just pass it as a first argument of the array.
I'm not too sure that having the main command separately really adds any benefits

src/Psl/Shell/execute.php
azjezz: It replaces NULL byte with ? only on windows tho

src/Psl/Shell/execute.php
azjezz: as i mentioned below, we don't want anything more complex until #127 is merged.

src/Psl/Shell/execute.php
azjezz: you mean stderr? 

yes, but not in most case, `Shell\execute` is a replacement for `exec` and `shell_exec`, not `proc_*` ( that will be replaced probably by `Psl\Process` once we have #127 ), and these functions don't return the content in stderr, instead it's redirected to PHPs own stderr, but that's something i don't think we should do.

src/Psl/Shell/execute.php
azjezz: assuming we merge `$command` with `$arguments`, `$command` will also be escaped as an argument, but if you don't want that, you will have to set `$escape_arguments` to `false`, and escape arguments yourself before calling `Shell\execute` in order to leave the command as is.

src/Psl/Shell/execute.php
azjezz: that's why i said this is not a replacement for `proc_*`, it is more of a replacement to `exec` and `shell_exec` but it allows you to change environment variables, working_directory, escape things for you by default .. etc.

we will have something in the future that can totally replace `proc_*`, but it's not the goal of `Shell\execute`

src/Psl/Shell/execute.php
veewee: Yes stderr indeed.

What addition would the Psl process component have on top of Shell\Execute then? Why not go for 1 way of interacting with processes? 

If you only return stdout, you loose stderr if the command finished correctly.
So in case of the `composer run` example, you miss information.

Maybe return a chronological stderr + stdout string as a result instead?

src/Psl/Shell/execute.php
veewee: In which case wouldn't you want the command itself to be escaped?
Are you ever providing a command escaped directly as input?

Most of the times, you just pass the commandname. Like E.g: run  `php` as command, then it wraps it with quotes like this `"php"`. Which is basically what symfony/process does as well.

src/Psl/Shell/execute.php
azjezz: > If you only return stdout, you loose stderr if the command finished correctly.
> So in case of the composer run example, you miss information.

that's the same with `exec` and `shell_exec`, which `Shell\execute` is meant to replace.

the advantage of using `Shell\execute` instead of `exec`/`shell_exec`:

1. in case of an error, a `Shell\Exception\FailedExecutionException` is thrown instead of an error+stderr being redirected to phps stderr.
2. you can change working directory
3. you can add additional environment variables
4. it escapes arguments for you by default

if you want to have an interactive process + always get your hand on the stderr output of the underlying process, you should probably use `proc_*` or `symfony/process` directly for now, until we have IO handles, in which case we can introduce `Psl\Process` component that would be more powerful, and would be able to replace `proc_*` and `symfony/process` completely.

we want to keep this simple

src/Psl/Shell/execute.php
azjezz: in this case, the command is escaped - or known to be safe ( `@psalm-taint-sink shell $command` ), psalm taint analysis will catch it in case it's something like `Shell\execute($_POST['command'])`.

and another thing: `Shell\execute('foo');` is better/easier to type than `Shell\execute(['foo']);`

> Which is basically what symfony/process does as well.

you shouldn't really compare this to `symfony/process`, but rather to `exec`/`shell_exec`

src/Psl/Shell/execute.php
veewee: My expectations of this component might have been different then yours.
If you are sure about the API and implementation, then I don't see any reason not to merge it.

azjezz: you should wait for `Psl\Process` 😉 already on the work

PR #144, opened 2021-02-23
[Filesystem] Introduce filesystem component

### LINKED ISSUE #141: Filesystem component
Introduce a filesystem component containing a replacement for the following PHPs builtin filesystem functions:

-    basename ( `Psl\Filesystem\basename` )
-    chgrp ( `Psl\Filesystem\chgrp` )
-    chmod ( `Psl\Filesystem\chmod` )
-    chown ( `Psl\Filesystem\chown` )
-    copy ( `Psl\Filesystem\copy` )
-    dirname ( `Psl\Filesystem\dirname` )
-    file_​exists ( `Psl\Filesystem\file_exists` )
-    glob ( `Psl\Filesystem\glob` )
-    is_​dir ( `Psl\Filesystem\is_dir` )
-    is_​executable ( `Psl\Filesystem\is_executable` )
-    is_​file ( `Psl\Filesystem\is_file` )
-    is_​link ( `Psl\Filesystem\is_link` )
-    is_​readable ( `Psl\Filesystem\is_readable` )
-    is_​writable ( `Psl\Filesystem\is_writable` )
-    lchgrp ( `Psl\Filesystem\link_chgrp` )
-    lchown ( `Psl\Filesystem\link_chown` )
-    link ( `Psl\Filesystem\link` )
-    mkdir ( `Psl\Filesystem\mkdir` )
-    realpath ( `Psl\Filesystem\realpath` )
-    rename ( `Psl\Filesystem\rename` )
-    rmdir ( `Psl\Filesystem\rmdir` )
-    symlink ( `Psl\Filesystem\symlink` )
-    touch ( `Psl\Filesystem\touch` )
-    unlink ( `Psl\Filesystem\unlink` )

azjezz:


azjezz:
@veewee this on the other hand, can replace `symfony/filesystem` 😛 ~( apart from the writing to files part, that will come after we have File handles, which is the next step after #127 )~ ( added `Filesystem\write_file` / `Filesystem\read_file`, these should be a replacement for `file_put_contents`/`file_get_contents`, `fwrite`/`fread`/`fgets` .. etc functions will later be replaced with handles, which replace resources )

src/Psl/Filesystem/change_group.php
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Change the group ownership of $filename.
+ *
+ * @param bool          $recursive  Whether change the group ownership recursively or not.
+ *
+ * @throws Exception\RuntimeException   If unable to change the group ownership for $filename.
+ * @throws Psl\Exception\InvariantViolationException If $filename does not exist.
+ */
+function change_group(string $filename, int $group, bool $recursive = false): void
```
azjezz: with function names, i decided to go with something more descriptive than what php ( also c, c++, unix .. ) has, but i think having more descriptive names are better, and are more newbie friendly ( e.g: `change_permissions` vs `chmod`, `create_file` vs `touch`, `create_directory` vs `mkdir`, `delete_file` instead of `unlink`, `delete_directory` instead of `rmdir` )

src/Psl/Filesystem/copy.php
```diff
@@ -0,0 +1,67 @@
...
+use function stream_copy_to_stream;
+
+/**
+ * Change the group ownership of $filename.
+ *
+ *
+ * @throws Exception\RuntimeException                   If unable to change the group ownership for $filename.
+ * @throws Psl\Exception\InvariantViolationException    If $source does not exist or is not readable.
+ */
+function copy(string $source, string $destination, bool $overwrite = false): void
```
veewee, line 21: It might have been better if you had IO + Filehandle ready to use inside this function.

veewee (commented):
Besides the missing tests, the implementations look good to me! nice one

src/Psl/Filesystem/copy.php
azjezz: yea, this is currently just for local files, i think it's okay to keep it even if we have `IO\copy(ReadHandle $source, WriteHandle $destination): void`

src/Psl/Filesystem/delete_directory.php
```diff
@@ -0,0 +1,49 @@
...
+            new FilesystemIterator(
+                $directory,
+                FilesystemIterator::CURRENT_AS_PATHNAME | FilesystemIterator::SKIP_DOTS,
+            ),
+            Fun\when(
+                static fn(string $node) => is_directory($node),
+                static fn(string $node) => delete_directory($node, true),
+                static fn(string $node) => delete_file($node),
+            )
+        );
```
azjezz: this is the first time for me using `Fun` 😛

veewee: It's fun to use! ;)

src/Psl/Filesystem/exists.php
```diff
@@ -0,0 +1,20 @@
...
+ * @param string $filename  Path to the file.
+ *
+ * If $filename is a relative filename, it will be checked relative to
+ * the current working directory.
+ *
+ * @return bool true if $filename exists, false otherwise.
+ */
+function exists(string $filename): bool
+{
+    return is_file($filename) || is_directory($filename);
```
BackEndTea: Why not use `file_exists()` here?

azjezz:
#itWorksOnMyMachine

image

azjezz:
#itWorksOnMyMachine

image

phpunit.xml.dist
```diff
@@ -6,12 +6,26 @@
...
       <file>src/Psl/Math/constants.php</file>
+      <file>src/Psl/Filesystem/constants.php</file>
+
+      <!-- Things that are not easily tested -->
+      <file>src/Psl/Filesystem/get_group.php</file>
+      <file>src/Psl/Filesystem/change_group.php</file>
+      <file>src/Psl/Filesystem/Internal/change_group.php</file>
+      <file>src/Psl/Filesystem/get_owner.php</file>
+      <file>src/Psl/Filesystem/change_owner.php</file>
+      <file>src/Psl/Filesystem/Internal/change_owner.php</file>
```
azjezz, line 28: no way for us to really test these.

veewee: Other packages do this by not using fully-qualified function calls.
This way you can overwrite the internal php functions in the namespace of the function and assert based on the attributes passed to that function.

A bit dirty and it doesn't  really test the result, but it gives you more than not testing...

veewee (commented):
> #itWorksOnMyMachine
> 
> image

Tested it as well on my mac and works. Sometimes github actions filesystem does some strange stuff. Got similar issues in other packages.

src/Psl/Filesystem/write_file.php
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * Write $content to $file.
+ *
+ * @param bool $append If true, and $file exists, append $content to $file instead of overwriting it.
+ *
+ * @throws Psl\Exception\InvariantViolationException If the file specified by
+ *  $file does not exist, or is not writeable.
+ * @throws Exception\RuntimeException In case of an error.
+ */
+function write_file(string $file, string $content, bool $append = false): void
```
veewee: Don't like the `$append` flag.
Can we split the function into:

- `write_file($file, $content)`
- `append_file($file, $content)`

src/Psl/Filesystem/Internal/change_group.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * @param iterable<string>  $files
+ * @param string|int        $group
+ *
+ * @throws Exception\RuntimeException   If unable to change the ownership for
+ *                                      the given file.
+ *
+ * @internal
+ */
+function change_group(iterable $files, $group, bool $recursive = false): void
```
veewee: What about splitting into:

- change_group
- change_group_recursive

src/Psl/Filesystem/Internal/change_owner.php
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * @param iterable<string>  $files
+ * @param string|int        $user
+ *
+ * @throws Exception\RuntimeException   If unable to change the ownership for
+ *                                      the given file.
+ *
+ * @internal
+ */
+function change_owner(iterable $files, $user, bool $recursive = false): void
```
veewee: - change_owner()
- change_owner_recursive()

src/Psl/Filesystem/Internal/change_permissions.php
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * @param iterable<string>  $files
+ *
+ * @throws Exception\RuntimeException   If unable to change the ownership for
+ *                                      the given file.
+ *
+ * @internal
+ */
+function change_permissions(iterable $files, int $permission, bool $recursive = false): void
```
veewee: change_permissions
change_permissions_recursive

src/Psl/Filesystem/change_group.php
veewee: change_group
change_group_recursive

src/Psl/Filesystem/Internal/change_permissions.php
veewee: Same for all other bool functions BTW :D

phpunit.xml.dist
azjezz: i would rather not test it, otherwise we are doing a hack just to make it seem like we are covering it, but we are not actually.

src/Psl/Filesystem/write_file.php
azjezz: i agree on this, will move this internally, and have both `write_file` and `append_file` call it with the flag instead.

src/Psl/Filesystem/Internal/change_group.php
azjezz: I'm having my doubts here, i think a better option would be to get rid of the recursive flag.

i had so many problems with `change_permissions` recursively, that i think it's better not to allow it.

e.g:

assuming we have directory `foo`, which contains `bar.txt`, and `baz.txt`.

if `foo` is readable, and we want to make it not readable recursively, we have to first read the directory, change the permissions of the files, then the change the permissions of `foo` itself.

however, if `foo` is not readable, and we want to make it readable, we have to do the complete opposite, i.e: change permissions of `foo` to readable, then all of it's content recursively.

another issue is that we can't change anything recursively if `foo` is not readable, and the permission we want to set also doesn't allow reading.

veewee: Wow .. I am not sure it's worth the hassel. The amount of times I ever used `change_groups` in PHP is around 0.

PR #147, opened 2021-03-02
remove composer autoloading

azjezz:
solving https://twitter.com/azjezz/status/1365076594721751042

since `Psl\` namespace is not autoloaded anymore ( everything is loaded via `Psl\Internal\Loader` ), calling `class_exists('Psl\Str\join')` will trigger autoloader still, but composer will not require `src/Psl/Str/join.php` anymore.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,6 @@
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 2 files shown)


veewee:
Nice!

PR #151, opened 2021-03-05
[Regex] Introduce *_match() functions to fetch captured data groups

### LINKED ISSUE #29: Regex API

veewee:
This PR introduces a `first_match()` and `every_match()` function in which you can specify the shape of the result:

```php
/** @var array{0: string, name: string}|null $matches */
$matches = Regex\first_match(
    'Hello world',
    '/(hello) (?<name>world)/i',
    Regex\capture_groups(['name'])
);
```

```php
/** @var list<array{0: string, name: string, digit: string}>|null $matches */
$matches = Regex\every_match(
    <<<EODATA
    a: 1
    b: 2
    c: 3
    EODATA,
    '@(?P<name>\w+): (?P<digit>\d+)@i',
    Regex\capture_groups(['name', 'digit'])
);
```


src/Psl/Regex/Type/capture_groups.php
```diff
@@ -0,0 +1,32 @@
...
+
+use function Psl\Dict\from_keys;
+use function Psl\Dict\unique;
+use function Psl\Type\shape;
+use function Psl\Type\string;
+
+/**
+ * @param list<array-key> $groups
+ *
+ * @return TypeInterface<array<array-key, string>>
```
veewee: Any ideas on how to stop psalm from losing track of the groups?

Now it returns this on the example above:
```
ERROR: Trace - src/test.php:9:1 - $matches: array<array-key, string> (see https://psalm.dev/224)
```

src/Psl/Regex/matching.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @throws Exception\RuntimeException        If an internal error accord.
+ * @throws CoercionException                 If the capture groups don't match the pattern matches
+ * @throws Exception\InvalidPatternException If $pattern is invalid.
+ */
+function matching(string $subject, string $pattern, TypeInterface $capture_groups, int $offset = 0): array
+{
+    $matching = Internal\call_preg(
+        'preg_match',
+        static function () use ($subject, $pattern, $offset): array {
```
veewee: Had to put all this logic in the `call_preg` callback, otherwise it could not place the matches in `$matching`

src/Psl/Regex/Type/capture_groups.php
veewee: Nevermind, got it through a custom plugin!

```
ERROR: Trace - src/test.php:9:1 - $matches: array{0: string, 1: string, name: string} (see https://psalm.dev/224)
```

src/Psl/Regex/matching.php
```diff
@@ -0,0 +1,46 @@
...
+            $matching = [];
+            $matches = preg_match($pattern, $subject, $matching, 0, $offset);
+
+            if ($matches === 0) {
+                throw new RuntimeException(
+                    'Could not match the pattern ' . $pattern . ' to the given input: ' . $subject . '.'
+                );
+            }
```
azjezz: how about `Psl\invariant(matches($subject, $pattern), '$subject must match $pattern');` instead? or nullable return?

src/Psl/Regex/Type/capture_groups.php
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * @psalm-suppress MixedReturnTypeCoercion - Psalm uses track of the types: Trust the return type above.
```
azjezz: is this still neeeded?

src/Psl/Regex/matching.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @template T of array
+ *
+ * @param non-empty-string $pattern        The pattern to match against.
+ * @param TypeInterface<T> $capture_groups What shape does the matching items have?
+ *
+ * @return T
+ *
+ * @throws Exception\RuntimeException        If an internal error accord.
+ * @throws CoercionException                 If the capture groups don't match the pattern matches
```
azjezz: we should wrap this inside `RuntimeException`

src/Psl/Regex/matching.php
```diff
@@ -0,0 +1,46 @@
...
+                throw new RuntimeException(
+                    'Could not match the pattern ' . $pattern . ' to the given input: ' . $subject . '.'
+                );
+            }
+
+            return $matching;
+        }
+    );
+
+    return $capture_groups->coerce($matching);
```
azjezz: we should try asserting first, see `Json\typed` for example.

src/Psl/Regex/matching.php
veewee: In this case, I don't like the invariant - since it throws a too generic exception.
Maybe better to introduce a `NoMatchesFound` exception instead of `RuntimeException`?


About nullable vs Exception : how do you want it to behave?
Most of the times you are expecting stuff to be there. So in that case an exception might make sense.
But maybe other problems lean more to a nullable solution.

integration/Psalm/EventHandler/RegexCaptureGroupsFunctionReturnTypeProvider.php
```diff
@@ -0,0 +1,76 @@
...
+            $name = $type->value;
+
+            $properties[$name] = new Type\Union([new Type\Atomic\TString()]);
+        }
+
+        return new Type\Union([new Type\Atomic\TGenericObject('Psl\Type\TypeInterface', [
+            new Type\Union([
+                new Type\Atomic\TKeyedArray($properties)
+            ])
+        ])]);
```
azjezz: this results for `TypeInterface<array{foo: string}>` for `capture_groups(['foo'])`, shouldn't it be `TypeInterface<array{0: string, foo: string}>` instead?

src/Psl/Regex/Type/capture_groups.php
veewee: Yes, it keeps on complaining about this. Not sure why that is. The types on `from_keys` seem right.

src/Psl/Regex/matching.php
azjezz: I think nullable makes sense in here, regex is expensive ( CPU wise ), so having to run `preg_matches` twice ( first to check if string matches, then to retrieve the matching parts ) is not ideal.

integration/Psalm/EventHandler/RegexCaptureGroupsFunctionReturnTypeProvider.php
veewee: indeed, nice catch!

src/Psl/Regex/matching.php
veewee: This does mean that psalm makes it `null|array{...}` and you kinda always have to check for null on it.

src/Psl/Regex/Type/capture_groups.php
azjezz: /** @psalm-trace $var **/ is your friend :D

src/Psl/Regex/Type/capture_groups.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Regex\Type;
```
azjezz: let's leave it in `Regex\` namespace, no need for `Type` sub-namespace

src/Psl/Regex/Type/capture_groups.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Psl\Type\TypeInterface;
+
+use function Psl\Dict\from_keys;
+use function Psl\Dict\unique;
+use function Psl\Type\shape;
+use function Psl\Type\string;
```
azjezz:
```suggestion
use Psl\Dict;
use Psl\Type;
```

src/Psl/Regex/Type/capture_groups.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    return shape(
+        from_keys(
+            unique([0, ...$groups]),
+            /**
+             * @return TypeInterface<string>
+             */
+            static fn() => string()
+        )
+    );
```
azjezz:
```suggestion
    return Type\shape(
        Dict\from_keys(
            Dict\unique([0, ...$groups]),
            /**
             * @return TypeInterface<string>
             */
            static fn(): Type\TypeInterface => Type\string()
        )
    );
```

src/Psl/Regex/matching.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @param non-empty-string $pattern        The pattern to match against.
+ * @param TypeInterface<T> $capture_groups What shape does the matching items have?
+ *
+ * @return T|null
+ *
+ * @throws Exception\RuntimeException        If an internal error accord.
+ * @throws Exception\InvalidPatternException If $pattern is invalid.
+ */
+function matching(string $subject, string $pattern, TypeInterface $capture_groups, int $offset = 0): ?array
```
azjezz:
```suggestion
function first_match<T>(string $subject, string $pattern, TypeInterface<T> $capture_groups, int $offset = 0): T
```

Let's also add `every_match`:

```php
function every_match<T>(string $subject, string $pattern, TypeInterface<T> $capture_groups, int $offset = 0): dict<int, T>
```

see: https://3v4l.org/uNLsi

src/Psl/Regex/matching.php
azjezz: yep, but better than wasting CPU on regex.

src/Psl/Regex/Type/capture_groups.php
veewee: Just did: The data going into the shape function is correct: `array<array-key, Psl\Type\TypeInterface<string>>`
So I guess it's a fault somewhere in the existing shape psalm plugin?

Gonna log off for the weekend and try again next week

src/Psl/Regex/matching.php
veewee: That kinda conflicts with the capture_groups type though, since now it will be an array of strings().
Would it make sense for introducing 2 types for that as well?

src/Psl/Regex/matching.php
```diff
@@ -0,0 +1,52 @@
...
+
+    try {
+        return $capture_groups->assert($matching);
+    } catch (AssertException $e) {
+    }
+
+    try {
+        return $capture_groups->coerce($matching);
+    } catch (CoercionException $e) {
+        return null;
```
veewee: What do you want it to do if the coercing fails?
Throw a runtime exception, null or the matching results?
I'dd say throw

azjezz: throw runtime exception, and we set the coercion exception as the previous exception

`throw new Exception\RuntimeException('Invalid capture groups', 0, $e);`

veewee: Introduced a new `Type\cast() function to avoid some double try-catches in the code.`

src/Psl/Regex/Type/capture_groups.php
veewee: The issue is int the plugin:
https://github.com/azjezz/psl/blob/1.5.x/integration/Psalm/EventHandler/ShapeFunctionReturnTypeProvider.php#L56

It is a `TArray` instead of a `TKeyedArray`, meaning it lost the keys in the array (which makes sense!)
In other words : The plugin can't iterator over the properties, since they are not known.

I think the suppress is valid here since the `RegexCaptureGroupsFunctionReturnTypeProvider` makes the keys known again, unless you have a beter solution?

src/Psl/Regex/matching.php
veewee: Are you ok with splitting those functions as well:

- `single_capture_groups()`
- `every_capture_groups()`

It sounds a bit crappy and is harder to 'guess' from an editor.
Another option is a `every: bool` flag which I also don't particularly like.

src/Psl/Type/cast.php
```diff
@@ -0,0 +1,27 @@
...
+ */
+function cast($data, TypeInterface $type)
+{
+    try {
+        return $type->assert($data);
+    } catch (AssertException $e) {
+    }
+
+    return $type->coerce($data);
+}
```
azjezz: hm, actually, let's not add this, and use `$type->coerce` in both regex and json. all `coerce` implementations will return the same result as assert, so i don't think we really need to assert-then-coerce.

src/Psl/Regex/matching.php
azjezz: no, first_match returns `T`, every_match should return `vec<T>` ( meaning we have to remap the result of preg_match_all )

veewee:
Introduced the `every_match` function as well. The failing test is not caused by the code, but by GH actions:

```
✗ PHP Could not setup PHP 7.4
```

src/Psl/Regex/every_match.php
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @throws Exception\RuntimeException        If an internal error accord.
+ * @throws Exception\InvalidPatternException If $pattern is invalid.
+ */
+function every_match(
+    string $subject,
+    string $pattern,
+    Type\TypeInterface $capture_groups,
+    int $offset = 0
+): ?VectorInterface {
```
azjezz: I'm still not sure if this is the best way to do this ( `$capture_groups` ), the reason we have this argument is because we want to provide type safety, but is there another way we can provide type safety ( as well as runtime safety ) without having this argument?

i think we should look at other languages and how they implement this ( beside hack, because regex there is part of the type checker, and there's a special type for it ).

for example, Java has [`Matcher`](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Matcher.html), which doesn't use generics, but provides type safety.

veewee: Agree that this is a bit cumbersome.

There is no "easy way", since you'll need to parse the regex in order to detect the different capture groups in PHP.
I've seen this project: https://github.com/BackEndTea/Regexer - but we don't want to pull in dependencies and it might be a bit heavy during runtime as well.


Otherwise, You could go for a pattern builder and use that class as instead of the regex string, but that also isn't an easy solution. + you won't be able to use regular patterns anymore if you go that way. 

That java "Matcher" solution will most likely result in a custom regex implementation no?

veewee: I've thought about it again and see 2 options:

1. Either you can do it like this. The advantage is that you can specify your own shape and cast the types to e.g. integers where it makes sense. The downside is that you manually need to declare it, even though the groups could be known from the regex.
2. Move the type part completely to the psalm plugin, meaning we'll have to write a custom regex lexer that searches for capture groups and then calculates the returned type based on that information. Advantage is that type-safe group support will be inside the language and only be calculated during static analysis. The downside is that we'll need to write a custom regex parser, it only works with psalm and you cannot cast types from the function signature. An additional benefit could be that you can parse all regexes and look for regex errors from psalm for all regex functions if done right. however, thid whole feature looks like a completely decoupled regex psalm plugin to me (in which you could use the regex parser library shared above).

src/Psl/Regex/every_match.php
```diff
@@ -0,0 +1,51 @@
...
+            return $matches === 0 ? null : $matching;
+        }
+    );
+
+    if ($matching === null) {
+        return null;
+    }
+
+    try {
+        return Type\vector($capture_groups)->coerce($matching);
```
azjezz:
```suggestion
        return Type\vec($capture_groups)->coerce($matching);
```

azjezz:
I'm starting to think that we might not need to expose `capture_groups` function at all, as outside `*_match` functions, it doesn't really serve a purpose, WDYT?

e.g:

with `every_match($subject, $pattern)`, the return type should be `list<array<array-key, string>>`
with `every_match($subject, $pattern, capture_groups: [1, 'foo']);` the return type should be `list<array{0: string, 1: string, foo: string}>`
with `first_match($subject, $pattern)` the return type should be `array<array-key, string>`
with `first_match($subject, $pattern, capture_groups: [1, 'foo']);` the return type should be `array{0: string, 1: string, foo: string}`

WDYT?

src/Psl/Regex/every_match.php
azjezz: parsing regex is the last things we really want 😛 unless we can find a parser that we can declare as an optional dependency and add note about it in the documentation so people can install it to get better static analysis for regex related things.

veewee:
We could surely make it an optional parameter and fallback to a more high-level return type.

I've thought about passing in an array instead of a type as well.
It sure is possible. The downside is that you cannot insert a custom shape.
For example: you might not be interesed in key 0 and 1 and you might want to convert numbers directly to ints.
In those cases, you could pass in a specific shape instead of providing a list of all possible arguments and next a shape of what you want it to be.

To make it a bit more clear with an example:


```php
first_match(
    '/^(x)([\d]+)(z)$/', 
    'x123z',
    shape([2 => int()], allow_unknown_fields: true)
);
```

Instead of:

```php

shape([2 => int()], allow_unknown_fields: true)->coerce(
   first_match(
        '/^(x)([\d]+)(z)$/', 
        'x123z',
        capture_groups: [0, 1, 2, 3]
    )
);
```

Therefore I also added the capture_groups function : it provides a shortcut to create a sensible shape for all provided capture groups. Nothing advanced and all required fields.
I'm not against providing the capture_groups() function, since it is easier to add than creating a custom shape of all strings:

```php
shape([
    0 => string(),
    1 => string(),
    /// ....
])
```


So in short, I think my counter-proposal is to make $capture_groups of type `?TypeInterface<T>` so that it is not required for the user.
Wheter you want a shortcut function to create a capture_groups type shape or not is up to you to decide. I can always remove it of course.

azjezz:
i think we should make it `?TypeInterface<T>`, and fallback to the high-level type if not present, if present, we need to enforce that the given type is array{}/list<array{}> via the plugin.

veewee:
Allright, it looks like this is possible without creating a plugin - with psalm conditional return statements.
However, there seems something wrong with casting named groups to `dict(array_key(), string())` : it transforms it to an int instead of a string. Will have to look into that issue another day

veewee:
@azjezz Fixed it without using a psalm plugin. Instead the conditional return types of psalm did the trick together with a fallback to a custom `$capture_groups` fallback type.

veewee:
Rebased it to 1.6.X. Let me know if there is anything else I can do in order to finish this one!

veewee:
@azjezz  How do you want to continue with this PR? Is it something you want to include? Because if so, I need to move a part to the new plugin repository.

azjezz:
yes, let's move things to the new plugin, merge changes there, then we merge this.

veewee:
Allright, will do it soon - but having some time constraints due to yet another lockdown here.
Hopefully something for next week.

veewee:
Added the PR to the psalm plugin, but It will conflict with the open issue - so it's not completely done.
https://github.com/php-standard-library/psalm-plugin/pull/5

Next rebased against 1.7.x - shall I keep the psalm plugin changes in here as well or remove it from the PR?

azjezz:
> Next rebased against 1.7.x - shall I keep the psalm plugin changes in here as well or remove it from the PR?

remove it, we don't want to add anything to the old plugin, just keeping it to not break BC.

php-standard-library/psalm-plugin#3 has been merged, so i guess you can rebase php-standard-library/psalm-plugin#5

tests/Psl/Regex/EveryMatchTest.php
```diff
@@ -0,0 +1,191 @@
...
+        yield [
+            [
+                [
+                    0 => 'PHP',
+                    'language' => 'PHP'
+                ],
+            ],
+            'PHP is the web scripting language of choice.',
+            '/(?P<language>PHP)/',
+            capture_groups([0, 'language'])
```
azjezz: `0` should be added by capture_groups, right?
```suggestion
            capture_groups(['language'])
```

tests/Psl/Regex/FirstMatchTest.php
```diff
@@ -0,0 +1,125 @@
...
+    public function provideMatchingData(): iterable
+    {
+        yield [
+            [
+                0 => 'PHP',
+                1 => 'PHP',
+            ],
+            'PHP is the web scripting language of choice.',
+            '/(php)/i',
+            capture_groups([0, 1])
```
azjezz: same here ( and the next ones )

azjezz:
overall, this LGTM.

I'll probably do another review later and merge if everything is okay.

we can release this, along side #138 in 1.7 :)

tests/Psl/Regex/EveryMatchTest.php
veewee: Yes, it is added by default.
Doubles are filtered out, so it doesnt really matter if you add it or not - besides being more explicit.

azjezz (approved):

PR #152, opened 2021-03-06
[Type] add optional, and literal types

azjezz:
example:

```php
$result = Json\typed($content, Type\shape([
  'username' => Type\non_empty_string(),
  'password' => Type\non_empty_string(),
  'country' => Type\optional(Type\non_empty_string())
]);

// $result = array{username: non-empty-string, password: non-empty-string, country?: non-empty-string}

takeNonEmptyString($result['username']); // ok
takeNonEmptyString($result['password']); // ok
takeNonEmptyString($result['country']); // error: 'country' might not be defined.
takeNonEmptyString($result['country'] ?? 'foo'); // ok
```

---

TODO:
- [ ] add more tests
- [x] ensure that all types cannot contain optional types - e.g: map, vector, vec, dict .. etc ( optional should always be the outermost )

---

### features
- added `Type\optional()` function that returns an optional type which can be used with `Type\shape`
- added `Type\TypeInterface::isOptional` method

### backward compatibility breaking changes:
- a new method `isOptional` was added to `Type\TypeInterface` interface
- `Type\nullable(_)->toString()` now uses `?T` notation instead of `null|T`.


src/Psl/Type/Internal/ShapeType.php
```diff
@@ -28,12 +23,15 @@ final class ShapeType extends Type\Type
...
+    private bool $allow_unknown_fields;
+
     /**
      * @param array<Tk, Type\TypeInterface<Tv>> $elements_types
      */
-    public function __construct(array $elements_types)
+    public function __construct(array $elements_types, bool $allow_unknown_fields = false)
     {
         $this->elements_types = $elements_types;
+        $this->allow_unknown_fields = $allow_unknown_fields;
```
veewee, line 34: What should happen psalm type-wise if you set this prop to true in the shape function?
Should it be typed as the shape you provide or as the more generic default as templates above the class?

azjezz: currently nothing. this is more of a runtime configuration.

veewee: Alright, so if you access an undefined key directly, psalm will lose track of the value type and you'll need to set it manually?

azjezz: this better explains it:
```php
$result = Json\typed($content, Type\shape([
  'username' => Type\non_empty_string(),
  'password' => Type\non_empty_string(),
  'country' => Type\optional(Type\non_empty_string())
], true));

// $result = array{username: non-empty-string, password: non-empty-string, country?: non-empty-string}

$username = $result['username']; // psalm: non-empty-string
$password  = $result['password']; // psalm: non-empty-string
$country = $result['country']; // psalm error: 'country' might not be defined.
$country  = $result['country'] ?? null; // psalm: non-empty-string|null
$bar = $result['foo']; // psalm error: 'foo' is not defined
$bar = $result['foo'] ?? null; // psalm error: unable to infer the type $bar is being assigned to.
/** @var string|null $bar */
$bar = $result['foo'] ?? null; // psalm: string|null
```

at the end, psalm will still remember that `$result` is `array{username: non-empty-string, password: non-empty-string, country?: non-empty-string}`, however, accessing fields that are not specified in the original type will result in an error, using `?? null` ( or `isset($shape['foo'])` ), psalm will tell you about another error ( unable to infer the type the variable is being assigned to ), so in that case, you have to either: 

a. use `@var`
b. use `Type\` function to assert it's type.

veewee: Alright, exactly what I had in mind. makes sense! Thanks for the info

PR #153, opened 2021-03-07
Add doc gen, and check committed docs are up-to-date.

### LINKED ISSUE #64: Documentation
We should probably have documentation that people can use to lookup functions and their descriptions + a couple of examples.

Danack:
`php doc_gen.php` updates the docs.

The check in CI only compares the checked-in version against what would be generated by the current code.

Having it be automatically committed by a github action results in an extra commit all the time, which is far too annoying.


azjezz:
oh, i like this! 1 file containing all symbols with a link to them is really nice!

i think we can update this over the time, and improve it ( e.g: ignore internal symbols, add function signature .. etc )

I'll take a better look tomorrow, but i think we can merge this as is.

doc_gen.php
```diff
@@ -0,0 +1,164 @@
+<?php
```
azjezz: i think we can move this to `doc/generate.php`, and add it to [.php_cs.dist](https://github.com/azjezz/psl/blob/1.5.x/.php_cs.dist#L7) and [.phpcs.xml](https://github.com/azjezz/psl/blob/1.5.x/.phpcs.xml#L6) to ensure it follows our CS rules.

doc_gen.php
```diff
@@ -0,0 +1,164 @@
...
+        $link,
+        $functionWithoutNamespace
+    );
+
+    return $line;
+}
+
+function generateDocList(): string
+{
+    $version = '1.5.x';
```
azjezz: since this will always be ran against the default branch ( upcoming release ), i think we can use the API to retrieve the default branch.

```suggestion
    $json = file_get_contents('https://api.github.com/repos/azjezz/psl');
    $version = Json\typed($json, Type\shape([
      'default_branch' => Type\non_empty_string(),
    ]))['default_branch'];
```

also `use Psl\Json` and `use Psl\Type` :)

however, we might wanna run this against other branches, so we can probably add `--sha` argument for it in the future.

doc_gen.php
```diff
@@ -0,0 +1,164 @@
...
+
+    require_once __DIR__ . "/vendor/autoload.php";
+
+    $contents = [];
```
veewee: Maybe out of scope for this PR : but it would be nice to group functions, interfaces and classes per namespace.
Imagine you want to see what is in the result namespace : Now it's a bit all over the place.

doc_gen.php
```diff
@@ -0,0 +1,164 @@
...
+
+Please click through to read the docblock comment details for each of them.
+
+MD;
+
+$contents .= generateDocList();
+
+function checkDocsUpToDate($docFilename, $contents)
+{
+    $newLines = explode("\n", $contents);
```
veewee: You can also use `Psl\Shell\execute()` with `git diff --exit-code $docFilename`.
This gives you a nice diff of the changed lines and returns the correct exit code.
Meaning you won't have to count the lines and check for differences.

doc_gen.php
veewee: Currently all tasks are also in `composer.json`. Maybe add a shortcut in there as well and add it to the `composer ci` task?

doc_gen.php
azjezz: they already are, no? https://github.com/Danack/psl/blob/doc_gen_squash/docs/index.md

veewee: The main group now is functions, interfaces and classes.
Next they are grouped per namespace.

What I meant was group them by namespace first and next per type : functions, interfaces, classes

azjezz: done

azjezz:
Thank you for initiating the work on this @Danack, i went ahead a finished this PR!

PR #155, opened 2021-03-15
replace `@use` annotation by `@see`

b1rdex:
`@use` is used in psalm/phpstan for generic traits: https://psalm.dev/docs/annotating_code/templated_annotations/#template-inheritance

### DIFF
--- a/docs/component/arr.md
+++ b/docs/component/arr.md
@@ -49,7 +49,7 @@
 - [slice](./../../src/Psl/Arr/slice.php#L35) ( deprecated )
 - [sort](./../../src/Psl/Arr/sort.php#L24) ( deprecated )
 - [sort_by](./../../src/Psl/Arr/sort_by.php#L27) ( deprecated )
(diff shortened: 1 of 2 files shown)


azjezz:
Hey @b1rdex, thank you for the PR; but i want to understand the reason behind it, we use `@use` in multiple places, not just here, so if we are replacing it, we should replace it everywhere.

does phpstan complain about this? as psalm doesn't seem to do that currently.

b1rdex:
I think this is a bug introduced during refactoring. Check out the full commit diff that brought the @use.
Also, there is no point in replacing @see with @use as it isn’t supported by phpstorm or any other phpdoc reader. 

Отправлено с iPhone

azjezz:
@b1rdex can you please rebase this on 1.6.x and replace all other use cases of `@use` by `@see`?

b1rdex:
@azjezz I don't see any other `@use` usages. Could you please verify?

azjezz:
correct, i think we are already using `@see` everywhere else.

thank you 👍

PR #156, opened 2021-03-17
split documentation

azjezz:


azjezz:
also better checks!

image

docs/documenter.php
```diff
@@ -179,9 +177,9 @@ static function (string $symbol) use ($namespace): bool {
 /**
  * @return list<string>
  */
-function get_all_namespaces(): array
+function get_all_components(): array
 {
-    return [
+    $components = [
```
veewee, line 182: Future improvement is to try to guess all namespaces based on the folder structure (whilst removing the internal once).

docs/documenter.php
```diff
@@ -122,7 +111,7 @@ function get_namespace_documentation(string $namespace): string
                 }
 
                 $definition_line = get_symbol_definition_line($symbol, $type);
```
veewee, line 113: The definition line for constants always return 0
Maybe we could use better-reflection to guess constants lines in the future.

https://github.com/Roave/BetterReflection/blob/478ed89e695cbff3ac037ccaa713cb56aee359b4/src/Reflector/ConstantReflector.php#L54

veewee (approved):
Very nice and Love the diff!
The documentor code really shows the strength of this package.
Added some possible future improvements as a comment.

docs/documenter.php
azjezz: The main issue is with things like `Psl\Str\Byte`, `Psl\Encoding\Hex` .. etc.

docs/documenter.php
azjezz: maybe later, for now it's not really important.

PR #157, opened 2021-03-18
[Type] string to float coerce

### LINKED ISSUE #146: [Type] Coerce float to string
Currently, it is not possible to convert a float -> string.
https://github.com/azjezz/psl/blob/c34f75bcb27298a1fdb017c07cd90f07c78046e2/src/Psl/Type/Internal/StringType.php#L39

The other way around (string -> float) is possible.

We could go for one of these solutions to make it possible in a locale-safe way:

- https://github.com/moneyphp/money/blob/512333ef834bd77f76e865a30eb5963d480e2ed7/src/Number.php#L64-L71
- https://github.com/brick/math/blob/49d9d6d83d9f97ebc6e8f3d7a4adbc7cd649afc2/src/BigNumber.php#L165-L175

Would it make sense to add it and do you see any additional risks?

veewee:
Fixes  #146

### DIFF
--- a/tests/Psl/Type/ArrayKeyTypeTest.php
+++ b/tests/Psl/Type/ArrayKeyTypeTest.php
@@ -21,12 +21,12 @@ final class ArrayKeyTypeTest extends TypeTest
         yield ['123', '123'];
         yield ['1e23', '1e23'];
         yield [$this->stringable('123'), '123'];
-        yield [1.0, 1];
+        yield [1.0, '1.0'];
+        yield [1.23, '1.23'];
     }
 
     public function getInvalidCoercions(): iterable
     {
-        yield [1.23];
         yield [[]];
         yield [[1]];
         yield [Type\bool()];
--- a/src/Psl/Type/Internal/StringType.php
+++ b/src/Psl/Type/Internal/StringType.php
@@ -4,10 +4,12 @@ declare(strict_types=1);
 
 namespace Psl\Type\Internal;
 
+use Psl\Str;
 use Psl\Type;
 use Psl\Type\Exception\AssertException;
 use Psl\Type\Exception\CoercionException;
 
+use function is_float;
 use function is_int;
 use function is_object;
 use function is_string;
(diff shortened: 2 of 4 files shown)


tests/Psl/Type/ArrayKeyTypeTest.php
```diff
@@ -21,12 +21,12 @@ public function getValidCoercions(): iterable
         yield ['123', '123'];
         yield ['1e23', '1e23'];
         yield [$this->stringable('123'), '123'];
-        yield [1.0, 1];
+        yield [1.0, '1.0'];
```
veewee, line 24: This might be a BC break?

azjezz: hm, yes, and i don't think we should do this for array keys, we should probably flip the array key union type to check  for int first, then string.

veewee: In that case, line 21 will return an integer instead of a string. Which is also not correct :)

azjezz: It shouldn't, array-key is a union of `string` and `int` type, and union type should always try to assert that the value is one of either types before trying do any coercing, so if line 21 fails, it's a bug. https://github.com/azjezz/psl/blob/1.5.x/src/Psl/Type/Internal/UnionType.php#L69-L75

veewee: I'm sorry, it is the one on line 23 : a stringable object that returns a numeric string will return an int.
Since the integer type is also coercing stringables.

azjezz: Hm, maybe we shouldn't extend union type for array-key then ( this is actually the exact same problem why i didn't finish https://github.com/php/php-src/pull/4073, i didn't know how to handle this case 😆 )

azjezz: also, php doesn't support stringable -> int, and i don't think we should either: https://3v4l.org/0G6St

azjezz:
@veewee do you want to finish this for 1.7 or should i just tag 1.7 now and leave this for later?

veewee:
Gonna close this PR due to inactivity: The approach needs a full rewrite as mentioned in the comments. The issue is still open, so feel free to re-implement anytime.

PR #159, opened 2021-03-20
improve metaphone errors

ntzm:
As of PHP 8, `metaphone` no longer returns `false` on error: https://github.com/php/php-src/commit/84be22f1f5db499001379f762869d00fc3bc6f55

Previously, the only way `false` could be returned is the `phonemes` parameter to be negative, which this PR handles.

I guess this could considered be a BC beak, as if a user is on PHP 7.4, and was passing in an invalid phoneme parameter, it will throw instead of returning `null` now.


azjezz (approved):

azjezz:
personally i don't mink small BC breaks in minor releases 👍 

Thank you

PR #160, opened 2021-03-20
[Str][Byte] simplify reverse implementation

ntzm:


src/Psl/Str/Byte/reverse.php
```diff
@@ -9,14 +9,5 @@
...
-    $hi = namespace\length($string) - 1;
-
-    for (; $lo < $hi; $lo++, $hi--) {
-        $temp        = $string[$lo];
-        $string[$lo] = $string[$hi];
-        $string[$hi] = $temp;
-    }
-
-    return $string;
+    return strrev($string);
```
azjezz, line 16: please add `use function strrev` statement at the top :) 

Also, if you could add a line in the docblock explaining what this function does

azjezz (changes requested):

src/Psl/Str/Byte/reverse.php
ntzm: Done!

azjezz (approved):
Thanks @ntzm

PR #161, opened 2021-03-20
[Asio] Introduce Async I/O component

azjezz:


src/Psl/Asio/Exception/TimeoutException.php
```diff
@@ -0,0 +1,15 @@
...
+
+declare(strict_types=1);
+
+namespace Psl\Asio\Exception;
+
+use Psl\Exception;
+
+final class TimeoutException extends Exception\RuntimeException implements ExceptionInterface
+{
+    public function __construct(string $message = 'operation timed out.', int $code = 0)
```
BackEndTea, line 9: Should the exception expose how long it took?

azjezz: probably, will see ( this PR is not getting merged until PHP 8.1 is released, so there's still time to change the API 😛 )

src/Psl/Asio/stub.php
```diff
@@ -0,0 +1,49 @@
+<?php
+
+declare(strict_types=1);
+
+if (false) {
+    final class Fiber
+    {
+        public function __construct(callable $c)
+        {
+        }
```
azjezz: TODO: remove stub before merge

src/Psl/Asio/Awaitable.php
```diff
@@ -0,0 +1,65 @@
...
+     *  (callable(?Throwable, ?T): Awaitable<mixed>)|
+     *  (callable(?Throwable): void)|
+     *  (callable(?Throwable): Awaitable<void>)|
+     *  (callable(?Throwable, ?T): void)|
+     *  (callable(?Throwable, ?T): Awaitable<void>)
+     * ) $callable
+     *
+     * @internal
+     */
+    public function onJoin(callable $callable): void;
```
veewee, line 64: Where does this method name come from?
In amp, they use on resolve, which is more clear IMO.

azjezz: This method ( maybe the whole class? ) is going to be removed soon thanks to some clever work by @:trowski and @:kelunik.

veewee: Cool! Is it a loop / promise rfc to php-src perhaps? ;)

PR #163, opened 2021-03-23
fix psalm issues

azjezz:


azjezz:
@muglug 

something is weird about this error, it's been present for a while, but never caught in CI, it only happens locally, i assume you encountered this error as well because i can see it was added in your baseline here: https://github.com/muglug/psl/blob/1.6.x/psalm-baseline.xml#L4

image

psalm.xml
```diff
@@ -59,6 +59,10 @@
         <RedundantCondition errorLevel="suppress"/>
         <RedundantCast errorLevel="suppress"/>
         <RedundantCastGivenDocblockType errorLevel="suppress"/>
+
+        <!-- Not using the result of pure functions is common within PSL -->
+        <!-- e.g: with and Psl\invariant_violations() -->
+        <UnusedFunctionCall errorLevel="suppress" />
```
azjezz, line 65: not sure if this should be considered a bug in psalm or not.

1. pure function are considered to have no side effects
2. calling a function with no side effects without using it's results is useless
3. psalm consider throwing an exception not a side effect
4. hacklang considers throwing an exception a side effect ( https://github.com/facebook/hhvm/blob/master/hphp/hack/doc/HIPs/contexts_and_coeffects.md )
5. if throwing an exception is a side effect, this is most likey to result in all functions being non pure, as 90% of calls result in an exception
6. if throwing an exception is not a side effect, how can a pure function tell psalm that it's okay not to use it's result?

wdyt @muglug?
 
/cc @ocramius

muglug:
Do you think `UnusedVarAnnotation` should be a level 1 Psalm issue? Maybe I should backtrack and only make `UnusedVariable` (which can have other ramifications) a level 1 issue

azjezz:
I think `UnusedVarAnnotation` should be a level 7 issue, it's easy to solve.
`UnusedVariable` on the other hand should be a level 1 issue, the change would be to add `_` as a prefix to the variable name, and people might not want to bother with it if they just added psalm to their codebase.

I previously added those `@var` annotations because psalm wasn't detecting the type correctly, so it makes sense to remove them.

muglug:
Yeah, I guess people can always suppress them

PR #165, opened 2021-03-24
remove unused function call suppression

azjezz:


src/Psl/Hash/Context.php
```diff
@@ -88,6 +88,7 @@ public static function hmac(string $algorithm, string $key): Context
     public function update(string $data): Context
     {
         $internal_context = hash_copy($this->internalContext);
+        /** @psalm-suppress UnusedFunctionCall */
         hash_update($internal_context, $data);
```
azjezz: should be fixed in psalm, dismissing the return value of pure functions with by-reference arguments should be okay.

azjezz: fixed in https://github.com/vimeo/psalm/pull/5463

azjezz: hmm, actually `hash_update` is not pure, `$internal_context` is an instance of `HashContext`, which is being modified in this case.

meaning that `hash_update` is impure, and is mutating `$internal_context` 🤔

azjezz: hash_update returns a bool, so i think it's better to wrap this in Psl\invariant to ensure that the context is being actually updated.

Issue #167, opened 2021-03-25
Performance of reimplemented functions?


ghost:
I understand that a completely reimplemented function in userland code will be slower than the compiled C counterpart, yet I thought that for many functions PSL simply implemented a wrapper for native functions, thus having little overhead.

I got a little surprised when I was replacing some instances of `\array_unique` of suddenly getting a measurable drop in performance (those are rather large arrays, but that's what I get).

[Here](https://phpsandbox.io/n/shrill-wood-lolj-cehre) I can replicate it with this code:

```php
$arr = [];
$iterations   = 10;
$array_size   = 1000;
$random_range = 5000;

for ($i = 0; $i < $array_size; $i++) {
    $arr[] = Psl\SecureRandom\int(0, $random_range);
}

$total = 0;
$ta0 = microtime(true);
for ($e = 0; $e < $iterations; $e++) {
    $total += \count(\array_unique($arr));
}
$ta1 = microtime(true);

echo "Native\n";
echo $total / $iterations, "\n";
echo 'Time: ' . round($ta1 - $ta0, 4), "\n\n";

$total = 0;
$ta0   = microtime(true);
for ($e = 0; $e < $iterations; $e++) {
    $total += \Psl\Iter\count(\Psl\Dict\unique($arr));
}
$ta1 = microtime(true);

echo "Psl\n";
echo $total / $iterations, "\n";
echo 'Time: ' . round($ta1 - $ta0, 4);
```

The result in this case is 0.0009 seconds for native implementations, against 0.0841 seconds for using Psl. True, for this case 0,08 seconds is still something I can live with, but it's still a very marked cost increase. For my particular use case can even become problematic.

Is this what's expected, or I'm using the library incorrectly? Mind, this is not in no way a complaint, I love the library even if this is a tradeoff to be aware of, but just wanted to be sure this was not me "holding it wrong".

azjezz:
this is expected, if you are using PHP >= 7.4, i recommend including `vendor/azjezz/psl/src/preload.php` in your preload file, or using it as a preload file if you don't have anything else to preload, which should give better performance as functions will be cached after the first run ( assuming you are using php-fpm or fastcgi ), if it's a CLI script, i cannot recommand anything that would improve performance here.

azjezz:
Hm, after https://phpsandbox.io/n/wispy-leaf-75nn-pnldg, i noticed that the unqiue function is causing the performance issue here, not `count` ( which has almost the same performance ).

I think there's a way to fix this.

[...]

PR #168, opened 2021-03-25
Dict\unique_scalar() function to avoid performance penalty hit when using Dict\unique() on large scalar arrays

ghost:
this avoids a rather large performance hit. (#167)

### DIFF
--- a/src/Psl/Dict/unique.php
+++ b/src/Psl/Dict/unique.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Psl\Dict;
 
+use array_unique;
+use is_array;
+
 /**
  * Returns a new dict in which each value appears exactly once.
  *
@@ -17,8 +20,8 @@ namespace Psl\Dict;
 function unique(iterable $iterable): array
 {
     
-    if (\is_array($iterable)) {
-        return \array_unique($iterable);
+    if (is_array($iterable)) {
+        return array_unique($iterable);
     }
     
     return unique_by(
--- a/src/Psl/Dict/unique.php
+++ b/src/Psl/Dict/unique.php
@@ -4,9 +4,6 @@ declare(strict_types=1);
 
 namespace Psl\Dict;
 
-use array_unique;
-use is_array;
-
 /**
  * Returns a new dict in which each value appears exactly once.
  *
@@ -19,11 +16,7 @@ use is_array;
  */
 function unique(iterable $iterable): array
 {
-    
-    if (is_array($iterable)) {
-        return array_unique($iterable);
-    }
-    
+
     return unique_by(
         $iterable,
         /**
--- a/src/Psl/Dict/unique.php
+++ b/src/Psl/Dict/unique.php
@@ -16,6 +16,11 @@ namespace Psl\Dict;
  */
 function unique(iterable $iterable): array
 {
+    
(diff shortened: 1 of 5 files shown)


src/Psl/Dict/unique.php
```diff
@@ -16,6 +16,11 @@
  */
 function unique(iterable $iterable): array
 {
+    
+    if (\is_array($iterable)) {
+        return \array_unique($iterable);
+    }
```
azjezz: please import functions instead using `\` ( `use is_array;` and `use array_unique;` )

azjezz:
Ah, this fails when values are not scalar ( which is something that we do support ) 🤔 

how about we add a new function instead ( e.g: `Dict\unique_scalar` )?

ghost:
Gave it a try to `unique_scalar()`.

src/Psl/Dict/unique_scalar.php
```diff
@@ -0,0 +1,35 @@
...
+
+use function is_array;
+use function array_unique;
+
+/**
+ * Returns a new dict in which each value appears exactly once.
+ *
+ * @psalm-param iterable<array-key, scalar> $iterable
+ *
+ * @psalm-return array<scalar>
```
azjezz: we don't use `@psalm-` prefix, and we always try to keep the return type narrowed as much as possible 😉 

```suggestion
 * @template Tk of array-key
 * @template Tv of scalar
 *
 * @param iterable<Tk, Tv> $iterable
 *
 * @return array<Tk, Tv>
```

src/Psl/Dict/unique_scalar.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+function unique_scalar(iterable $iterable): array
+{
+
```
azjezz: extra line.
```suggestion
```

src/Psl/Dict/unique.php
```diff
@@ -16,6 +16,7 @@
  */
 function unique(iterable $iterable): array
 {
+
```
azjezz: extra new line
```suggestion
```

tests/Psl/Dict/UniqueScalarTest.php
```diff
@@ -0,0 +1,25 @@
...
+
+    public function testUniqueScalars(): void
+    {
+        $array   = Vec\fill(10, 'foo');
+        $array[] = 'bar';
+
+        $unique = Dict\unique_scalar($array);
+
+        static::assertCount(2, $unique);
+        static::assertSame('foo', Iter\first($unique));
```
azjezz: we should also add a test case for other Iterators ( use `Iter\Iterator` class ), and IteratorAggregates ( use `Collection\Map` or any other collection )

```suggestion
        static::assertCount(2, $unique);
        static::assertSame(['foo', 'bar'], $unique);
        static::assertSame('foo', Iter\first($unique));
```

azjezz (commented):
few comments.

also, please regenerate the documentation using `php docs/documentor.php`

ghost:
Let me know if I can do anything else. I'll be rather busy until for a few hours, but I can come back to this later today or tomorrow. Thanks for all the feedback.

azjezz (approved):
this is perfect. thanks 👍

PR #174, opened 2021-03-28
move tools to a sub directory

azjezz:


azjezz:
@veewee let me know what you think.

locally, you would use `make install` to install all dependencies, and `make check` to run all checks.

azjezz:
i think we need to add CONTRIBUTING.md file to explain this process to people who would like to contribute, as they might not be familiar with it.

veewee:
Looks OK to me!

PR #177, opened 2021-04-06
Added positive int as type

michaelpetri:
Nothing special just added support for positive integers, would create another patch for natural integers afterwards.


michaelpetri:
`make coding-standard-fix` always "fixes" my psalm-return annotation from:
```php
    /**
     * @param mixed $value
     *
     * @throws CoercionException
     *
     * @psalm-return positive-int
     */
    public function coerce($value): int
```
to
```php
    /**
     * @param mixed $value
     *
     * @throws CoercionException
     *
     * @return positive-int
     */
    public function coerce($value): int
```

any idea?

azjezz:
That's intentional, we don't use psalm- prefix :)

michaelpetri:
Removed psalm prefix, can you share some insights why you don't use them?

azjezz:
annotations which we don't prefix with psalm are also supported by phpstan, and phpstorm ( recently started supporting lowercase-string, positive-int, no-return .. etc ), so there's no need for the use of the `psalm-` prefix IMHO.

azjezz (approved):
Thank you @michaelpetri

PR #184, opened 2021-04-26
Feature/variadic union type improvemence

### LINKED ISSUE #183: Union for variadic types
There are a lot of situations may occur when we need to define union from several cases.
For example: need to define enum of states.

This would look like:
```php
use Psl\Type as T;

$codec = T\shape([
    // ... rest of fields
    'state' => T\union(
        T\literal_scalar('NEW'),
        T\union(
            T\literal_scalar('APPROVED'),
            T\union(
                T\literal_scalar('REJECTED'),
                T\union(
                    T\literal_scalar('COMPLETED'),
                    T\union(
                        T\literal_scalar('FROZEN'),
                        T\literal_scalar('ERROR')
                    )
                )
            )
        )
    ),
]);
```

Inferred type would be:
```
array{id: string, state: "APPROVED"|"COMPLETED"|"ERROR"|"FROZEN"|"NEW"|"REJECTED"}
```

---

I guess it would be great to have better solution like:
```php
use Psl\Type as T;

$codec = T\shape([
    'id' => T\string(),
    'state' => T\union_of(
        T\literal_scalar('NEW'),
        T\literal_scalar('APPROVED'),
        T\literal_scalar('REJECTED'),
        T\literal_scalar('COMPLETED'),
        T\literal_scalar('FROZEN'),
        T\literal_scalar('ERROR'),
    ),
]);
```

I do understand that `Psl\Type\Internal\UnionType` has only 2 branches: **left** and **right**
That's why propose just implement a function with signature:
```php
/**
 * @template T
 *
 * @param TypeInterface<T> $left_type
 * @param TypeInterface<T> $right_type
 * @param TypeInterface<T> ...$others
 * @return TypeInterface<T>
 *
 * @no-named-arguments
 */
function union_of(
    TypeInterface $left_type,
    TypeInterface $right_type,
    TypeInterface ...$others
): TypeInterface { // actually Psl\Type\Internal\UnionType returns here
    ...
}
```

---

As alternatives I may guess there is more suitable name for function.
This may be named like `literals` or `one_of` or etc.

This function may also be implemented in user land, but I think it would be great to have unified solution inside core of the library.

---

Want to hear other guys thoughts.
If this would be approved, I may produce PR.
As I can see, this is very light enhancement.

nzour:
#183 

### Changelog
- function `Psl\Type\union` now takes variadic third parameter
- function `Psl\Type\union` arguments `$left_type` and `$right_type` were renamed to `$first` and `$second`
    This is a **breaking change**, because if somebody uses PHP8+ and does 
    `union(left_type: $left, right_type: $right)` - this would trigger runtime `Error`.
    
### Future scope
- I think we should use `@no-named-arguments` [psalm's annotation](https://psalm.dev/docs/annotating_code/supported_annotations/#no-named-arguments)
    I haven't used this in my PR, cause I think it should be discussed and (if accepted) added atomically for all possible functions and methods
- As @azjezz mentioned in #183, we should also make variadic function `Psl\Type\intersection`

### DIFF
--- a/src/Psl/Type/literal_scalar.php
+++ b/src/Psl/Type/literal_scalar.php
@@ -18,7 +18,7 @@ use Psl;
 function literal_scalar($value): TypeInterface
 {
     /** @psalm-suppress MissingThrowsDocblock */
-    union(union(string(), bool()), num())->assert($value);
+    union(string(), bool(), num())->assert($value);
 
     return new Internal\LiteralScalarType($value);
 }
--- a/src/Psl/Type/union.php
+++ b/src/Psl/Type/union.php
@@ -7,19 +7,26 @@ namespace Psl\Type;
 use Psl;
 
 /**
- * @template Tl
- * @template Tr
+ * @template T
  *
- * @param TypeInterface<Tl> $first
- * @param TypeInterface<Tr> $second
+ * @param TypeInterface<T> $first
+ * @param TypeInterface<T> $second
+ * @param TypeInterface<T> ...$rest
  *
- * @throws Psl\Exception\InvariantViolationException If $first, or $second is optional.
+ * @throws Psl\Exception\InvariantViolationException If $first, $second or one of $rest is optional.
  *
- * @return TypeInterface<Tl|Tr>
+ * @return TypeInterface<T>
  */
 function union(
     TypeInterface $first,
-    TypeInterface $second
+    TypeInterface $second,
+    TypeInterface ...$rest
 ): TypeInterface {
-    return new Internal\UnionType($first, $second);
+    $accumulated_type = new Internal\UnionType($first, $second);
+
+    foreach ($rest as $type) {
+        $accumulated_type = new Internal\UnionType($accumulated_type, $type);
+    }
+
+    return $accumulated_type;
 }
--- a/src/Psl/Type/union.php
+++ b/src/Psl/Type/union.php
@@ -10,16 +10,16 @@ use Psl;
  * @template Tl
  * @template Tr
  *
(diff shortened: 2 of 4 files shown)


nzour:
I think I've chosen wrong branch to merge to. @azjezz do you have 2.0 branch?

azjezz:
As mentioned in the issue, renaming parameters is not considered a BC break ( for me ), we can merge this into 1.7

azjezz:
@zour228 can you apply the same change to intersection? I think these two should stay consistent with each other.

nzour:
> @zour228 can you apply the same change to intersection? I think these two should stay consistent with each other.

Sure, but maybe next day, it's late time on my watch

nzour:
About named parameters and BC:
I still think it's important to mention parameters renaming in changelog.
To prevent unexpected error that somebody may have

nzour:
@azjezz didn't know that psalm has an issues with intersection types (even with such a simple example)
https://psalm.dev/r/737e004fcd

Therefore the following example have type inference issues too:
```php
$first = shape(['first' => string()]);
$second = shape(['second' => string()]);

$res = intersection($first, $second);
// inferred type of $res is TypeInterface<array{first: string}>
// but I was expecting TypeInterface<array{first: string, second: string}>
```

So, I am not sure if I know hot to implement function with variadic parameters.
But maybe type inference is not main feature here?

azjezz:
Intersection types work only with objects

azjezz (approved):
LGTM, you can send a separate PR for intersection types later on.

Thank you 🎉

PR #185, opened 2021-04-27
Feature/variadic intersection type

### LINKED ISSUE #183: Union for variadic types

nzour:
Union variadic types issue #183 and PR #184

### DIFF
--- a/tests/static-analysis/Type/intersection.php
+++ b/tests/static-analysis/Type/intersection.php
@@ -36,9 +36,6 @@ function test(): void
     /** @psalm-suppress MissingThrowsDocblock */
     takes_valid_intersection($old_school_codec->assert('any'));
 
-    /**
-     * @psalm-suppress MissingThrowsDocblock
-     * @psalm-suppress InvalidArgument
-     */
+    /** @psalm-suppress MissingThrowsDocblock */
     takes_valid_intersection($new_codec->assert('any'));
 }
--- a/tests/static-analysis/Type/intersection.php
+++ b/tests/static-analysis/Type/intersection.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+use Psl\Collection\Map;
+use Psl\Result\ResultInterface;
+use Psl\Type;
+
+/**
+ * @psalm-suppress UnusedParam
+ *
+ * @param Map&ResultInterface&stdClass $value
+ */
+function takes_valid_intersection($value): void
+{
+}
+
+function test(): void
+{
+    /** @psalm-suppress MissingThrowsDocblock */
+    $old_school_codec = Type\intersection(
+        Type\object(Map::class),
+        Type\intersection(
+            Type\object(ResultInterface::class),
+            Type\object(stdClass::class),
+        ),
+    );
+
+    /** @psalm-suppress MissingThrowsDocblock */
+    $new_codec = Type\intersection(
+        Type\object(Map::class),
+        Type\object(ResultInterface::class),
+        Type\object(stdClass::class),
+    );
+
+    /** @psalm-suppress MissingThrowsDocblock */
+    takes_valid_intersection($old_school_codec->assert('any'));
+
+    /**
+     * @psalm-suppress MissingThrowsDocblock
+     * @psalm-suppress InvalidArgument
+     */
+    takes_valid_intersection($new_codec->assert('any'));
+}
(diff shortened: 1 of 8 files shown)


tests/static-analysis/Type/intersection.php
```diff
@@ -0,0 +1,41 @@
...
+            Type\object(stdClass::class),
+        ),
+    );
+
+    /** @psalm-suppress MissingThrowsDocblock */
+    $new_codec = Type\intersection(
+        Type\object(Map::class),
+        Type\object(ResultInterface::class),
+        Type\object(stdClass::class),
+    );
```
azjezz: hm, can you test this with 4 objects? i don't think it's going to work....

tests/static-analysis/Type/intersection.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+use Psl\Collection\Map;
+use Psl\Result\ResultInterface;
+use Psl\Type;
+
+/**
+ * @psalm-suppress UnusedParam
```
azjezz, line 11: p.s: you can rename the argument to `$_value` and psalm won't annoy you about unused argument ;)

tests/static-analysis/Type/intersection.php
azjezz: hm, it seems to work here https://psalm.dev/r/433e9e351b, not sure why i thought it wouldn't ...

azjezz:
could you please add a unit test to create an intersection with more than 2 types? ( to keep code coverage the same )

tests/static-analysis/Type/intersection.php
nzour: Sure, also check https://psalm.dev/r/e1d0238b41

nzour:
> could you please add a unit test to create an intersection with more than 2 types? ( to keep code coverage the same )

Yep, I'll do it

tests/unit/Type/IntersectionTypeTest.php
```diff
@@ -53,14 +53,36 @@ public function getToStringExamples(): iterable
...
+            '(Psl\Collection\CollectionInterface|Iterator)&Psl\Collection\IndexAccessInterface'
+        ];
+
+        yield [
+            Type\intersection(
+                Type\object(IndexAccessInterface::class),
+                Type\object(CollectionInterface::class),
+                Type\object(Iterator::class),
+                Type\shape(['id' => Type\string()]),
+            ),
```
azjezz, line 84: technically an impossible type, but okay :p

azjezz (approved):

azjezz:
Thank you @zour228 !

PR #193, opened 2021-05-17
[RandomSequence] Introduce random sequence component

azjezz:
This PR provides an object scoped RNG implementation in pure PHP.

see RFC: https://wiki.php.net/rfc/object_scope_prng

the main difference between this and the proposed RFC is that PSL provides a less cluttered API.

### DIFF
--- a/docs/component/random-sequence.md
+++ b/docs/component/random-sequence.md
@@ -1,0 +1,23 @@
+<!--
+    This markdown file was generated using `docs/documenter.php`.
+
+    Any edits to it will likely be lost.
+-->
+
+[*index](./../README.md)
+
+---
+
+### `Psl\RandomSequence` Component
+
+#### `Interfaces`
+
+- [SequenceInterface](./../../src/Psl/RandomSequence/SequenceInterface.php#L7)
+
+#### `Classes`
+
+- [MersenneTwisterPHPVariantSequence](./../../src/Psl/RandomSequence/MersenneTwisterPHPVariantSequence.php#L10)
+- [MersenneTwisterSequence](./../../src/Psl/RandomSequence/MersenneTwisterSequence.php#L10)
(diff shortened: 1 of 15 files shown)


docs/component/random-sequence.md
```diff
@@ -0,0 +1,23 @@
...
+    This markdown file was generated using `docs/documenter.php`.
+
+    Any edits to it will likely be lost.
+-->
+
+[*index](./../README.md)
+
+---
+
+### `Psl\RandomSequence` Component
```
veewee, line 11: Would it make sense to put it in the *random namespaces that are already available?

`Psl\SecureRandom\Sequence` and `Psl\PseudoRandom\Sequence\MT...`? 

Not sure where to place the interface in that case.

src/Psl/Internal/Loader.php
```diff
@@ -651,6 +654,9 @@ final class Loader
         'Psl\IO\Writer',
         'Psl\IO\Reader',
         'Psl\IO\MemoryHandle',
+        'Psl\RandomSequence\MersenneTwisterSequence',
+        'Psl\RandomSequence\MersenneTwisterPHPVariantSequence',
```
veewee, line 665: I like this naming here! Better than what is was before :)

docs/component/random-sequence.md
veewee: Was thinking : since the secure random sequence is the same as `SecureRandom\int()`, we might not need to use the interface for secure random?

azjezz: you might need it if you want predictable results during testing ( you would use MT ), but you want secure random sequence in production ( think an online casino, which would *need* to use a secure sequence for generating stuff, since MT is predictable ), so have an interface + different implementations is the way to go there.

> Not sure where to place the interface in that case.

yea, that's the issue :)

PR #196, opened 2021-05-19
fix strict type ignored when mapping or filtering arrays

### LINKED ISSUE #194: strict types is ignored when mapping or filtering arrays due to the usage of PHP builtin functions
This bug was introduced in #169 by mistake, the solution would be to wrap all callbacks provided by the end user in another closure to ensure that strict types is used.

**Environment (please complete the following information):**
 - OS: *
 - PHP version: 7.4, 8.0
 - Version: 1.6, 1.7

**Additional context**

see https://github.com/Roave/BackwardCompatibilityCheck/pull/306#discussion_r635004153

azjezz:
fixes #194


src/Psl/Dict/filter.php
```diff
@@ -37,7 +37,13 @@ function filter(iterable $iterable, ?callable $predicate = null): array
     $predicate = $predicate ?? Closure::fromCallable('Psl\Internal\boolean');
 
     if (is_array($iterable)) {
-        return array_filter($iterable, $predicate);
+        return array_filter(
+            $iterable,
+            /**
+             * @param Tv $v
+             */
+            static fn($v): bool => $predicate($v)
```
azjezz, line 45: `$predicate` is now called in user-land, so strict types apply :D

PR #200, opened 2021-05-28
[Type] add non_empty_vec

ntzm:


src/Psl/Type/Internal/NonEmptyVecType.php
```diff
@@ -0,0 +1,144 @@
...
+    ) {
+        Psl\invariant(!$value_type->isOptional(), 'Optional type must be the outermost.');
+    }
+
+    /**
+     * @psalm-assert-if-true non-empty-list<Tv> $value
+     */
+    public function matches(mixed $value): bool
+    {
+        if (!is_array($value)) {
```
azjezz, line 41: missing `use function is_array;`

src/Psl/Type/Internal/NonEmptyVecType.php
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * @throws CoercionException
+     *
+     * @return non-empty-list<Tv>
+     */
+    public function coerce(mixed $value): iterable
+    {
+        if (is_iterable($value)) {
+            $value_trace = $this->getTrace()
+                ->withFrame(Str\format('list<%s>', $this->value_type->toString()));
```
azjezz:
```suggestion
                ->withFrame(Str\format('non-empty-vec<%s>', $this->value_type->toString()));
```

( if the vec type uses list, it should be updated to vec as well ).

src/Psl/Type/Internal/NonEmptyVecType.php
```diff
@@ -0,0 +1,144 @@
...
+     *
+     * @return non-empty-list<Tv>
+     *
+     * @psalm-assert non-empty-list<Tv> $value
+     */
+    public function assert(mixed $value): array
+    {
+        if (is_array($value)) {
+            $value_trace = $this->getTrace()
+                ->withFrame(Str\format('list<%s>', $this->value_type->toString()));
```
azjezz: same here.

tests/unit/Type/NonEmptyVecTypeTest.php
```diff
@@ -0,0 +1,93 @@
...
+        yield [Type\bool()];
+        yield [null];
+        yield [false];
+        yield [true];
+        yield [STDIN];
+    }
+
+    public function getToStringExamples(): iterable
+    {
+        yield [$this->getType(), 'non-empty-list<int>'];
```
azjezz: should be non-empty-vec 🤔 while psalm calls this type 'list', we call it 'vec' here, since list cannot be an actual type in PHP due to list being a reserved keyword.

ntzm: `VecType` uses `list<>` instead of `vec<>`, does that need updating too?

azjezz: yes

azjezz (approved):
Thanks @ntzm !

PR #201, opened 2021-05-29
[Type] add non_empty_dict

ntzm:


src/Psl/Type/Internal/NonEmptyDictType.php
```diff
@@ -0,0 +1,120 @@
...
+    /**
+     * @throws CoercionException
+     *
+     * @return non-empty-array<Tk, Tv>
+     */
+    public function coerce(mixed $value): array
+    {
+        if (is_iterable($value)) {
+            $key_trace   = $this->getTrace()
+                ->withFrame(Str\format('non-empty-array<%s, _>', $this->key_type->toString()));
```
azjezz: can we use also use dict instead of array here? :) thanks.

ntzm: oops!

azjezz (approved):
Thank you again @ntzm

PR #205, opened 2021-06-03
Filesystem - add argument values into InvariantViolationException message

### LINKED ISSUE #204: Filesystem - read_file - InvariantViolationException message should contains filename

When I try to `read_file` which does not exist,  is not file or is not readable `InvariantViolationException` is thrown with message like `$file does not exist`.

I would like to see given filename in exception message. It's than easier to debug and fix possible problem.

Replace 

```php
Psl\invariant(exists($file), '$file does not exist.');
```

with

```php
Psl\invariant(exists($file), Str\format('File "%s" does not exist.', $file));
```

Thrown `RuntimeException` contains provided `$file` too.

vaclavvanik:
I mentioned in #204 that thrown `InvariantViolationException` messages should contain argument values. It's easier to debug and fix code.

I updated everything I have found. Unit tests just updated. No additional unit tests added.

### DIFF
--- a/src/Psl/Filesystem/Internal/write_file.php
+++ b/src/Psl/Filesystem/Internal/write_file.php
@@ -27,8 +27,8 @@ use const LOCK_EX;
 function write_file(string $file, string $content, bool $append): void
 {
     if (Filesystem\exists($file)) {
-        Psl\invariant(Filesystem\is_file($file), Str\format('File "%s" is not a file.', $file));
-        Psl\invariant(Filesystem\is_writable($file), Str\format('File "%s" is not writeable.', $file));
+        Psl\invariant(Filesystem\is_file($file), 'File "%s" is not a file.', $file);
+        Psl\invariant(Filesystem\is_writable($file), 'File "%s" is not writeable.', $file);
     } else {
         Filesystem\create_file($file);
     }
--- a/src/Psl/Filesystem/Internal/write_file.php
+++ b/src/Psl/Filesystem/Internal/write_file.php
@@ -27,8 +27,8 @@ use const LOCK_EX;
 function write_file(string $file, string $content, bool $append): void
 {
     if (Filesystem\exists($file)) {
-        Psl\invariant(Filesystem\is_file($file), '$file is not a file.');
-        Psl\invariant(Filesystem\is_writable($file), '$file is not writeable.');
+        Psl\invariant(Filesystem\is_file($file), Str\format('File "%s" is not a file.', $file));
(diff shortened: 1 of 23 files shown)


src/Psl/Filesystem/Internal/write_file.php
```diff
@@ -27,8 +27,8 @@
 function write_file(string $file, string $content, bool $append): void
 {
     if (Filesystem\exists($file)) {
-        Psl\invariant(Filesystem\is_file($file), '$file is not a file.');
-        Psl\invariant(Filesystem\is_writable($file), '$file is not writeable.');
+        Psl\invariant(Filesystem\is_file($file), Str\format('File "%s" is not a file.', $file));
+        Psl\invariant(Filesystem\is_writable($file), Str\format('File "%s" is not writeable.', $file));
```
azjezz: we shouldn't format these in place, instead pass it to invariant to do the formatting if the given fact is false.

```suggestion
        Psl\invariant(Filesystem\is_writable($file), 'File "%s" is not writeable.', $file);
```

azjezz (changes requested):
Thank you for working on this 👍

azjezz:
P.S: you can regenerate the docs using `php docs/documentor.php`.

src/Psl/Filesystem/Internal/write_file.php
vaclavvanik: Sure, you are right. I missed `Psl\invariant` can format message itself. I have updated the code.

vaclavvanik:
I see inconsistency in input argument `$file` vs `$filename`. Almost everywhere is `$filename`. Just in

- append_file
- Internal\write_file
- read_file
- write_file

is $file. I am not sure which is better. But I do not want to rename it in this PR.

azjezz:
yea, i can see how that is confusing, we should rename $filename to $path or $node, as everywhere else it means "file or directory", while the functions you mentioned explicitly require a file, not a directory.

vaclavvanik:
Docs updated. I think I am done.

azjezz:
make coding-standards-fix 😅

azjezz (dismissed):
Thank you @vaclavvanik!

azjezz:
ops, doc needs another update ( you can run checks locally using `make check` ( now i see that it doesn't check documentation, probably need to update that later on )

azjezz (approved):

vaclavvanik:
I have updated docs. Now should be everything ok.

Issue #212, opened 2021-06-24
Typing for `shape` makes psalm think it's a no-op


zerkms:
**Describe the bug**
I'm not sure if it's Psl typing or psalm-related bug, so trying it here first :-)

**To Reproduce**
```php
    private function augment(array $data): array
    {
        $data2 = Type\shape([
            'meta' => Type\dict(Type\array_key(), Type\mixed()),
        ], true)->assert($data);

        $data2['meta']['id'] = 42;

        return $data2;
    }

    /**
     * @param mixed $data
     */
    private function augment2($data): array
    {
        $data2 = Type\shape([
            'meta' => Type\dict(Type\array_key(), Type\mixed()),
        ], true)->assert($data);

        $data2['meta']['id'] = 42;

        return $data2;
    }
```
This code emits just one error
```
ERROR: RedundantCondition - filename.php:96:19 - Type array<array-key, mixed> for $data is always array (see https://psalm.dev/122)
        ], true)->assert($data);
```
and it refers to the `augment` function. It only fails type check if it's already known to be an array, but not when `mixed`.

**Expected behavior**
I think in both cases there should be no type check warnings.

**Environment (please complete the following information):**
 - OS: [e.g. iOS, Ubuntu] N/a
 - PHP version 7.4
 - Version 1.6.1

**Additional context**

azjezz:
are you using the psl psalm plugin ( github.com/php-standard-library/psalm-plugin )? 🤔

zerkms:
Sure, if it's not easily reproducible for you I can make up a repro repository tomorrow.

azjezz:
what psalm version are you using exactly? could you please send a PR containing a failing static analysis test? ( see: https://github.com/azjezz/psl/tree/1.8.x/tests/static-analysis/Type ), this will be easier for me to debug.

zerkms:
@azjezz just done it https://github.com/azjezz/psl/pull/214

Presumably, that problem is masked with suppressing `RedundantCondition`.

I understand the reason `RedundantCondition` is used (to have extra runtime checks), but why isn't it suppressed in place: like in cases when you know it for sure, so can tell it explicitly, instead of suppressing it everywhere and masking real issues?

That question is valid to all suppressions as well: would it be better to only suppress in-place so that it didn't accidentally mask real unexpected problems?

azjezz:
> That question is valid to all suppressions as well: would it be better to only suppress in-place so that it didn't accidentally mask real unexpected problems?

[...]

PR #214, opened 2021-06-24
Added repro for #212

zerkms:
Because I removed a global suppression there would be errors around the codebase, but I think those should addressed particularly in every specific case.

### DIFF
--- a/src/Psl/Math/base_convert.php
+++ b/src/Psl/Math/base_convert.php
@@ -40,6 +40,7 @@ use function bcpow;
  */
 function base_convert(string $value, int $from_base, int $to_base): string
 {
+    /** @psalm-suppress RedundantCondition */
     Psl\invariant('' !== $value, 'Unexpected empty string, expected number in base %d', $from_base);
     Psl\invariant(
         $from_base >= 2 && $from_base <= 36,
--- a/src/Psl/Math/from_base.php
+++ b/src/Psl/Math/from_base.php
@@ -28,6 +28,7 @@ use Psl\Str\Byte;
  */
 function from_base(string $number, int $from_base): int
 {
+    /** @psalm-suppress RedundantCondition */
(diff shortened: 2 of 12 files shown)


azjezz:
Hm, this doesn't seem to be failing 🤔

azjezz:
🙃 the static analysis configuration doesn't include `tests/static-analysis` could you please add it?

zerkms:
@azjezz it's because

```
static-analysis:
	php tools/psalm/vendor/bin/psalm -c tools/psalm/psalm.xml --show-info=true --no-cache
	php tools/psalm/vendor/bin/psalm -c tools/psalm/psalm.xml tests/static-analysis --no-cache
```

the first command fails, so the second is not even run.

If run separately it's:

```
I have no name!@d0576c8e7d27:/app$ php tools/psalm/vendor/bin/psalm -c tools/psalm/psalm.xml tests/static-analysis --no-cache
Scanning files...
Analyzing files...

░░░░░░░░░░░░░░░░

To whom it may concern: Psalm cannot detect unused classes, methods and properties
when analyzing individual files and folders. Run on the full project to enable
complete unused code detection.

ERROR: RedundantCondition - ../../tests/static-analysis/Type/shape.php:12:15 - Type array<array-key, mixed> for $data is always array (see https://psalm.dev/122)
    ], true)->assert($data);


------------------------------
1 errors found
------------------------------

Checks took 0.99 seconds and used 166.654MB of memory
Psalm was unable to infer types in the codebase
```

azjezz:
yea, i think we should make it into just one command and add the tests to psalm config

zerkms:
@azjezz see another commit

azjezz:
ref: https://github.com/vimeo/psalm/issues/6573

azjezz:
now that we know the source of this issue, i think it's safe to close it.

The best solution for you as a user right now is to suppress redundant conditions until this is fixed in Psalm.

PR #215, opened 2021-06-30
[Fun] Add lazy() for doing lazy evaluations

veewee:
This `Fun\lazy()` helper can be used for doing lazy evaluations:
https://en.wikipedia.org/wiki/Lazy_evaluation

Examples:

**Lazy proxy:**

```php
$soap = Fun\lazy(fn () => new \SoapClient('http://slowwsdl.com'));
$functions = $soap()->__getFunctions();
```

**Lazy evaluations:**

```php
$a = Fun\lazy(fn(): int => 1);
for ($i = 1; $i<10; $i++) {
    $b = $a;
    $a = Fun\lazy(fn() => $b() + $b());
}

$aResult = $a();
// = 2**10 -> all intermediate values are calculated lazily when calling the $a() callable.
```

**Lazy streams**

```php
$incrementalNumbersStream = Fun\lazy(function() {
    $i = 0;
    while (true) {
        yield ++$i;
    }
});

take(10, $incrementalNumbersStream()); // [1...10]
take(20, $incrementalNumbersStream()); // [11...30]
```


tests/unit/Fun/LazyTest.php
```diff
@@ -0,0 +1,78 @@
...
+    public function testItCanBeUsedAsALazyStream(): void
+    {
+        $incrementalNumbersStream = Fun\lazy(static function () {
+            $i = 0;
+            while (true) {
+                yield ++$i;
+            }
+        });
+
+        $take = static function (Generator $stream, int $n) {
```
veewee, line 85: FYI: `Dict\take()` and `Dict\take_while()` both try to rewind the closed stream().
This might be something that needs fixing?

src/Psl/Fun/Internal/LazyEvaluator.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Fun\Internal;
+
+/**
+ * @internal
```
veewee, line 8: Not sure this should be internal, since psalm complains about typehinting the results with `callable(): ExpectedResult`

veewee: (Fixed in the `lazy()` method by returning a closure around the evaluator)

tests/unit/Fun/LazyTest.php
```diff
@@ -0,0 +1,78 @@
...
+        $x = new class {
+            public function doSomething(): int
+            {
+                return 132;
+            }
+        };
+        $proxy = Fun\lazy(static fn (): object => $x);
+
+        static::assertSame(132, $proxy()->doSomething());
+    }
```
azjezz: can we add an assertation that it returns the same value every time?

tests/unit/Fun/LazyTest.php
```diff
@@ -0,0 +1,78 @@
...
+
+    public function testItCanBeUsedAsALazyEvaluator(): void
+    {
+        $a = Fun\lazy(static fn(): int => 1);
+        for ($i = 1; $i <= 10; $i++) {
+            $b = $a;
+            $a = Fun\lazy(static fn() => $b() + $b());
+        }
+
+        static::assertSame($a(), pow(2, 10));
```
azjezz, line 73: pow() is not imported

tests/unit/Fun/LazyTest.php
azjezz: yea, i guess we should fix these 👍 ( can you open a separate issue? )

veewee: Reported it for now to keep track of it : https://github.com/azjezz/psl/issues/216

src/Psl/Fun/Internal/LazyEvaluator.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        $this->initializer = $initializer;
+    }
+
+    /**
+     * @return T
+     */
+    public function __invoke(): mixed
+    {
+        if (!$this->instance) {
```
azjezz: what if the initializer is returning null?

i think it's better to have another property `bool $initialized`.

e.g:
```
$get_user = Fun\lazy(function(): ?User {
  $user_id = $this->session->get('id');
  return $this->users->findOne($id);
});
```

veewee: Covered that case now.
Why not `throw UserNotFoundException` or return `NullUser` ? ;)

azjezz: just an example :p don't overthink it hehehe

veewee: I was just joking :)

tests/unit/Fun/LazyTest.php
```diff
@@ -0,0 +1,101 @@
...
+        static::assertSame($x, $proxy());
+        static::assertSame($x, $proxy());
+
+        static::assertSame(132, $proxy()->doSomething());
+        static::assertSame(132, $proxy()->doSomething());
+    }
+
+    public function testItCanDealWithNullValues(): void
+    {
+        $counter = (object)(['tick' => 0]);
```
azjezz, line 50: huh, we should probably add `Psl\Ref` 😛 https://docs.hhvm.com/hsl/reference/class/HH.Lib.Ref/

veewee: It is not completely clear from that page how to use it.
Maybe rather something like react's `useRef()` ? 

https://reactjs.org/docs/hooks-reference.html#useref

```php
$pointer = useRef(0);
$pointer->current(); // 0
$pointer->current(fn ($previous) => $previous + 1); // 1
$pointer->current(); // 1
```

azjezz: it's basically:

```php
/**
 * @template T
 */
final class Ref {
  /**
   * @param T $value
   */
  public function __construct(
    public mixed $value
  ) {}
}
```

azjezz: https://github.com/hhvm/hsl/blob/v3.28.x/src/Ref.php#L25

veewee: Very similar to react's useRef.
Sounds like something that could be useful indeed!

azjezz (approved):

azjezz:
Thanks @veewee 🎉

PR #219, opened 2021-07-09
[Regex] Fix replace_with() signature

caugner:
The $matches array that is passed to `preg_replace_callback()`'s $callback 
can have string keys when the pattern uses named capture groups.

See: https://github.com/vimeo/psalm/pull/5608


src/Psl/Regex/replace_with.php
```diff
@@ -15,7 +15,7 @@
  * `$callback`.
  *
  * @param non-empty-string $pattern The pattern to search for.
- * @param (callable(array<int, string>): string) $callback The replacement callable.
+ * @param (callable(string[]): string) $callback The replacement callable.
```
azjezz, line 18: same as https://github.com/vimeo/psalm/pull/5608/files#r610848313

azjezz (changes requested):

src/Psl/Regex/replace_with.php
azjezz:
```suggestion
 * @param (callable(array<array-key, string>): string) $callback The replacement callable.
```

azjezz (changes requested):

src/Psl/Regex/replace_with.php
azjezz: we don't use phpstorm generics here :)

azjezz:
could you please add some unit tests and static analysis tests for this?

src/Psl/Regex/replace_with.php
caugner: `string[]` and `array<array-key, string>` are exactly the same: https://psalm.dev/r/212f934d6c

Do you still prefer the more verbose version?

caugner: `string[]` is actually a valid array type according to the proposed PSR-5 PHPDoc: https://github.com/php-fig/fig-standards/blob/d82fd79a13ab3cd2676ed65dafbf4196de9fd322/proposed/phpdoc.md#abnf

caugner:
> could you please add some unit tests and static analysis tests for this?

Since I didn't actually fix any code, just PHPDoc, I'm not sure what could/should be unit tested here. Once https://github.com/vimeo/psalm/pull/5608 is merged and you upgrade to the next Psalm version, your code will produce an error *without this change* though.

PS: The main reason why I created this PR is that `vimeo/psalm` tests a fork of `psl` as part of CI.

src/Psl/Regex/replace_with.php
azjezz: yes, we don't use `T[]` syntax in PSL

azjezz (approved):

PR #234, opened 2021-10-06
[Fun] tap()

veewee:
This PR provides a `Fun\tap()` function.
It can be used to perform side effects in e.g. a pipeline whilst still keeping track of the original object.

```php
Fun\pipe(
    Fun\tap(
        function (User $user) use ($eventDispatcher): void {
            $eventDispatcher->dispatch(
                new UserCreatedEvent($user->id)
            );
        }
    )    
)($user);
```
```
=> $user
```

It can be handy for debugging intermediate pipeline stages this way:

```php
$log = new Ref('');
$result = Fun\pipe(
    static fn (string $x) => md5($x),
    Fun\tap(static function ($x) use ($log): void {
        $log->value = $x;
    }),
    static fn (string $x): string => Str\truncate($x, 0, 1),
)('abc');
```

Or for making broken architectures a bit more easy to integrate:

```php
$user = tap(
    static function (User $user): void {
        $user->name = 'taylor';
        $user->save();
    }
)(User::first());
```

The implementation is based on the one from [ramda](https://ramdajs.com/docs/#tap) without the value part.
You will get back a callable that takes a value as input and returns the input value as a result of that function. You can perform any action in between input and output.

Other languages / implementations

- Ramda : https://ramdajs.com/docs/#tap
- Ruby : https://medium.com/aviabird/ruby-tap-that-method-90c8a801fd6a
- Laravel : https://medium.com/@tanmaymishu/the-anatomy-of-laravels-tap-function-ea239c9846ab (This one has a higher order tap version - which I find very confusing)

### DIFF
--- a/src/Psl/Fun/tap.php
+++ b/src/Psl/Fun/tap.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Fun;
+
+/**
+ * Runs the given function with the supplied object, then returns the object.
+ *
+ * Executes the function if a value is provided.
+ * Provides a curries version if no value was provided.
+ *
+ * Example:
+ *
+ *      $run = Fun\tap(
+ *          static function(int $i): void {
+ *              echo 'Result: '.$i.PHP_EOL;
+ *          }
+ *      );
+ *
+ *      $run(1)
+ *      Prints: Result: 1
+ *      => Int(1)
+ *
+ *      $run(2)
+ *      Prints: Result: 2
+ *      => Int(2)
+ *
+ *      $run(3)
+ *      Prints: Result: 3
+ *      => Int(3)
+ *
+ * More real life example:
+ *      Fun\pipe(
+ *          Fun\tap(
+ *              function (User $user) use ($eventDispatcher): void {
+ *                  $eventDispatcher->dispatch(
+ *                      new UserCreatedEvent($user->id)
+ *                  );
+ *              }
+ *          )
+ *       )($user);
+ *
+ *       => $user
+ *
+ * @template T
+ *
+ * @param (callable(T): void) $callback
+ *
+ * @return (callable(T): T)
+ *
+ * @pure
+ */
+function tap(callable $callback): callable
+{
+    return static function (mixed $value) use ($callback): mixed {
+        $callback($value);
+
+        return $value;
+    };
+}
--- a/tests/unit/Fun/TapTest.php
+++ b/tests/unit/Fun/TapTest.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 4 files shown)


src/Psl/Fun/tap.php
```diff
@@ -0,0 +1,70 @@
...
+ *                  $eventDispatcher->dispatch(
+ *                      new UserCreatedEvent($user->id)
+ *                  );
+ *              }
+ *          )
+ *       )($user);
+ *
+ *       => $user
+ *
+ * @template V
```
azjezz:
```suggestion
 * @template T
```

src/Psl/Fun/tap.php
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * @param (callable(V): void) $callback
+ * @param (V|null) $value
```
azjezz:
```suggestion
```

we should remove this, i'm not a fun of conditional returns, and we should avoid them as much as possible.

tap($c)($v) is just as simple as tap($c, $v);

veewee: true, was in doubt about that one myself - but went with the way ramda worked.
it always returns a callable now.

src/Psl/Fun/tap.php
```diff
@@ -0,0 +1,61 @@
...
+ *                  );
+ *              }
+ *          )
+ *       )($user);
+ *
+ *       => $user
+ *
+ * @template T
+ *
+ * @param (callable(T): void) $callback
```
veewee, line 48: It currently expects `:void` as return type. 
In other implementations, this return type can be `mixed` - but the returned value gets ignored.

Both are possible, which do you prefer?

tests/unit/Fun/TapTest.php
```diff
@@ -0,0 +1,47 @@
...
+        $log = new Ref('');
+        $result = Fun\pipe(
+            static fn (string $x) => md5($x),
+            Fun\tap(static function ($x) use ($log): void {
+                $log->value = $x;
+            }),
+            static fn (string $x): string => Str\truncate($x, 0, 1),
+        )('abc');
+
+        $md5 = md5('abc');
```
azjezz:
```suggestion
        $md5 = Psl\Hash\hash('abc', 'md5');
```

azjezz (dismissed):
just nitpicking 😛

src/Psl/Fun/tap.php
azjezz: I think void works best.

azjezz: using a function that returns a value as an argument here is not an issue ( see: https://psalm.dev/r/850a78d176 )

azjezz (commented):
needs rebase

azjezz (approved):

PR #238, opened 2021-10-17
Add a new Psl\Str\reverse() function

yivi:
Psl\Str\Byte\reverse() exists, providing the functionality of strrev(), but there is no multibyte equivalent on the Psl\Str namespace.

This still breaks on some complex characters, as some multi byte emojis, but for that it would probably better to create a Psl\Str\Grapheme\reverse() for completion sake.


src/Psl/Str/reverse.php
```diff
@@ -0,0 +1,16 @@
...
+namespace Psl\Str;
+
+use Psl\Vec;
+
+/**
+ * Returns the given string reversed.
+ */
+function reverse(string $string, ?string $encoding = null): string
+{
+        $parts = split($string, '', null, $encoding);
```
azjezz:
```suggestion
        $chunks = chunk($string, encoding: $encoding);
```

src/Psl/Str/reverse.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Str;
+
+use Psl\Vec;
+
+/**
+ * Returns the given string reversed.
```
azjezz, line 10:
```suggestion
 * Returns the given string reversed.
 *
 * @psalm-pure
```

src/Psl/Str/reverse.php
```diff
@@ -0,0 +1,16 @@
...
+ */
+function reverse(string $string, ?string $encoding = null): string
+{
+        $parts = split($string, '', null, $encoding);
+        return join(Vec\reverse($parts), '');
```
azjezz: psalm will probably complain about `reverse` not being pure, but you can suppress it here using `@psalm-suppress` since `reverse` is conditionally pure :)

tests/unit/Str/ReverseTest.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Tests\Unit\Str;
+
+use PHPUnit\Framework\TestCase;
+use Psl\Str;
+
+class ReverseTest extends TestCase
```
azjezz:
```suggestion
final class ReverseTest extends TestCase
```

tests/unit/Str/ReverseTest.php
```diff
@@ -0,0 +1,31 @@
...
+
+    public function provideData(): array
+    {
+        return [
+            ['Hello World', 'dlroW olleH'],
+            ['héllö wôrld', 'dlrôw ölléh'],
+            ['Iñigo Montoya', 'ayotnoM ogiñI'],
+            ['某物', '物某' ],
+            ['что-то', 'от-отч']
+        ];
```
azjezz: we need more test cases here:

```suggestion
        return [
            ['Hello World', 'dlroW olleH'],
            ['héllö wôrld', 'dlrôw ölléh'],
            ['Iñigo Montoya', 'ayotnoM ogiñI'],
            ['某物', '物某' ],
            ['что-то', 'от-отч']
            ['🙂😟', '😟🙂'],
            ['مرحبا', 'ابحرم'],
        ];
```

yivi: added on separate commit, missing a `,` on the suggestion.

yivi:
Mmmh. I think this would be better accomplished with a regex_split. With this it would be possible to capture multibyte extended unicode sequences:

```php
use Psl\Vec;

$rev_array = Vec\reverse(\preg_split('/(\X)/u', $string, -1, \PREG_SPLIT_DELIM_CAPTURE | \PREG_SPLIT_NO_EMPTY));
```
I've been testing and this would work with much more complex unicode sequences.

The "problem" is that as far as I can see, we cannot pass additional flags to `Regex\split()`

For a use-case like this, would using the native `preg_split()` over the Psl version be fine?

Also, going through `preg_split` instead of `mb_*` functions would make `$encoding` not needed. Which I don't know if it's a good or a bad thing.

azjezz:
I think that's not needed, I'm happy with the current version, and i don't think we need to overdo it here.

azjezz (approved):

yivi:
Good. Thanks for all your help.

Further checking shows that the regex version also seems to choke with extended unicode chars (trying with emojis like this '👩‍👩‍👦👩🏽‍❤️‍👨🏼👩🏽‍🏫', given my ignorance of more interesting languages).

Would it be ok if I also created a PR for `Str\Grapheme\reverse()`, for completeness sake?

azjezz:
> Would it be ok if I also created a PR for Str\Grapheme\reverse(), for completeness sake?

Yes, i think that would be a nice addition as well :)

PR #239, opened 2021-10-18
Str\Grapheme\reverse() function

yivi:
Add Str\\Grapheme\reverse() to match Str\\reverse() and Str\Byte\reverse()

### DIFF
--- a/src/Psl/Str/Grapheme/reverse.php
+++ b/src/Psl/Str/Grapheme/reverse.php
@@ -8,6 +8,9 @@ namespace Psl\Str\Grapheme;
  * Reverses the string.
  *
  * @pure
+ *
+ * @throws \Psl\Exception\InvariantViolationException If unable to convert $string to UTF-16,
+ *                                                    or split it into graphemes
  */
 function reverse(string $string): string
 {
--- a/src/Psl/Str/Grapheme/reverse.php
+++ b/src/Psl/Str/Grapheme/reverse.php
@@ -4,30 +4,18 @@ declare(strict_types=1);
 
 namespace Psl\Str\Grapheme;
 
-use Psl\Exception\InvalidArgumentException;
-
-use function grapheme_strlen;
-use function grapheme_substr;
-
 /**
  * Reverses the string.
  *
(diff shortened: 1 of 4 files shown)


yivi:
Regarding the coverage decrease... I'm afraid I do not yet know how to produce an input that `grapheme_strlen` would choke on.

Docs (https://www.php.net/manual/en/function.grapheme-strlen.php) say the function returns `int|false|null` with `false` and on failure (?) and, with no indication of when `null` would be returned.

I'm trying to see how to create a string that would be an invalid UTF-8 string to see if we can recover the lost coverage.

yivi:
Ok, got it. Pushing recovered coverage. :)

tests/unit/Str/Grapheme/ReverseTest.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Tests\Unit\Str\Grapheme;
+
+use PHPUnit\Framework\TestCase;
+use Psl;
```
azjezz: we usually import the component, so `use Psl\Str\Byte;` + `use Psl\Str\Grapheme;` for consistency.

src/Psl/Str/Grapheme/reverse.php
```diff
@@ -0,0 +1,34 @@
...
+
+    $length = grapheme_strlen($string);
+
+    if ($length === false || $length === null) {
+        throw new InvalidArgumentException('Cannot get string length. Possibly invalid UTF-8 sequence');
+    }
+
+    while ($length-- > 0) {
+        $reversed .= grapheme_substr($string, $length, 1);
+    }
```
azjezz: handling these errors shouldn't be done here, there's already wrappers for these grapheme_* functions.

```suggestion
    $length = length($string);

    while ($length-- > 0) {
        $reversed .= slice($string, $length, 1);
    }
```

azjezz (approved):

azjezz:
Thank you @yivi 🎉

PR #262, opened 2021-11-09
Call clearstatcache after writing to a file to reset file stat

dragosprotung:
`write_file` internally calls `is_file` before writing to the file. This means the file stat is cached before the write operation.

If we do not clear the cache, many functions like `filesize` will report invalid data.


azjezz (approved):
Thank you 👍

PR #267, opened 2021-11-10
chore(io): do not throw on non-blocking resource

azjezz:


azjezz:
helps with windows support.

PR #272, opened 2021-11-13
[DataStructures] Set

veewee:
This PR can be used to discuss the syntax of a new Set datastructure.

https://en.wikipedia.org/wiki/Set_(abstract_data_type)
https://www.php.net/manual/en/class.ds-set.php

Additionally, we can introduce `Vec\unique` and `Vec\unique_by` functions based on this data structure.

### DIFF
--- a/docs/component/collection.md
+++ b/docs/component/collection.md
@@ -20,7 +20,9 @@
 - [MutableCollectionInterface](./../../src/Psl/Collection/MutableCollectionInterface.php#L20)
 - [MutableIndexAccessInterface](./../../src/Psl/Collection/MutableIndexAccessInterface.php#L16)
 - [MutableMapInterface](./../../src/Psl/Collection/MutableMapInterface.php#L14)
+- [MutableSetInterface](./../../src/Psl/Collection/MutableSetInterface.php#L12)
 - [MutableVectorInterface](./../../src/Psl/Collection/MutableVectorInterface.php#L13)
+- [SetInterface](./../../src/Psl/Collection/SetInterface.php#L12)
 - [VectorInterface](./../../src/Psl/Collection/VectorInterface.php#L12)
 
 #### `Classes`
--- a/src/Psl/Collection/MutableSetInterface.php
+++ b/src/Psl/Collection/MutableSetInterface.php
@@ -1,0 +1,370 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Collection;
+
+/**
+ * @template T
+ *
+ * @extends MutableVectorInterface<T>
+ */
+interface MutableSetInterface extends MutableVectorInterface
+{
+    /**
+     * Get an array copy of the current vector.
+     *
+     * @return list<T>
+     *
+     * @psalm-mutation-free
+     */
+    public function toArray(): array;
+
+    /**
+     * Returns a `MutableSetInterface` containing the values of the current
+     * `MutableSetInterface`.
+     *
+     * @return MutableSetInterface<T>
+     *
+     * @psalm-mutation-free
+     */
+    public function values(): MutableSetInterface;
+
+    /**
+     * Returns a `MutableSetInterface` containing the keys of the current `MutableSetInterface`.
+     *
+     * @return MutableSetInterface<int>
+     *
+     * @psalm-mutation-free
+     */
+    public function keys(): MutableSetInterface;
+
+    /**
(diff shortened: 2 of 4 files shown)


azjezz:
Do you think `Psl\Collection` is a better home for this? as we can implement same interfaces implemented by vector 🤔 

Also, i think this should go to 2.0.x, there isn't going to be 1.10, so we should only merge bug fixes in 1.9.

veewee:
There are quite some methods on the vector interface that don't really apply to sets. It is possible to implement them, but maybe just not as intuitive to use in that context. I can try to create another interface based on the methods inside vector?

The difference between collections and data structures is not that clear to me, since e.g. php-ds has all of them. So I leave the decision of the namespace up to you in this one.

veewee:
The more I think about it, the less it makes sense to make it apply to the vector interface.
It is a separate data structure with its own use-case.
Therefore, it should have its own interface.

Since it does not behave like a collection, I think it's fine to put it in the data structures namespace.
Other data structures also have a more minimal API than the collections.

In previous commit, I went with the absolute basics for a set and made it more consistent with the existing interfaces for collections and data structures. It is more in line with the [Set from JS](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set). You can implement the interface in both a mutable and immutable way.


I don't think it makes sense to put in more functionality. Instead, I made a shortcut to transform it to a vector. This vector already has the more advanced API available.

Feel free to improve or disagree with me :)

azjezz:
> make it apply to the vector interface.

I probably should have been clearer, i meant implement the same interfaces vector implements ( and other interface in collection namespace ), not to implement the vector interface itself.

i.e IndexAccessInterface, and CollectionInterface.

It would also make sense to have both `SetInterface`, and `MutableSetInterface` to encourage the usage of immutable objects.

veewee:
I understood, but those functions inside these collection interfaces don't really cover the set use-case. 

For example: at(),  get(), *Key(), everything position/index based. Those don't make sense for a set.

Here are some example implementations:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#implementing_basic_set_operations

As you can see, it's less about iterating and more about keeping unique internal lists

azjezz:
a set IMO is just a vector of unique values, they still have a position that is tracked internally, the same way as a vector, which would be used in `firstKey()`, `lastKey()`, `at()` and `get()`.

see: 
- https://docs.hhvm.com/hack/reference/interface/MutableSet/
- https://docs.hhvm.com/hack/reference/class/HH.Set/
- https://docs.hhvm.com/hack/reference/class/HH.ImmSet/

even if JS `Set` doesn't provide key-based methods, people find a way around it ( https://stackoverflow.com/questions/44715210/how-to-get-index-based-value-from-set )

---

since we are discussing Keyed and non-Keyed collections, it might be a good idea to split the two concepts for PSL 2.0, where KeyedCollectionInterface will have all functions that relay on the collection being key-based ( map ), but that should be a different PR/issue

veewee:
Alright, I give up and went for the collection approach :')

The HSL docs state
> Sets only support int values and string values. If a value of a different type is used, an exception will be thrown.

This is most likely because they use a map internally to store unique values as key. However, it can be usefull to add classes like enums to sets as well. Not sure how you think about this? I'dd rather have support for objects as well.

azjezz:
> This is most likely because they use a map internally to store unique values as key. However, it can be useful to add classes like enums to sets as well. Not sure how you think about this? I'd rather have support for objects as well.

filtering an array by unique values of `string|int` is way much faster than doing it on all types.

when it's just strings and integers, we would use Dict\unique_scalar, otherwise we would need Dict\unique ( see #167 #168 ), so if you think the performance penalty is okay, we can support other types.

veewee:
We could use `array_unique` internally instead of the custom `unique_by` function. So it might not be that much of an issue.
Part of the performance issue is the unnecessary additional iteration on the uniques to transform it into a result in the `unique_by` function.

I might change the logic based on sets as well afterwards. Maybe best to add benchmarking first for those functions.
But will have to figure it all out :)

azjezz:
> We could use array_unique internally

array_unique works only with scalar values.

veewee:
Gonna close this PR: I currently don't need this structure anymore and lack some time to finish it up.
Feel free to pick it up if you would like to use it!

PR #286, opened 2021-11-24
chore: enable mutation tests

azjezz:
This is currently on hold as we need to wait for a new release of infection that fixes PHP 8.1 compatibility, as well as a new release of roave-infection-static-analysis-plugin that uses latest infection release ~+ contains roave/infection-static-analysis-plugin#248~.

~This PR is also partially-blocked by roave/infection-static-analysis-plugin#27 as currently we cannot place infection configuration in `config/` directory with the rest of the configuration files ( we can make an exception for infection, so no pressure @Ocramius ^^ )~

---
current results:

```shell
1438 mutations were generated:
    1021 mutants were killed
     394 mutants were not covered by tests
       0 covered mutants were not detected
       4 errors were encountered
       0 syntax errors were encountered
      19 time outs were encountered
       0 mutants required more time than configured

Metrics:
         Mutation Score Indicator (MSI): 72%
         Mutation Code Coverage: 72%
         Covered Code MSI: 100%
```

i suppose the 394 mutation not covered are the ones that we purposely ignore when doing error handling ( since it's not easy to duplicate every failing case for PHP builtin functions )

The time outs are result of mutating the async component ( especially Awaitable State and Deferred ), and other places where Deferred and EventLoop Suspensions are used.

### DIFF
--- a/.github/workflows/mutation-tests.yml
+++ b/.github/workflows/mutation-tests.yml
@@ -1,0 +1,48 @@
+name: "mutation tests"
+
+on:
+  pull_request: ~
+  push: ~
+
+jobs:
+  mutation-tests:
+    name: "mutation tests"
+
+    runs-on: ${{ matrix.operating-system }}
+
+    strategy:
+      matrix:
+        php-version:
+          - "8.1"
+        operating-system:
+          - "ubuntu-latest"
+
+    steps:
+      - name: "checkout"
+        uses: "actions/checkout@v2"
+
+      - name: "installing PHP"
+        uses: "shivammathur/setup-php@v2"
+        with:
+          php-version: "${{ matrix.php-version }}"
+          ini-values: memory_limit=-1
+          tools: composer:v2, cs2pr
+          extensions: bcmath, mbstring, intl, sodium, json
+
+      - name: "caching dependencies"
+        uses: "actions/cache@v2.1.7"
+        with:
+          path: |
+            ~/.composer/cache
+            vendor
+          key: "php-${{ matrix.php-version }}"
+          restore-keys: "php-${{ matrix.php-version }}"
+
+      - name: "installing dependencies"
+        run: |
+          make install -j10
+
+      - name: "running mutation tests"
+        run: make mutation-tests
+        env:
+          STRYKER_DASHBOARD_API_KEY: ${{ secrets.STRYKER_DASHBOARD_API_KEY }}
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "phpunit/phpunit": "^9.5.10",
         "vimeo/psalm": "^4.11.2",
         "php-standard-library/psalm-plugin": "^1.1.1",
-        "php-coveralls/php-coveralls": "^2.5"
+        "php-coveralls/php-coveralls": "^2.5",
+        "roave/infection-static-analysis-plugin": "1.13.x-dev"
     },
     "autoload": {
         "psr-4": {
--- a/config/infection.json.dist
+++ b/config/infection.json.dist
@@ -1,0 +1,26 @@
+{
+    "source": {
+        "directories": [
+            "src/Psl/"
+        ],
+        "excludes": [
+            "RandomSequence",
+            "Async",
+            "IO",
+            "Unix",
+            "TCP",
+            "Internal",
+            "{.*/Internal/.*}"
+        ]
+    },
+    "logs": {
+        "text": "php://stderr",
+        "github": true,
+        "badge": {
+            "branch": "/^\\d+\\.\\d+\\.x$/"
+        }
+    },
+    "timeout": 10,
+    "minMsi": 98,
+    "minCoveredMsi": 100
+}
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,3 @@
 
 # phpunit cache
 /config/.phpunit.result.cache
-
-# test logs
-/tests/logs/*
--- a/Makefile
+++ b/Makefile
@@ -34,8 +34,11 @@ security-analysis:                                                              
 unit-tests:                                                                     ## run unit test suite
 	./vendor/bin/phpunit -c config/phpunit.xml.dist
 
+mutation-tests:                                                                     ## run mutation tests
+	./vendor/bin/roave-infection-static-analysis-plugin run --configuration=config/infection.json.dist --psalm-config=config/psalm.xml
+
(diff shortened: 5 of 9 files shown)


infection.json.dist
```diff
@@ -0,0 +1,30 @@
+{
+    "source": {
+        "directories": [
+            "src/Psl/"
+        ],
+        "excludes": [
+            "RandomSequence"
```
azjezz: too much randomness to actually test.

psalm.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0"?>
```
azjezz: This file is to be removed one https://github.com/Roave/infection-static-analysis-plugin/pull/248 is merged.

infection.json.dist
```diff
@@ -0,0 +1,30 @@
...
+    },
+    "timeout": 10,
+    "logs": {
+        "text": "var/infection.log",
+        "summary": "var/summary.log",
+        "json": "var/infection-log.json",
+        "perMutator": "var/per-mutator.md",
+        "github": true,
+        "badge": {
+            "branch": "2.0.x"
```
Ocramius: You could use a regex here: `/^\d+\.\d+\.x$/` perhaps? 

See https://github.com/infection/site/pull/220/files

psalm.xml
```diff
@@ -0,0 +1,53 @@
...
+            <file name="src/bootstrap.php" />
+            <file name="src/preload.php" />
+            <file name="src/Psl/Internal/Loader.php" />
+        </ignoreFiles>
+    </projectFiles>
+
+    <issueHandlers>
+        <DuplicateFunction errorLevel="suppress" />
+
+        <LessSpecificReturnType errorLevel="error" />
```
Ocramius: Note: everything with `="error"` can be dropped, as psalm defaults to error

psalm.xml
Ocramius: Isn't that what `-c` is supposed to do within psalm?

.github/workflows/mutation-tests.yml
```diff
@@ -0,0 +1,46 @@
...
+        with:
+          path: |
+            ~/.composer/cache
+            vendor
+          key: "php-${{ matrix.php-version }}"
+          restore-keys: "php-${{ matrix.php-version }}"
+
+      - name: "installing dependencies"
+        run: |
+          make install -j10
```
Ocramius, line 43: Suggestion: add `-O` to have readable output in the end, in case of failures :D

Ocramius (changes requested):

psalm.xml
azjezz: flags are completely ignored when using roave infection-static-analysis-plugin ( https://github.com/Roave/infection-static-analysis-plugin/issues/27 )

Ocramius: Yes, see https://github.com/Roave/infection-static-analysis-plugin/pull/248#issuecomment-977702397 on the patch you proposed

psalm.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0"?>
+<psalm totallyTyped="true" forbidEcho="true" strictBinaryOperands="true" phpVersion="8.1" allowStringToStandInForClass="true" rememberPropertyAssignmentsAfterCall="false" checkForThrowsDocblock="true" checkForThrowsInGlobalScope="true" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://getpsalm.org/schema/config" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd" errorBaseline="config/psalm-baseline.xml">
```
bendavies: `totallyTyped=true` is deprecated and is the same as `errorLevel=1`

config/infection.json.dist
```diff
@@ -0,0 +1,30 @@
...
+        "excludes": [
+            "Async",
+            "RandomSequence",
+            "Internal",
+            "{.*/Internal/.*}"
+        ]
+    },
+    "timeout": 10,
+    "logs": {
+        "text": "var/infection.log",
```
Ocramius: One that I like using in CI is `php://stderr`

config/infection.json.dist
```diff
@@ -0,0 +1,30 @@
...
+        "json": "var/infection-log.json",
+        "perMutator": "var/per-mutator.md",
+        "github": true,
+        "badge": {
+            "branch": "2.0.x"
+        }
+    },
+    "mutators": {
+        "@default": true,
+        "@function_signature": true
```
Ocramius: Probably not necessary, when used in combination with the static analysis plugin

Ocramius (dismissed):

config/infection.json.dist
```diff
@@ -0,0 +1,26 @@
...
+            "src/Psl/"
+        ],
+        "excludes": [
+            "RandomSequence",
+            "Async",
+            "IO",
+            "Unix",
+            "TCP",
+            "Internal",
+            "{.*/Internal/.*}"
```
azjezz, line 13: most async IO stuff, result either in timeouts, or false-positives, e.g:

change `timeout` in `stream_socket_accept($resource, timeout: 0.0)` from `0.0` to `1.0`, but this is never called unless the server socket is ready, so if it happened to go above the timeout ( 0.0 ), that's a bug in PHP, not PSL ( which we handle by throwing a runtime exception, but exclude it from code coverage, as it can't be tested ), and by increasing the timeout, nothing happens, as it returns the socket immediately

config/infection.json.dist
```diff
@@ -0,0 +1,26 @@
+{
+    "source": {
+        "directories": [
+            "src/Psl/"
+        ],
+        "excludes": [
+            "RandomSequence",
```
azjezz, line 7: too random to actually test fully.

config/infection.json.dist
azjezz: maybe if we could enable some mutators for these files only, it would make sense to include them, however, we want all default mutators to be enabled for the rest of the project.

composer.json
```diff
@@ -25,7 +25,8 @@
         "phpunit/phpunit": "^9.5.10",
         "vimeo/psalm": "^4.11.2",
         "php-standard-library/psalm-plugin": "^1.1.1",
-        "php-coveralls/php-coveralls": "^2.5"
+        "php-coveralls/php-coveralls": "^2.5",
+        "roave/infection-static-analysis-plugin": "1.13.x-dev"
```
azjezz, line 29: I'm gonna assume this won't break and merge it as is.

PR #291, opened 2021-12-01
feat(async): introduce more async helper functions

azjezz:


azjezz:
@veewee stuff you like ;)

veewee:
Nice stuff! 
Looks like amp is taking a slightly different approach in v3. But at the moment, I prefer the helpers in here over launching futures. It might be habit though 🙂

azjezz:
well, the good thing is that they can be used together!

```php
use Amp;
use Amp\Future;
use Psl\Async;
use Psl\Shell;
use Psl\Result;

/**
 * @var Result\ResultInterface<string> $result1
 * @var Result\ResultInterface<string> $result2
 */
[$result1, $result2] = Amp\Future\all(
  Amp\launch(Async\reflect(fn() => Shell\execute('foo'))),
  Amp\launch(Async\reflect(fn() => Shell\execute('foo'))),
);
```

PR #317, opened 2021-12-19
chore: remove type verification at runtime

azjezz:
previously, psl would validate types of arguments at runtime, this is no longer true.

If you want to ensure you are not using PSL wrong, use a static analysis tool, as PSL will no longer tell you that what you are doing is wrong.


src/Psl/IO/WriteHandleInterface.php
```diff
@@ -32,7 +32,7 @@ public function tryWrite(string $bytes): int;
      * @throws Exception\RuntimeException If an error occurred during the operation.
      * @throws Exception\TimeoutException If reached timeout before completing the operation.
      *
-     * @return positive-int|0 the number of bytes written, which may be less than the length of input string.
+     * @return int<0, max> the number of bytes written, which may be less than the length of input string.
```
Ocramius, line 35: What is max? O.o

Ocramius (dismissed):

src/Psl/IO/WriteHandleInterface.php
azjezz: PHP_INT_MAX

azjezz: int<0, max> is same as `0|positive-int`

drupol:
I fully support this initiative :)

src/Psl/Math/base_convert.php
```diff
@@ -21,20 +21,15 @@
...
  */
 function base_convert(string $value, int $from_base, int $to_base): string
 {
-    Psl\invariant('' !== $value, 'Unexpected empty string, expected number in base %d', $from_base);
-    Psl\invariant(
-        $from_base >= 2 && $from_base <= 36,
-        'Expected $from_base to be between 2 and 36, got %d',
-        $from_base
-    );
     Psl\invariant($to_base >= 2 && $to_base <= 36, 'Expected $to_base to be between 2 and 36, got %d', $to_base);
```
azjezz: i left this on purpose to see if psalm would complain about it, but apparently it can't detect that this is redundant.

drupol:
That was quick :)

PR #318, opened 2021-12-19
chore(iter): optimize iterator

azjezz:


src/Psl/Iter/Iterator.php
```diff
@@ -94,34 +94,43 @@ public static function create(iterable $iterable): Iterator
...
-     * Move forward to the next element.
+     * Checks if current position is valid.
      */
-    public function next(): void
+    public function valid(): bool
     {
-        $this->position++;
-        if (null === $this->generator || !$this->generator->valid()) {
-            return;
+        if (isset($this->entries[$this->position])) {
```
drupol: I think you can safely use `array_key_exist`, it's a bit faster.

src/Psl/Iter/Iterator.php
```diff
@@ -199,43 +187,39 @@ public function seek(int $position): void
     }
```
drupol, line 189: Same here

drupol (commented):
I just added a minor note.

PR #323, opened 2021-12-21
chore: remove all references to callable, replace by closure

azjezz:


veewee:
Any specific reason for accepting a less broad set of callbacks?

bendavies:
@azjezz any reason for this change? making upgrading quite a chore :)

bendavies:
i guess it's because of  first-class callables in 8.1

```php
map($api->getAll($params), [PostingPeriod::class, 'fromArray']);
```
becomes

```php
map($api->getAll($params), PostingPeriod::fromArray(...));
```

bendavies:
although we did have some callable classes (with `__invoke`) which we now can't use so nicely.
that needs:

before:
```php
map($value, $this->foo);
```
after:

```php
map($value, Closure::fromCallable($this->foo));
```
or

```php
map($value, $this->foo->__invoke(...));
```

because this doesn't work:
```php
map($value, $this->foo(...));
```

not very nice

bendavies:
i'm an idiot.
this works:
```php
map($value, ($this->foo)(...));
```

azjezz:
The reasoning behind this change is in revolt-php, which uses Closure for all callbacks within the event loop, we can use Closure in async related functions, but that would be an inconsistency that i would rather avoid.

Issue #198, opened 2021-05-27
rework exceptions


azjezz:
we should refactor all of our exceptions, and use error instead where it makes sense.

  - [ ] move exceptions from `Exception` namespace to `Throwable` namespace in all components.
  - [ ] extend Error instead of  Exception in `InvariantViolationException` ( and rename to `InvariantViolation` ) - #308

PR #325, opened 2021-12-23
chore(data-structure): rework exceptions

azjezz:
1. throw underflow exception instead of invariant violation ( ref #198 )
2. optimize data structures by using builtin functions instead of PSL ( while messier, it's faster, and is not exposed in the public API )

### DIFF
--- a/src/Psl/DataStructure/PriorityQueue.php
+++ b/src/Psl/DataStructure/PriorityQueue.php
@@ -4,12 +4,10 @@ declare(strict_types=1);
 
 namespace Psl\DataStructure;
 
-use Psl;
-use Psl\Dict;
-use Psl\Iter;
 use Psl\Math;
-use Psl\Vec;
 
+use function array_keys;
+use function array_shift;
 use function count;
 
 /**
@@ -20,7 +18,7 @@ use function count;
 final class PriorityQueue implements PriorityQueueInterface
 {
     /**
-     * @var array<int, list<T>>
+     * @var array<int, non-empty-list<T>>
      */
     private array $queue = [];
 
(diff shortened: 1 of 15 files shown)


src/Psl/DataStructure/PriorityQueue.php
```diff
@@ -69,63 +67,50 @@ public function peek(): mixed
...
-         *
-         * @var int $priority
+         * retrieve the list of nodes with the priority `$priority`.
          */
-        $priority = Math\max(Vec\keys($this->queue));
-
+        $nodes = $this->queue[$priority];
         /**
-         * Retrieve the list of nodes with the priority `$priority`.
+         * shit the first node out.
```
jrmajor: Did you mean shift? :)

azjezz: LMAO

azjezz:
@veewee okay with this change? ( trying to make exceptions make sense a bit, invariant violation everywhere seemed a bit too generic, and doesn't allow for multiple catch blocks depending on the issue )

veewee:
Sure, makes sense!
At first sight it's ok, but gonna go through it tomorrow on my computer with a fresh mind.

veewee:
Looks good!
Any idea about the speed improvements by using array functions instead of psl internally?

azjezz:
should be 2 to 3 times as fast ( no just because of that change, i did some other changes as well ).

also, `Psl\Iter\Iterator` in 2.0.x is 12 times faster than the one in 1.9.x ;)

veewee:
Nice!

azjezz:
Oh, Collection is also 3 times faster ;)

PR #328, opened 2021-12-29
chore(filesystem): rework exceptions

azjezz:
ref: #198


src/Psl/File/temporary.php
```diff
@@ -15,9 +15,13 @@
  *                                         the system default temporary directory.
  * @param non-empty-string|null $prefix The prefix of the generated temporary filename.
  *
+ * @throws Filesystem\Exception\RuntimeException If unable to create the file.
+ * @throws Filesystem\Exception\RuntimeException If unable to create the file.
+ * @throws Filesystem\Exception\NotFoundException If $directory is not found.
+ * @throws Filesystem\Exception\NotDirectoryException If $directory is not a directory.
+ * @throws Filesystem\Exception\InvalidArgumentException If $prefix contains a directory separator.
  * @throws Psl\Exception\InvariantViolationException If $directory doesn't exist or is not writable.
  * @throws Psl\Exception\InvariantViolationException If $prefix contains a directory separator.
```
jrmajor, line 24: I think these cases are already covered by `NotFoundException` and `InvalidArgumentException`, except for “is not writable” part.

src/Psl/File/temporary.php
```diff
@@ -15,9 +15,13 @@
  *                                         the system default temporary directory.
  * @param non-empty-string|null $prefix The prefix of the generated temporary filename.
  *
+ * @throws Filesystem\Exception\RuntimeException If unable to create the file.
+ * @throws Filesystem\Exception\RuntimeException If unable to create the file.
```
jrmajor, line 19: Duplicate line.

src/Psl/Filesystem/delete_file.php
```diff
@@ -13,22 +12,28 @@
...
  * Delete the file specified by $filename.
  *
- * @param non-empty-string $filename
+ * @param non-empty-string $file
  *
  * @throws Exception\RuntimeException If unable to delete the file.
- * @throws Psl\Exception\InvariantViolationException If the file specified by
- *                                                   $filename does not exist.
+ * @throws Exception\NotFileException If $file is not a file.
+ * @throws Exception\NotFoundException If $file is not a found.
```
jrmajor, line 19:
```suggestion
 * @throws Exception\NotFoundException If $file is not found.
```

src/Psl/Filesystem/get_basename.php
```diff
@@ -14,17 +14,19 @@
  *
  * In other environments, it is the forward slash `/`.
  *
- * @param string|null $suffix If the filename ends in a suffix, this will also be cut off.
+ * @param non-empty-string|null $suffix If the filename ends in a suffix, this will also be cut off.
```
jrmajor, line 17: Why suffix can not be an empty string?

src/Psl/File/temporary.php
azjezz: working on a PR to rework exceptions for File component as well, this is just done to satisfy psalm :)

src/Psl/Filesystem/get_basename.php
azjezz: provide null in that case ^^ otherwise, it's useless ( this is done to encourage better typing down stream for users, so they know when they have a non-empty-string/empty-string. )

jrmajor: Wouldn't `string $suffix = ''` be more convinient for users? `$maybeNull ?? ''` is easier to do than `$maybeEmpty === '' ? null : $maybeEmpty`.

PR #334, opened 2022-02-03
feat(shell): introduce error output behavior feature

azjezz:
credits for this idea goes to @fredemmott ( https://github.com/facebook/hhvm/discussions/8946#discussioncomment-2105883 )


azjezz:
@Ocramius support for https://github.com/laminas/automatic-releases/blob/f4a0b35bf35db0b8a66adcf9b557d4709e4b533f/src/Gpg/ImportGpgKeyFromStringViaTemporaryFile.php#L20-L21 without disabling escaping ;)

Ocramius:
@azjezz not sure I follow 🤔

src/Psl/Shell/ErrorOutputBehavior.php
```diff
@@ -0,0 +1,27 @@
...
+    /**
+     * Prepend the standard error output content to the standard output content.
+     */
+    case PREPEND;
+
+    /**
+     * Replace the standard output content with the standard error output content.
+     */
+    case REPLACE;
+}
```
azjezz: feature behaviors to include:

- merge: merge both stdout and stderr, the order is not guaranteed, as both streams will be read concurrently, and which ever returns first will be first, this doesn't mean that one will come after the over, as stdout can write a chunk, then stderr, then stdout, so it will be mixed ( exactly what you see in your terminal when stderr is not redirected )
- redirect: redirect the stderr output of the process, to the stderr output of the current process.

azjezz:
in laminas automatic release we had to disable escaping to redirect stderr to stdout, since gpg writes the key to stderr, now that is not needed since you can do:

```php
$stderr = Shell\execute('gpg', ['--import', $keyFileName], error_output_behavior: ErrorOutputBehavior::Replace);
```

instead of:
```php
$stderr = Shell\execute('gpg', ['--import', Shell\escape_argument($keyFileName), '2>&1'], escape_arguments: false);
```

src/Psl/Shell/ErrorOutputBehavior.php
azjezz: #335 introduces a streaming function which helps implement the merge, and redirect behaviors.

PR #336, opened 2022-02-04
[Result] Collect stats from result sets

veewee:
A helper function for collecting statistics on a set of results.


azjezz (dismissed):

src/Psl/Result/with_stats.php
```diff
@@ -0,0 +1,29 @@
...
+use Psl\Vec;
+
+/**
+ * @template R
+ *
+ * @param callable(): iterable<array-key, ResultInterface<R>> $action
+ *
+ * @return array{0: list<ResultInterface<R>>, 1: Stats}
+ */
+function with_stats(callable $action): array
```
veewee: alternatively we could just pass an iterable of results here instead of the callable. It makes it a bit easier to combine with e.g. the `pipe()`function.

Maybe it would be better to do `stats(iterable<array-key, ResultInterface>): Stats`

WDYT?

azjezz: as you see fit, I'm okay with either.

veewee: Went with the iterable solution, since it is easier to compose.
You can even compose it into something that works like the first version if you like.

azjezz (approved):

PR #337, opened 2022-02-23
Backport static analysis fixes to 1.6

veewee:
Backported these fixes to resolve Type\shape() issues:

* https://github.com/azjezz/psl/commit/821391fe0539b2158efff4c72916d18d829be9a0
* https://github.com/azjezz/psl/commit/0118dbcd97660723af918999f0d4db4eca43292b

### DIFF
--- a/src/Psl/Iter/is_empty.php
+++ b/src/Psl/Iter/is_empty.php
@@ -7,7 +7,7 @@ namespace Psl\Iter;
 /**
  * Check whether an iterable is empty.
  *
- * @psalm-assert-if-true empty $value
+ * @psalm-assert-if-true empty $iterable
  */
 function is_empty(iterable $iterable): bool
 {
--- a/src/Psl/Math/mean.php
+++ b/src/Psl/Math/mean.php
@@ -9,7 +9,7 @@ use Psl\Iter;
 /**
  * Return the arithmetic mean of the numbers in the given iterable.
  *
- * @param iterable<numeric> $numbers
+ * @param iterable<int|float> $numbers
  */
 function mean(iterable $numbers): ?float
 {
@@ -20,8 +20,8 @@ function mean(iterable $numbers): ?float
 
     $mean = 0.0;
     foreach ($numbers as $number) {
-        $mean += $number / $count;
+        $mean += (float)$number / $count;
     }
 
-    return (float) $mean;
+    return $mean;
 }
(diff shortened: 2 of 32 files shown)


veewee:
@azjezz  any clue on the failing unit test?
It only fails on mac for some reason, and it didn't change in versions above...
Looks like a rounding issue

```
1) Psl\Tests\Math\TanTest::testTan with data set #0 (-3.380515006246586, 5.0)
Failed asserting that -3.380515006246585 is identical to -3.380515006246586.

/Users/runner/work/psl/psl/tests/Psl/Math/TanTest.php:17

```

Should be "-3.380515006246586" on all versions though : https://3v4l.org/Bq0B4
My mac also states it is "-3.380515006246585". So might be a mac issue, but then again - it should fail on 2.x and 1.9.x as well

azjezz:
I'm guessing this fails for PHP 7.4, but not +8.0? 🤔 no idea honestly.

If it's just this test case, we can suppress it for 1.6 branch.

veewee:
Tested it on all php versions on Mac and got the same error.
Don't understand why it does work on never versions of PSL, since nothing changed to the code or test..


Will suppress it on this branch

azjezz (approved):

azjezz:
@veewee can you release a new 1.6?

veewee:
Done! Thanks.

azjezz:
@veewee the MacOS issue exists in 2.0.x branch as well, showed up now after i ran CI :/

veewee:
Probably related with newer Mac versions and underlying libraries then... We could round the result to 5 digits in the test cases or report it to internals? The rounding issue doesn't seem that dramatic to me though. But then again: I haven't used cos,sin,tan in ages

jrmajor:
@veewee There's a dedicated PHPUnit method for this use case: `assertEqualsWithDelta()`. Unfortunately, there's nothing like `assertSameWithDelta()`.

Edit: looking at the source code, I think `assertEquals()` has a default delta: https://github.com/sebastianbergmann/comparator/blob/926815b015c4c459f8ec5185328eb34ade07c51f/src/DoubleComparator.php#L53-L55.

azjezz:
> assertEqualsWithDelta()

we shouldn't use this, the float value is correct, something weird is going on in MacOs that results in incorrect result.

PR #339, opened 2022-02-27
chore(psalm): update type hint

patoui:
This update fixes the psalm error from the GitHub actions:

https://github.com/azjezz/psl/runs/5347964245?check_suite_focus=true

P.S. first PR to this repo, please let me know if any additional changes or specific formats should be followed

### DIFF
--- a/tests/static-analysis/Iter/count.php
+++ b/tests/static-analysis/Iter/count.php
@@ -12,7 +12,7 @@ function take_positive_integer(int $_foo): void
 {
 }
 
-/** @param 0 $_foo */
+/** @param int $_foo */
 function take_zero(int $_foo): void
 {
 }


azjezz:
Thank you for contributing!

I should have opened an issue about this, this is not the correct fix.

the problem is in the [`php-standard-library/psalm-plugin`](https://github.com/php-standard-library/psalm-plugin) library, the psalm plugin should detect that `Iter\count` would return `0` when `array<empty, empty>` is passed as the first argument.

I think this broke at this commit https://github.com/php-standard-library/psalm-plugin/commit/dff80efbbfdf2268eff554b94055145ab5742a63

but i haven't looked into it yet :)

If you would like to send a PR to the psalm-plugin repository, and then a PR here to update the plugin in the dependencies, that would be great!

azjezz:
looking at it, this is what broke the test: https://github.com/php-standard-library/psalm-plugin/commit/dff80efbbfdf2268eff554b94055145ab5742a63#diff-bd20de44af4c46c4992070b2ab1dcb9ef23235150fd5e30a46c6fb4ece5f2d9cL71-R72, we replaced `isEmpty()` with `isNever()` but they don't care the same information.

azjezz:
The issue has been fixed 👍

PR #343, opened 2022-04-07
chore(dict): narrow `Dict\take` return type

jrmajor:


src/Psl/Dict/take.php
```diff
@@ -13,7 +13,7 @@
  * @param iterable<Tk, Tv> $iterable Iterable to take the elements from
  * @param int<0, max> $n Number of elements to take from the start
  *
- * @return array<Tk, Tv>
+ * @return ($iterable is list<Tv> ? list<Tv> : array<Tk, Tv>)
  */
 function take(iterable $iterable, int $n): array
```
azjezz, line 18: we are using `slice` here, does `slice()` return a list? if so, that's a bug. `Dict\` functions should always treat arrays given as dict, not vec.

but i see the use case, so maybe we can add `take`/`drop`/`drop` to `Psl\Vec` namespace?

jrmajor: > does slice() return a list? if so, that's a bug.

If you slice a list using `$start = 0`, it has to return a list, and it is the expected behavior.

> maybe we can add `take`/`drop`/`drop` to `Psl\Vec` namespace?

I can PR that :) Did you mean `take`/`drop`/`slice`?

azjezz: > If you slice a list using $start = 0, it has to return a list, and it is the expected behavior.

Ah, yes, that make sense.

> I can PR that :) Did you mean take/drop/slice?

Yes, that would be great 👍

PR #344, opened 2022-04-09
feat(vec): introduce slice(), take() and drop() functions

jrmajor:


src/Psl/Dict/drop.php
```diff
@@ -9,8 +9,8 @@
  *
  * Examples:
  *
- *      Dict\drop([1, 2, 3, 4, 5], 3)
- *      => Dict(4, 5)
+ *      Dict\drop(['a' => 1, 'b' => 2, 'c' => 3, 'd' => 4], 2)
+ *      => Dict('c' => 3, 'd' => 4)
```
jrmajor, line 13: I've used the old examples from `Dict\drop()` and `Dict\slice()` in their `Vec` counterparts, I think they make more sense there.

src/Psl/Internal/Loader.php
```diff
@@ -121,6 +121,7 @@ final class Loader
         'Psl\Vec\chunk',
         'Psl\Vec\chunk_with_keys',
         'Psl\Vec\concat',
+        'Psl\Vec\drop',
```
jrmajor, line 124: Not sure what's the correct order for these, they don't seem to be sorted alphabetically.

src/Psl/Vec/slice.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @param iterable<T> $iterable Iterable to take the slice from
+ * @param int<0, max> $start Start offset
+ * @param null|int<0, max> $length Length (if not specified all remaining values from the array are used)
+ *
+ * @return list<T>
+ */
+function slice(iterable $iterable, int $start, ?int $length = null): array
+{
+    $result = [];
```
azjezz, line 28: todo (for future PR if you would like): if `$iterable` is array, use `array_values(array_slice(...))` ( we would need to add a benchmark to see if there's a benefit first )

src/Psl/Internal/Loader.php
azjezz: it's random :p

azjezz (approved):

src/Psl/Vec/take.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace Psl\Vec;
+
+/**
+ * Take the first n elements from an iterable.
+ *
+ * Examples:
+ *
+ *      Vec\drop([1, 2, 3, 4], 2)
```
bendavies, line 12: wrong example

src/Psl/Dict/take.php
```diff
@@ -7,6 +7,11 @@
 /**
  * Take the first n elements from an iterable.
  *
+ * Examples:
+ *
+ *      Dict\drop(['a' => 1, 'b' => 2, 'c' => 3, 'd' => 4], 2)
```
bendavies, line 12: wrong example

src/Psl/Vec/take.php
jrmajor: Fixed in #354.

src/Psl/Dict/take.php
jrmajor: Fixed in #354.

PR #345, opened 2022-04-10
chore(filesystem): improve docblock descriptions

jrmajor:
Some of these changes are fixes of obvious mistakes, but the rest may be a bit opinionated. I'm open to discussing/reverting some of them :)


src/Psl/Filesystem/get_modification_time.php
```diff
@@ -16,6 +16,8 @@
  *
  * @throws Exception\NotFoundException If $node is not found.
  * @throws Exception\RuntimeException In case of an error.
+ *
+ * @return int The last content modification time as a Unix timestamp.
```
azjezz, line 20:
```suggestion
 * @return int<0, max> The last content modification time as a Unix timestamp.
```

azjezz (commented):
Thank you :)

one minor change.

src/Psl/Filesystem/get_modification_time.php
jrmajor:
```sh
touch -d "1960-05-19T12:00:00" test_file
```

```php
Psl\Filesystem\get_modification_time('test_file') === -303573600
```

jrmajor:
All `@param` annotations I checked are full sentences (begin with an uppercase letter and end with a period), but there are inconsistencies with `@return`. Do you have any preference on how they should be standardized?

azjezz:
> preference on how they should be standardized?

no, as long as things are consistent :)

src/Psl/Filesystem/get_modification_time.php
azjezz: ah, ignore me then :p

jrmajor:
Let's go with full sentences everywhere then :)

azjezz (approved):

PR #350, opened 2022-05-30
Fix Filesystem\get_extension() returning empty string

### LINKED ISSUE #349: `\Psl\Filesystem\get_extension()` can return empty string
**Describe the bug**
`\Psl\Filesystem\get_extension()` can return empty string

**To Reproduce**
```php
var_dump(\Psl\Filesystem\get_extension('test.')); // string(0) ""
```

**Expected behavior**
`null` to be returned.

**Additional context**
Not sure how this should be fixed.
One  aproach to fix this would be to change the return from `non-empty-string|null` to `string|null`.
Personally I would expect `null` to be returned for a file without an extension but with a `.` in the name (like in the example above).

I can make a PR once the direction is decided.

dragosprotung:
Fixes #349

### DIFF
--- a/src/Psl/Filesystem/get_extension.php
+++ b/src/Psl/Filesystem/get_extension.php
@@ -17,6 +17,7 @@ use function pathinfo;
  */
 function get_extension(string $node): ?string
 {
-    /** @var non-empty-string|null */
-    return pathinfo($node)['extension'] ?? null;
+    $extension = pathinfo($node)['extension'] ?? null;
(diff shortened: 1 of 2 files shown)


dragosprotung:
I am not sure if we should handle more exception cases like `test. ` (extension is only whitespace - yes unfortunatly it's possible) or `test. x ` (extension contains some white space).

azjezz (approved):

PR #356, opened 2022-07-02
feat(option): introduce option component

azjezz:

### DIFF
--- a/src/Psl/Option/Exception/NoneException.php
+++ b/src/Psl/Option/Exception/NoneException.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Option\Exception;
+
+use Psl\Exception\UnderflowException;
+
+final class NoneException extends UnderflowException implements ExceptionInterface
+{
+}
--- a/src/Psl/Option/Option.php
+++ b/src/Psl/Option/Option.php
@@ -1,0 +1,260 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Option;
+
+use Closure;
+
+/**
+ * @template T
+ */
+final class Option
+{
+    /**
+     * @param ?array{T} $option
+     *
+     * @internal
+     */
+    private function __construct(
+        private readonly null|array $option,
+    ) {
+    }
+
+    /**
+     * Create an option with some value.
+     *
+     * @template Tv
+     *
+     * @param Tv $value
+     *
+     * @return Option<Tv>
+     */
+    public static function some(mixed $value): Option
+    {
+        return new self([$value]);
+    }
+
+    /**
+     * Create an option with none value.
+     *
+     * @template Tn
+     *
+     * @return Option<Tn>
+     */
+    public static function none(): Option
+    {
+        return new self(null);
+    }
+
+    /**
+     * Returns true if the option is a some value.
+     */
+    public function isSome(): bool
+    {
+        return $this->option !== null;
+    }
+
+    /**
+     * Returns true if the option is a some and the value inside of it matches a predicate.
+     *
+     * @param (Closure(T): bool) $predicate
+     */
+    public function isSomeAnd(Closure $predicate): bool
+    {
+        return $this->option !== null && $predicate($this->option[0]);
+    }
+
+    /**
+     * Returns true if the option is a none.
+     */
+    public function isNone(): bool
+    {
+        return $this->option === null;
+    }
+
+    /**
+     * Returns the contained Some value, consuming the self value.
+     *
+     * because this function may throw, its use is generally discouraged.
+     * Instead, prefer to use `Option::unwrapOr()`, `Option::unwrapOrElse()`.
+     *
+     * @throws Exception\NoneException If the option is none.
+     *
+     * @return T
+     */
+    public function unwrap(): mixed
+    {
+        if ($this->option !== null) {
+            return $this->option[0];
+        }
+
+        throw new Exception\NoneException('Attempting to unwrap a none option.');
+    }
+
+    /**
+     * Returns the contained some value or the provided default.
+     *
+     * @note:   Arguments passed to `Option::unwrapOr()` are eagerly evaluated;
+     *          if you are passing the result of a function call, it is recommended to use `Option::unwrapOrElse()`, which is lazily evaluated.
+     *
+     * @param T $default
+     *
+     * @return T
+     */
+    public function unwrapOr(mixed $default): mixed
+    {
+        if ($this->option !== null) {
+            return $this->option[0];
+        }
+
+        return $default;
+    }
+
+    /**
+     * Returns the contained some value or computes it from a closure.
+     *
+     * @param (Closure(): T) $default
+     *
+     * @return T
+     */
+    public function unwrapOrElse(Closure $default): mixed
+    {
+        if ($this->option !== null) {
+            return $this->option[0];
+        }
+
+        return $default();
+    }
+
+    /**
+     * Return none if either `$this`, or `$other` options are none, otherwise returns `$other`.
+     *
+     * @template Tu
+     *
+     * @param Option<Tu> $other
+     *
+     * @return Option<Tu>
+     */
+    public function and(Option $other): Option
+    {
+        if ($this->option !== null && $other->option !== null) {
+            return $other;
+        }
+
+        return none();
+    }
+
+    /**
+     * Returns the option if it contains a value, otherwise returns $option.
+     *
+     * @note:   Arguments passed to `Option::or()` are eagerly evaluated;
+     *          if you are passing the result of a function call, it is recommended to use `Option::orElse()`, which is lazily evaluated.
+     *
+     * @param Option<T> $option
+     *
+     * @return Option<T>
+     */
+    public function or(Option $option): Option
+    {
+        if ($this->option !== null) {
+            return $this;
+        }
+
+        return $option;
+    }
+
+    /**
+     * Returns none if the option is none, otherwise calls `$predicate` with the wrapped value and returns:
+     *  - Option<T>::some() if `$predicate` returns true (where t is the wrapped value), and
+     *  - Option<T>::none() if `$predicate` returns false.
+     *
+     * @param (Closure(T): bool) $predicate
+     *
+     * @return Option<T>
+     */
+    public function filter(Closure $predicate): Option
+    {
+        if ($this->option !== null) {
+            return $predicate($this->option[0]) ? $this : none();
+        }
+
+        return $this;
+    }
+
+    /**
+     * Returns true if the option is a `Option<T>::some()` value containing the given value.
+     *
+     * @psalm-assert-if-true T $value
+     */
+    public function contains(mixed $value): bool
+    {
+        if ($this->option !== null) {
+            return $this->option[0] === $value;
+        }
+
+        return false;
+    }
+
+    /**
+     * Maps an `Option<T>` to `Option<Tu>` by applying a function to a contained value.
+     *
+     * @template Tu
+     *
+     * @param (Closure(T): Tu) $closure
+     *
+     * @return Option<Tu>
+     */
+    public function map(Closure $closure): Option
+    {
+        if ($this->option !== null) {
+            return some($closure($this->option[0]));
+        }
+
+        /** @var Option<Tu> */
+        return $this;
+    }
+
+    /**
+     * Applies a function to the contained value (if some),
+     * or returns `Option<Tu>::some()` with the provided `$default` value (if none).
+     *
+     * @note:   arguments passed to `Option::mapOr()` are eagerly evaluated;
+     *          if you are passing the result of a function call, it is recommended to use `Option::mapOrElse()`, which is lazily evaluated.
+     *
+     * @template Tu
+     *
+     * @param (Closure(T): Tu) $closure
+     * @param Tu $default
+     *
+     * @return Option<Tu>
+     */
+    public function mapOr(Closure $closure, mixed $default): Option
+    {
+        if ($this->option !== null) {
+            return some($closure($this->option[0]));
+        }
+
+        return some($default);
+    }
+
+    /**
+     * Applies a function to the contained value (if some),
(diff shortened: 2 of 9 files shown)


src/Psl/Option/Exception/NoneException.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Option\Exception;
+
+use Psl\Exception\UnderflowException;
+
+final class NoneException extends UnderflowException implements ExceptionInterface
```
azjezz, line 9: I don't like the exception name, not sure what else would work.

src/Psl/Option/Option.php
```diff
@@ -0,0 +1,260 @@
...
+ */
+final class Option
+{
+    /**
+     * @param ?array{T} $option
+     *
+     * @internal
+     */
+    private function __construct(
+        private readonly null|array $option,
```
azjezz, line 20: using `array{T}` instead of `T` to allow `null` as a valid value for `Option::some(null)`.

src/Psl/Option/Option.php
```diff
@@ -0,0 +1,260 @@
...
+     * Returns the contained some value or the provided default.
+     *
+     * @note:   Arguments passed to `Option::unwrapOr()` are eagerly evaluated;
+     *          if you are passing the result of a function call, it is recommended to use `Option::unwrapOrElse()`, which is lazily evaluated.
+     *
+     * @param T $default
+     *
+     * @return T
+     */
+    public function unwrapOr(mixed $default): mixed
```
veewee, line 105: you have both `unwrapOr` and `unwrapOrElse` which are basically the same, but accept another input argument type. This might be a bit confusing. Can we solve this by playing around with the name?

Maybe `unwrapOrFallback` and `unwrapOrCalculate`?

src/Psl/Option/Option.php
```diff
@@ -0,0 +1,260 @@
...
+     */
+    public function mapOrElse(Closure $closure, Closure $default): Option
+    {
+        if ($this->option !== null) {
+            return some($closure($this->option[0]));
+        }
+
+        return some($default());
+    }
+}
```
veewee, line 260: a `proceed` function could be added as well, to make it all consistent with result and either?

src/Psl/Option/Exception/NoneException.php
veewee, line 9: what about `UnwrapException::expectedSome()` ?

veewee (commented):
Looks nice!

src/Psl/Option/Option.php
someniatko: it looks like `unwrapOrElse()` has exactly the behavior of possible `proceed()`

src/Psl/Option/Option.php
someniatko: Interesting solution! Alternatively, it could be just an additional boolean flag indicating the value is none or some, but type-wise your solution is better I think.

With boolean flag there could be problems like here: https://github.com/marcosh/lamphpda/blob/master/src/Maybe.php#L118 (note `@psalm-suppress`)

src/Psl/Option/Option.php
someniatko: @veewee `unwrapOr()` accepts already computed default value, while `unwrapOrElse()` accepts a **callback computing the default value**, which means the former is non-lazy, and the latter is lazy, which might be desirable. Also, a callback may throw, so you can do it like

```php
$option->unwrapOrElse(fn() => throw new MyCustomException('Unexpected None'));
```

someniatko: generally I'd avoid too long names like `unwrapOrFallback` - these functions are expected to be used pretty often, so it's okay to shorten their names. Also as long as it's inspired by Rust, I don't see the reason of reinventing our names instead of Rust's, which already are good enough.

src/Psl/Option/Option.php
veewee: Not completely, proceed would have a callback for both some and  for none that resolve to a specific type

someniatko: @veewee `mapOrElse()` then

src/Psl/Option/Exception/NoneException.php
someniatko: `UnwrapException` / `UnwrapNoneException` ?

src/Psl/Option/Option.php
```diff
@@ -0,0 +1,260 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Option;
+
+use Closure;
+
+/**
+ * @template T
```
someniatko, line 10: it should be `@template-covariant` to accept subtypes. That's really useful in real applications: for example a function accepts `Option<Organization>`,  and you'd like to give it `Option<GovernmentOrganization>`.

someniatko: also, `@immutable` / `@psalm-immutable` is also desirable, because that allows to give an Option to constructors of `@immutable` classes.

someniatko: However, Psalm will then demand that all the Closures you give to Option's methods must be marked as pure (there is a pseudo-type in Psalm for that, `pure-callable`, btw Psalm doesn't have similar type exactly for Closure). I am not sure we'd want to define our API allowing only side-effect-free Closures though - then `@psalm-suppress ImpureFunctionCall` will have to be used in `map()`, `unwrapOrElse()` and similar methods.

azjezz: functions taking closures cannot be pure, you can suppress the issue locally if desired, but suppressing in PSL is not an option as the callback could still carry side effects, using pure-callable is also not an option ( though, i was the one to introduce that type into psalm ^^" ).

but i agree on immutable/covariant 👍

src/Psl/Option/Option.php
azjezz: process is mapOrElse, this follows the naming convention used in rust option type, not sure if we should use proceed, as in other instances ( result/either ), both proceed callbacks take a value, which is either `T`/`TError`, or `TRight`/`TLeft`, but here, the second callable doesnt' take an argument

src/Psl/Option/Option.php
azjezz: yes, this is a type-safe way of doing it, initially, it was `array{false}|array{true, T}`, but psalm doesn't really like that.

src/Psl/Option/Option.php
azjezz: purity in psalm is insanity: https://github.com/vimeo/psalm/issues/8212, as of now we can't declare option as immutable.

src/Psl/Option/Option.php
veewee: Not really, map returns an Option.
Proceed should be acombination of mapOrElse and unwrap.

src/Psl/Option/Option.php
someniatko: @azjezz hmm, in Psalm issue you wrote you can instead mark Option as `@readonly`. But that will not allow using it inside of `@immutable` classes.

azjezz: Hm 🤔 yes, the only way to actually have `Option` as immutable is to mark all Closure arguments are pure functions, which i'm not infavor of, since now you are *forced* to only use pure functions when using `Option` ( e.g: this is not possible anymore `$user->map(fn($u) => $repo->getUserOrgnizations($u));` )

veewee: You can add impure closures as arguments if you suppress all function calls inside the option class. It sucks, but it works...
We do that in our internal codebase as well

azjezz: the issue is end users would be forced to use pure functions only with option, which is not something i like to do

azjezz: ( same reason why `Vec\map` and friends are not pure )

veewee: No I meant you can use regular closures as param if you suppress the impure function call inside the option class.

That way you can provide impure functions to  the option method, even though it is immutable.

Not behind the computer for a couple of days, so it's hard to type an example on my phone  🙂

someniatko: Allowing non-pure closures also has its drawbacks, consider a mutable entity `User` with a method `setName()`:
```php
/** @var Option<User> */
$option->map(fn(User $user) => $user->setName('New name'));
```
which is not exactly what you'd expect from `map()`. However, restricting to only pure closures in PHP ecosystem is even worse. Therefore I guess, the best compromise is to suppress ImpureFunctionCall as @veewee suggests. PHP is not a functional language with clear side effects separation.

veewee: https://psalm.dev/r/066c2e385b

This took me ages to write on a phone 🤣
But as you can see, usercode can use impure  closures on immutable objects this way.

azjezz: that's not good, now psalm reports this as okay, while it's not: https://psalm.dev/r/52bae7b6bf

x($f) is impure, because $f is impure.

veewee: True ... That's the downside.

Not sure if that is something people would actually write, but nevertheless a good concern.


So it needs to be fixed in / added to psalm before it can be annotated as immutable-ish

someniatko: I am even thinking, maybe we should introduce `Option` accepting only pure callables, and another `MutableOption` which template is not covariant, and it accepts arbitrary callables?

veewee: Adding a mutable version would be a step backward imo.
Maybe not marking it as immutable for psalm is ok for now? It is immutable by design - so adding the typehint might be a bridge too far. Would that have any downsides?
The only thing I can think of is that you won't be able to use it in pure functions, unless we mark the static constructors as pure maybe?

someniatko: psalm does not allow non-immutable classes to be `@template-covariant`

azjezz: @someniatko it does! however, it won't allow them it accept callables that use the template as arguments :(

someniatko: > functions taking closures cannot be pure, you can suppress the issue locally if desired, but suppressing in PSL is not an option as the callback could still carry side effects

Do we even care if it carries side effects or not? As soon as it should not mutate the Option itself?

src/Psl/Option/Option.php
```diff
@@ -0,0 +1,260 @@
...
+     */
+    public static function some(mixed $value): Option
+    {
+        return new self([$value]);
+    }
+
+    /**
+     * Create an option with none value.
+     *
+     * @template Tn
```
Ocramius, line 41: This is problematic, since it requires an explicit `/** @var */` everywhere

azjezz: having `@return Option<null>` wouldn't be a good idea either :/ wish psalm had something like a wildcard for this specific case `@return Option<_>`

Ocramius: One thing I did in my example from today was:

```php
/**
 * @template T
 * @param \Psl\Type\TypeInterface<T>
 * @return self<T>
 */
public static function empty(\Psl\Type\TypeInterface $type)
```

That can be used for type inference

azjezz: oh, that's a nice idea!

someniatko: if Option's template was covariant, `Option<never>` would also work: https://github.com/someniatko/php-result-type/blob/master/src/None.php

veewee: @someniatko 
Splitting None and Some classes looks like a better approach as well here! 
That way, it's consistent with the PSL Result classes as well.

---

I wished we had built-in support for composed union types in PHP like:

```
type Option=Some<T>|None;
```

(Which is closer to None<never>)

Ocramius:
Just cross-linking here: I've provided a full example @ https://gist.github.com/Ocramius/368bd78c38cc5bb857e931801f6e4af7

Beware that `Optional` is **not** necessarily `OptionalField`.

 * `Optional` in Java attempted to abstract `T|null` (because Java sucks at handling `null` - it's so so bad!).
 * in my example, I attempt to abstract `T|null|absent`, where `absent`, which is a slightly different problem :+1:

azjezz:
> forPossiblyMissingArrayKey

I was thinking with the addition of `Option`, we could add the following functions to `Dict` and `Vec` components, which would implement basically that.

```php
Dict\get<Tk, Tv>(dict<Tk, Tv> $d, Tk $k): Option<Tv> {
  return Iter\contains_key($d, $k) ? Option\some($d[$k]) : Option\none();
}

Dict\get_typed<T>(dict<array-key, mixed> $d, array-key $k, Type\TypeInterface<T> $t): Option<T> {
  return Iter\contains_key($d, $k) ? Option\some($t->coerce($d[$k])) : Option\none();
}

Vec\get<T>(vec<T> $v, int $k): Option<T> {
  return Iter\contains_key($v, $k) ? Option\some($v[$k]) : Option\none();
}

Vec\get_typed<T>(vec<mixed> $v, int $k, Type\TypeInterface<T> $t): Option<T> {
  return Iter\contains_key($v, $k) ? Option\some($t->coerce($v[$k])) : Option\none();
}
```

azjezz:
> in my example, I attempt to abstract T|null|absent, where absent, which is a slightly different problem 👍

To note, `null` is a valid "Some" in this implementation of Option, where `Option\some(null)->isNone()` is `false`, it's up to you to create a "none" in case of null if that is what you desire.

src/Psl/Option/Exception/NoneException.php
veewee: Keeping this one for now seems ok.

src/Psl/Option/Option.php
veewee: We can always add later if we find the need for this.

veewee (approved):

veewee:
Discussed the open topics in this PR with @azjezz. There is nothing blocking that is stopping us from merging and releasing this PR. We can always take in additional improvements once people start using this component and start seeing better ways to do things.

PR #367, opened 2022-11-03
fix(type): add more valid float coerceings

bcremer:
Tweaked the float coercing regex so more stringy floats can be detected as valid.
Especially negative and positive exponent notation like the following:

```php
        yield ['1.23e-45', 1.23e-45];
        yield ['1.23e+45', 1.23e+45];
```


azjezz (dismissed):

azjezz (approved):

azjezz:
Thanks @bcremer 👍

PR #368, opened 2022-11-20
chore: introduce rust extension

azjezz:


src/Psl/Math/from_base.php
```diff
@@ -22,10 +25,10 @@
 function from_base(string $number, int $from_base): int
 {
     /** @psalm-suppress MissingThrowsDocblock */
-    $limit  = div(INT64_MAX, $from_base);
+    $limit  = intdiv(INT64_MAX, $from_base);
     $result = 0;
-    foreach (Byte\chunk($number) as $digit) {
-        $oval = Byte\ord($digit);
+    for ($i = 0, $l = strlen($number); $i < $l; $i++) {
+        $oval = ord($number[$i]);
```
azjezz, line 31: Thanks @arnaud-lb

.github/workflows/benchmarks.yml
```diff
@@ -5,42 +5,29 @@ on:
...
 
     strategy:
       matrix:
-        php-version:
+        php:
           - "8.1"
-        operating-system:
+        os:
           - "ubuntu-latest"
+          - "macos-latest"
```
azjezz, line 19: It seems that the extension is not loaded when running benchmarks on macOS, not sure why yet ( @veewee maybe you can have a look at that? )

azjezz:
blocked by phper-framework/phper#64

.github/workflows/benchmarks.yml
veewee: On mac, a compilation results in `libpsl.dylib` instead of `libpsl.so`.

The formats are interchangeable - so you should be able to add this step to make compile:

`mv libpsl.dylib libpsl.so || true`

Locally, This results in:

Screenshot 2022-11-21 at 07 25 43

However, if I do not run this additional step. I get this error on second benchmark run:

> azjezz/psl/vendor/phpbench/phpbench/lib/Environment/Provider/template/php.template" did not return an array, got:
  Warning: PHP Startup: Unable to load dynamic library 'extension/target/release/libpsl.so'


Not sure if I need to do additional things to compile. Tried setting the same env vars as the setup script:

```
LIBCLANG_PATH=/usr/local/opt/llvm@13/bin
SDKROOT=$(xcrun --show-sdk-path)
```

Note: needed to disable opcache here:

> azjezz/psl/vendor/phpbench/phpbench/lib/Environment/Provider/template/php.template" did not return an array, got:
  Warning: JIT is incompatible with third party extensions that override zend_execute_ex(). JIT disabled. in Unknown on line 0

azjezz: > JIT is incompatible with third party extensions that override zend_execute_ex()

the extension is compatibly with JIT, it's already used in CI with linux, and i use it locally, are you sure pcov and xdebug are disabled? ( both are not compatible with JIT iirc )

veewee: Yeah, pcov was enabled..

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,68 @@
...
+        if: "contains(inputs.os, 'macos')"
+        run: echo "SDKROOT=$(xcrun --show-sdk-path)" >> $GITHUB_ENV
+
+      - name: "caching dependencies"
+        uses: "actions/cache@v3"
+        with:
+          path: |
+            ~/.composer/cache
+            vendor
+            ~/.cargo
```
simonsan, line 60: There is https://github.com/Swatinem/rust-cache for this

azjezz:
will be reconsidered in the future.

PR #372, opened 2022-12-05
Adjusted testcases for better pipe plugin

veewee:
See https://github.com/php-standard-library/psalm-plugin/pull/8

The static-analysis tests are slightly adjusted to ensure better coverage.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "phpbench/phpbench": "^1.2.3",
         "phpunit/phpunit": "^9.5.16",
         "vimeo/psalm": "^5.0.0",
-        "php-standard-library/psalm-plugin": "^2.1.0",
+        "php-standard-library/psalm-plugin": "^2.2",
         "php-coveralls/php-coveralls": "^2.5.2",
         "roave/infection-static-analysis-plugin": "^1.26.0"
     },
--- a/src/Psl/Network/ServerInterface.php
+++ b/src/Psl/Network/ServerInterface.php
@@ -27,7 +27,7 @@ interface ServerInterface extends IO\CloseHandleInterface
      *
      * @throws Exception\RuntimeException In case failed to accept incoming connection.
      *
-     * @return Generator<null, SocketInterface, void, void>
+     * @return Generator<null, SocketInterface, void, null>
      */
     public function incoming(): Generator;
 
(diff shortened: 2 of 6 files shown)


composer.json
```diff
@@ -57,5 +57,11 @@
     },
     "suggest": {
         "php-standard-library/psalm-plugin": "Psalm integration"
-    }
+    },
+    "repositories": [
+        {
+            "type": "git",
```
veewee: Will revert this if psalm plugin is merged.

tests/static-analysis/Fun/pipe.php
```diff
@@ -46,11 +44,12 @@ function test_variadic_and_default_params(): void
  * @see https://github.com/vimeo/psalm/issues/7244
  *
  * @psalm-suppress InvalidArgument
+ * @psalm-suppress NoValue - Resolves into "\Closure(never): never" because of the issue linked above.
  */
-function test_empty_pipe(): void
+function test_empty_pipe(): string
```
veewee, line 49: This is still an issue....  Currently no way around this.

https://github.com/vimeo/psalm/issues/7244

src/Psl/Network/ServerInterface.php
```diff
@@ -27,7 +27,7 @@ public function nextConnection(): SocketInterface;
      *
      * @throws Exception\RuntimeException In case failed to accept incoming connection.
      *
-     * @return Generator<null, SocketInterface, void, void>
+     * @return Generator<null, SocketInterface, void, void|null>
```
azjezz: can we make this `null` as well ( same for the rest ).

azjezz:
plugin is merged.

src/Psl/Network/ServerInterface.php
veewee: Done

composer.json
veewee: Shall I release psalm-plugin as 2.2 (new feature) and after that this PSL as 2.3.1 (no changes besides tests)?

azjezz: yes

azjezz (dismissed):

azjezz (approved):

azjezz:
let's not release this, as i don't see any benefits down stream, the require-dev constraint is not enforced for users.

veewee:
Was thinking the same this morning. good!

PR #373, opened 2022-12-12
feat(option): introduce `Option\from_nullable` function 

veewee:
Introduces a new `Option\from_nullable()` factory that takes a `T|null` and converts it into an `Option<T>`

```php
/**
 * @param string|null $x
 *
 * @return Option<string>
 */
function foo($x): Option
{
    return Option\from_nullable($x);
}

```

it can be used as a shortcut function to opt-in on the Option component from functions that return something `nullable`.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -29,6 +29,7 @@
 - [Psl\Network](./component/network.md)
 - [Psl\OS](./component/os.md)
 - [Psl\Observer](./component/observer.md)
+- [Psl\Option](./component/option.md)
 - [Psl\Password](./component/password.md)
 - [Psl\Promise](./component/promise.md)
 - [Psl\PseudoRandom](./component/pseudo-random.md)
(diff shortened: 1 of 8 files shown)


jdreesen:
The title is wrong ;)

azjezz (approved):
great addition 👍

veewee:
Thanks @azjezz 
Shouldn't it have gone in 2.4 though?

Might be easier to  maintain a "vnext" or regular main branch instead of setting git's default branch to a specific version?

azjezz:
ouch 😬 yea, i think adding a `next` branch would be a good idea. i have done so and set it as a default branch, ~you should be able to revert this change from the `2.3.x` branch.~ i have done so, i reverted two commits from 2.3.0, now they are part of `next`

azjezz:
once `2.4.0` is released, we can checkout `2.4.x` branch.

veewee:
Alright awesome!

PR #384, opened 2022-12-20
Return might be non-empty-list/non-empty-array for map functions

dragosprotung:
If the input is `non-empty` the return will also be `non-empty-*`.

`iterable` is a more generic type, but the conditional return still helps with better understanding the return value in certain scenarios.


azjezz (dismissed):

dragosprotung:
I targeted by mistake the `next` branch

azjezz (approved):

PR #394, opened 2023-01-15
feat(type): add signed integer types

azjezz:


src/Psl/Math/constants.php
```diff
@@ -96,7 +96,7 @@
  *
  * @var int
  */
-const INT8_MAX = 128;
+const INT8_MAX = 127;
```
azjezz, line 99: i just noticed this...

PR #396, opened 2023-01-19
feat(type): add f32 and f64 types

KennedyTedesco:


azjezz (commented):
Thanks @KennedyTedesco

PR #398, opened 2023-01-24
Improve options type declarations and add andThen method

### LINKED ISSUE #379: andThen()
Introduce new `andThen` method: https://doc.rust-lang.org/rust-by-example/error/option_unwrap/and_then.html
This method can be used to e.g. map an Option into another option:

```php
some(1)->andThen(fn(int $value) => $value > 1 ? some($value) : none())  == none();
```

They apply to:

* [ ] Result
* [x] Option
* [ ] maybe:  Promise?

❗ Currently Result::then() allows returning a promise which automatically gets unwrapped. That logic should be deprecated and removed in v3 since it's rather magically. The new `andThen` function should be used for this purpose. We'll have to add a note for that in the code and release notes.

(to investigate: if it makes sense to keep on supporting it during runtime, but make psalm fail)

veewee:
Partially fixes #379:

```php
$option = Option\some(2);

static::assertSame(3, $option->andThen(static fn($i) => Option\some($i + 1))->unwrapOr(null));
```

Improves types of `unwrapOr` and `unwrapOrElse`:

```php
Option::<T>::unwrapOr::<O>(O $other): T|O
```

### DIFF
--- a/src/Psl/Option/Option.php
+++ b/src/Psl/Option/Option.php
@@ -98,9 +98,11 @@ final class Option
      * @note:   Arguments passed to `Option::unwrapOr()` are eagerly evaluated;
      *          if you are passing the result of a function call, it is recommended to use `Option::unwrapOrElse()`, which is lazily evaluated.
      *
-     * @param T $default
+     * @template O
      *
-     * @return T
+     * @param O $default
+     *
+     * @return T|O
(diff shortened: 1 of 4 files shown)


azjezz (approved):
LGTM 🚢

PR #400, opened 2023-01-26
Undeprecate `positive_int()`

### LINKED ISSUE #399: Undeprecate `positive_int`
`positive_int` means `int<1, max>`
`uint` means `int<0, max>`

I think we should keep `positive_int` because it represents a different type.
There is not alternative to represent `positive-int` (for example like an intersection).
Using `uint` instead of `positive_int` means manual checks to see if the value is also `0`

Do not deprecate `positive_int`

dragosprotung:
Closes #399

### DIFF
--- a/src/Psl/Type/positive_int.php
+++ b/src/Psl/Type/positive_int.php
@@ -7,7 +7,7 @@ namespace Psl\Type;
 /**
  * @return TypeInterface<positive-int>
  *
- * @deprecated use {@see uint} instead.
+ * @ara-return TypeInterface<0|uint>
  */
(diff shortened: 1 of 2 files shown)


src/Psl/Type/positive_int.php
```diff
@@ -6,8 +6,6 @@
 
 /**
  * @return TypeInterface<positive-int>
```
azjezz, line 8: please add `@ara-return TypeInterface<0|uint>` here.

azjezz (approved):

azjezz:
released as 2.4.1

PR #401, opened 2023-02-27
Imrovements on file creation during write modes

veewee:
There are some inconsistencies between `ReadWriteHandle` and `WriteHandle`.
The first one tries to create the file (and directories), the other does not.
The logic inside the handles are not in line with the docblocks above `WriteMode`: If all write-modes are trying to create the file, it doesn't make sense to disallow creation during e.g. `APPEND` and `TRUNCATE`.

This PR is a little draft to discuss a better approach that is consistent in between both handles and in line with the dockblocks of the write mode.

### DIFF
--- a/tests/unit/File/ReadWriteHandleTest.php
+++ b/tests/unit/File/ReadWriteHandleTest.php
@@ -10,7 +10,6 @@ use Psl\File;
 use Psl\Filesystem;
 use Psl\IO;
 use Psl\OS;
-use Psl\SecureRandom;
 
 final class ReadWriteHandleTest extends TestCase
 {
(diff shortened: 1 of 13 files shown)


tests/unit/File/ReadWriteHandleTest.php
```diff
@@ -90,10 +89,22 @@ public function testMustCreateExistingFile(): void
...
-        $this->expectException(File\Exception\NotFoundException::class);
-        $this->expectExceptionMessage('is not found.');
+        $temporary_file = Filesystem\create_temporary_file();
+        Filesystem\delete_file($temporary_file);
+
+        static::assertFalse(Filesystem\is_file($temporary_file));
 
-        new File\ReadWriteHandle(Env\temp_dir() . '/' . SecureRandom\string(20), File\WriteMode::APPEND);
+        $handle = File\open_read_write($temporary_file, File\WriteMode::APPEND);
+        $handle->tryWrite('hello');
```
azjezz:
```suggestion
        $handle->writeAll('hello');
```

try write can fail without an exception

azjezz (commented):
overall looks good, I'll have another look tomorrow when I'm a bit more caffeinated 😁

PR #403, opened 2023-02-28
Introduce Result\try_catch shortcut

veewee:
Inspired on ramda's tryCatch function:
https://ramdajs.com/docs/#tryCatch

> tryCatch takes two functions, a tryer and a catcher. The returned function evaluates the tryer; if it does not throw, it simply returns the result. If the tryer does throw, the returned function evaluates the catcher function and returns its result.

It can be seen as a shortcut for dealing with code that you know might error, for which you want to continue the flow with e.g. a default or calculated default value.

For example:

```php
$nullable_entity = Result\try_catch(
    static fn (): MyEntity => $repo->fetchById(1),
    static fn (): ?MyEntity => null,
);
```

I decided not to go the `Fun` way as it is in ramda, since it would require another dynamic psalm plugin.
Instead, you could make it composable yourself like this:

```php
$tryFetchById = static fn(int $id): ?MyEntity => Result\try_catch(
    static fn (): MyEntity => $repo->fetchById($id),
    static fn (): ?MyEntity => null,
);
```

It also works fine with `Option\from_nullable()` as you can from the examples above.

❗  Note: I inlined the `PromiseInterface` methods into the `ResultInterface` and made sure the return type for them is `ResultInterface`. Otherwise psalm thinks `ResultInterface::catch()` (and others) returns a `PromiseInterface` instead of a `ResultInterface`.

### DIFF
--- a/docs/component/result.md
+++ b/docs/component/result.md
@@ -14,6 +14,7 @@
 
 - [collect_stats](./../../src/Psl/Result/collect_stats.php#L14)
 - [reflect](./../../src/Psl/Result/reflect.php#L24)
+- [try_catch](./../../src/Psl/Result/try_catch.php#L24)
 - [wrap](./../../src/Psl/Result/wrap.php#L20)
 
 #### `Interfaces`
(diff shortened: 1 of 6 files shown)


azjezz (approved):
looks awesome!

PR #404, opened 2023-03-17
Allow an empty string to be coerced to bool

BackEndTea:
Since casting `false` to a string results in an empty string, it would feel fair to be able to coerce an empty string into false.

### DIFF
--- a/src/Psl/Type/Internal/BoolType.php
+++ b/src/Psl/Type/Internal/BoolType.php
@@ -34,7 +34,7 @@ final class BoolType extends Type\Type
             return $value;
         }
 
-        if (0 === $value || '0' === $value) {
+        if (0 === $value || '0' === $value || '' === $value) {
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


src/Psl/Type/Internal/BoolType.php
```diff
@@ -34,7 +34,7 @@ public function coerce(mixed $value): bool
             return $value;
         }
 
-        if (0 === $value || '0' === $value) {
+        if (0 === $value || '0' === $value || '' === $value) {
```
veewee, line 37: Isn't it a bit strange that only empty-string's will be allowed, but non-empty strings (except '0'|'1') result in a `CoercionException`?
Maybe an additional check for non-empty string could be added in the next if statement?

But that might break the intended way this type works internally... I'll leave that to @azjezz

azjezz: Yea, this in my opinion is not desired, i think `CoercionException` for `''` is the correct behaviour.

veewee:
Closing this one.
It is possible to impement it with the newly introduced `converted` type:

```php
$customBoolType = Type\converted(
    Type\string(),
    Type\bool(),
    static function (string $value): bool {
        return !in_array($value, ['', 'False', 'No', '0'], true);
    }
);
```

Issue #407, opened 2023-06-05
base64 url safe encoding


Gashmob:
**Is your feature request related to a problem? Please describe.**
For the implementation of WebAuthn in a web app, I need to send a key id in the uri. So I use base64 encoding for that. But following the [rfc](https://www.rfc-editor.org/rfc/rfc4648) the encoding uses + and / characters which are not really safe for an uri. Section 5 of rfc describe an url version of encoding using - and _ characters. Is it possible to implement it inside psl?

**Describe the solution you'd like**
It will be nice to have functions like this:
```php
function encodeUrl(string $binary): string;

function decodeUrl(string $base64): string;
```

**Describe alternatives you've considered**
Currently, as a workaround I'm doing that:
```php
strtr(encode($key_id), '+/', '-_');
```

azjezz:
I'm okay with adding two new functions to `Psl\Encoding\Base64`, or a new `Psl\Encoding\Base64\Url` sub-component.

a PR would be welcome.


```php
use Psl;
use Psl\Encoding\Base64;

$data = '...';

$encoded = Base64\url_encode($data);
$decoded = Base64\url_decode($encoded);

Psl\invariant($data === $decoded, 'expected $decoded to be the same as $data');
```

or


```php
use Psl;
use Psl\Encoding\Base64\Url;

$data = '...';

$encoded = Url\encode($data);
$decoded = Url\decode($encoded);

Psl\invariant($data === $decoded, 'expected $decoded to be the same as $data');
```

PR #408, opened 2023-06-05
feat(encoding): introduce `Base64\Variant` enum to support encoding/decoding different variants

Gashmob:
Closes issue #407

This pr introduces functions `url_encode` and `url_decode` in `Psl\Encoding\Base64` namespace. Following [rfc4648](https://www.rfc-editor.org/rfc/rfc4648) it uses base64url alphabet to make base64 encoding url safe.


src/Psl/Encoding/Base64/url_encode.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Encoding\Base64;
+
+/**
+ * Convert a binary string into a base64-encoded string url safe.
```
azjezz:
```suggestion
 * Convert a binary string into a url safe base64-encoded string.
```

src/Psl/Encoding/Base64/url_decode.php
```diff
@@ -0,0 +1,26 @@
...
+ * @pure
+ *
+ * @throws Exception\RangeException If the encoded string contains characters outside
+ *                                  the base64url characters range.
+ * @throws Exception\IncorrectPaddingException If the encoded string has an incorrect padding.
+ */
+function url_decode(string $base64url): string
+{
+    /** @var string */
+    return decode(strtr($base64url, '-_', '+/'));
```
veewee: Should we be dealing with padding?

According to [RFC 4648 - section 5](https://datatracker.ietf.org/doc/html/rfc4648#section-5), the `=` symbol can be used in URLs as well.

Since PHP's base64 decode [deals with padding internally](https://github.com/php/php-src/blob/79d024ac0e2a07b93362a5a2fd4ef36f61fca8fb/ext/standard/base64.c#L1243-L1261), padding could be removed as well.

> The pad character "=" is typically percent-encoded when used in an
   URI [[9](https://datatracker.ietf.org/doc/html/rfc4648#ref-9)], but if the data length is known implicitly, this can be
   avoided by skipping the padding; see [section 3.2](https://datatracker.ietf.org/doc/html/rfc4648#section-3.2).

Gashmob: Can we assert that user known the data length implicitly? base64url spec keep padding by default.

Instead we can add an optional argument `bool $padding = true`?

src/Psl/Encoding/Base64/url_decode.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Encoding\Base64;
+
+use Psl\Encoding\Exception;
```
azjezz:
```suggestion
use Psl\Encoding\Exception;

use function strtr;
```

src/Psl/Encoding/Base64/url_encode.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Encoding\Base64;
```
azjezz:
```suggestion
namespace Psl\Encoding\Base64;

use function strtr;
```

src/Psl/Encoding/Base64/url_decode.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Decode a base64url-encoded string into raw binary.
```
azjezz:
```suggestion
 * Decode a url safe base64-encoded string into raw binary.
```

src/Psl/Encoding/Base64/url_decode.php
azjezz: I think `bool $padding = false` as an optional argument is the right call.

wdyt @veewee ?

Gashmob: Just to be sure, when `false`, we remove padding in encode and add it again in decode, that's it?

veewee: I've got a suggestion for a slightly different approach:

```php

enum PaddingType {
    case Explicit;
    case Implicit;
}

// Encoding trims padding based on padding type
Base64\encode(
    string $encoded,
    PaddingType $padding = PaddingType::Explicit
): string;

// Proxies to Base64\Encode to trim padding
Base64\url_encode(
    string $binary,
    PaddingType $padding = PaddingType::Explicit
): string;


// Only performs the padding validation if padding is explicit.
// Otherwise it accepts any padding implicitly
Base64\decode(
    string $base64,
    PaddingType $padding = PaddingType::Explicit
): string;

// Proxies to Base64\decode to handle padding
Base64\url_decode(
    string $base64Url,
    PaddingType $padding = PaddingType::Explicit
): string;


```


Since PHP's base64 decode deals with padding - you can avoid running the padding validation step in current implementation of `Base64\decode` when `PaddingType::Implicit`:

* See https://3v4l.org/nZZrv
* https://github.com/azjezz/psl/blob/97341582b555c18ac9f46f7958fda47358f2190a/src/Psl/Encoding/Base64/decode.php#L36-L41


What do you think @azjezz ?

Gashmob: Is the enum really needed? With only 2 constants you can replace it with just a boolean:
```php
url_encode(string $binary, bool $explicitPadding = false): string
url_decode(string $base64url, bool $explicitPadding = false): string
```

veewee: There won't be a 3rth option, so a boolean might make sense as well..
The `explicit` padding should be the default to keep the function BC compatible though.

Gashmob: So in commit 4ed0d0e, I've added an argument for padding.

For encoding the argument is name `padding` to tell if we add padding or not. For decoding it's `explicitPadding` to tell if we check padding or not.
I've set these arguments to `true` for base64 and to `false` for base64url.

veewee: 🤔 Just wondering 2 things: 

* does it make sense to set these settings differently for regular vs url encoding? That would be rather inconsistent no?
* does it make sense to have 2 different names for the same thing in encode vs decode? Explicit padding works there as well maybe?

Gashmob: - In my opinion, we don't have the same use between regular and url encoding. For regular encoding we want to keep padding as it can be needed (and rfc don't tell that we can remove it). However, for url encoding, as we pass the result in an url, to avoid percent encoding of = characters, it can be better to remove it by default.
- I see the things like that :
    - for encoding, we keep or not the padding, so the parameter tell if we want it or not.
    - for decoding, we have padding or not, so the parameter tell if we check it or not

azjezz: this conversation made me realize that i already implemented this somewhere else.

see:

https://github.com/nuxed/crypto/blob/develop/src/Nuxed/Crypto/Base64/encode.hack
https://github.com/nuxed/crypto/blob/develop/src/Nuxed/Crypto/Base64/decode.hack
https://github.com/nuxed/crypto/blob/develop/src/Nuxed/Crypto/Base64/UrlSafe/encode.hack
https://github.com/nuxed/crypto/blob/develop/src/Nuxed/Crypto/Base64/UrlSafe/decode.hack

and since we are adding more variants, it would make more sense to also add 

https://github.com/nuxed/crypto/tree/develop/src/Nuxed/Crypto/Base64/DotSlash/encode.hack
https://github.com/nuxed/crypto/tree/develop/src/Nuxed/Crypto/Base64/DotSlash/decode.hack
https://github.com/nuxed/crypto/tree/develop/src/Nuxed/Crypto/Base64/DotSlash/Ordered/encode.hack
https://github.com/nuxed/crypto/tree/develop/src/Nuxed/Crypto/Base64/DotSlash/Ordered/decode.hack

however, this would mean we end up with too many functions, what if we do:


```php
namespace Psl\Encoding\Base64;

enum Variant {
  case Default;
  case UrlSafe;
  case DotSlash;
  case DotSlashOrdered;
}

function encode(string $data, Variant $variant = Variant::Default, bool $pad = false): string { ... }
function decode(string $data, Variant $variant = Variant::Default, bool $pad = false): string { ... }
```

wdyt?

veewee: Sounds good.
How do you see it implemented? 
Would it mean that we'll end up with 4 `Internal` base64 variant implementations?

Gashmob: Interesting! I don't know dot slash encoding neither hack language. I think I see how to do simple dot slash (as for url safe just replace +/ by ./), but for dot slash ordered it seems more complicated to me. The solution I see is implement all base64 encoding as you have done for Nuxed/crypto. For that we can take inspiration from https://github.com/paragonie/constant_time_encoding

For example :

```php
enum Variant {
    case Default;
    case UrlSafe;
    case DotSlash;
    case DotSlashOrdered;
}

function encode(string $data, Variant $variant = Variant::Default, bool $pad = false): string {
    for (...) {
        // ...
        encode6Bits($chunk, $variant);
       // ...
    }
}

function encode6Bits(string $chunk, Variant $variant): string {
    switch ($variant) {
    case Default:
        return encode6BitsDefault($chunk);
    case UrlSafe:
        return encode6BitsUrlSafe($chunk);
    // ...
    }
}
```

And same for decoding. It's a solution without use of class. Instead if we use class we can do exactly the same as for Nuxed but we are not bc. Which one do you prefer ? Or have you another idea ?

azjezz: I think we can take Nuxed code and use it here ( but using one singular function instead of having multiple ), and it would be BC, as the result of encode/decode is identical, it's just that Nuxed implementation is time constant, and if i remember correctly, Nuxed `Base64` is actually a hack re-write of the implementation in https://github.com/paragonie/constant_time_encoding.

If we are doing so, please make sure to add a link crediting https://github.com/paragonie/constant_time_encoding ( e.g: https://github.com/azjezz/psl/blob/next/src/Psl/Async/Awaitable.php#L17-L24 )

> Would it mean that we'll end up with 4 Internal base64 variant implementations?

Yes, similar to how `Nuxed\Crypto` has `Nuxed\Crypto\_Private`, we would have `Psl\Encoding\Base64\Internal`.

azjezz: > And same for decoding. It's a solution without use of class. 

Using internal classes is okay, multiple components do have classes in their internal namespace.

Gashmob: :+1: Ok, i will implement that this evening (and maybe push the commit as well this evening if all goes well)

src/Psl/Encoding/Base64/Internal/Base64DotSlashOrdered.php
```diff
@@ -0,0 +1,75 @@
...
+{
+    /**
+     * @pure
+     */
+    protected static function checkRange(string $base64): void
+    {
+        /** @psalm-suppress MissingThrowsDocblock - pattern is valid */
+        if (!Regex\matches($base64, '%^[a-zA-Z.-9]*={0,2}$%')) {
+            throw new Exception\RangeException(
+                'The given base64dotSlashOrdered string contains characters outside the base64dotSlashOrdered range.'
```
azjezz:
```suggestion
                'The given string contains characters outside the base64 range for the current variant.'
```

src/Psl/Encoding/Base64/Internal/Base64DotSlash.php
```diff
@@ -0,0 +1,79 @@
...
+{
+    /**
+     * @pure
+     */
+    protected static function checkRange(string $base64): void
+    {
+        /** @psalm-suppress MissingThrowsDocblock - pattern is valid */
+        if (!Regex\matches($base64, '%^[a-zA-Z0-9./]*={0,2}$%')) {
+            throw new Exception\RangeException(
+                'The given base64dotSlash string contains characters outside the base64dotSlash range.'
```
azjezz:
```suggestion
                'The given string contains characters outside the base64 range for the current variant.'
```

src/Psl/Encoding/Base64/Internal/Base64UrlSafe.php
```diff
@@ -0,0 +1,83 @@
...
+{
+    /**
+     * @pure
+     */
+    protected static function checkRange(string $base64): void
+    {
+        /** @psalm-suppress MissingThrowsDocblock - pattern is valid */
+        if (!Regex\matches($base64, '%^[a-zA-Z0-9-_]*={0,2}$%')) {
+            throw new Exception\RangeException(
+                'The given base64url string contains characters outside the base64url range.'
```
azjezz:
```suggestion
                'The given string contains characters outside the base64 range for the current variant.'
```

src/Psl/Encoding/Base64/decode.php
```diff
@@ -17,29 +13,30 @@
...
+ * 0x41-0x5a, 0x61-0x7a, 0x30-0x39, 0x2d, 0x5f
+ *
+ * Base64dotSlash character set:
+ * ./         [A-Z]      [a-z]     [0-9]
+ * 0x2e-0x2f, 0x41-0x5a, 0x61-0x7a, 0x30-0x39
+ *
+ * Base64dotSlashOrdered character set:
+ * [.-9]      [A-Z]      [a-z]
+ * 0x2e-0x39, 0x41-0x5a, 0x61-0x7a
+ *
```
azjezz: we can move the character set description to the enum instead.

src/Psl/Encoding/Base64/decode.php
```diff
@@ -17,29 +13,30 @@
...
  * @pure
  *
  * @throws Exception\RangeException If the encoded string contains characters outside
  *                                  the base64 characters range.
  * @throws Exception\IncorrectPaddingException If the encoded string has an incorrect padding.
  */
-function decode(string $base64, bool $explicitPadding = true): string
+function decode(string $base64, Variant $variant = Variant::Default, bool $strictPadding = true): string
```
azjezz:
```suggestion
function decode(string $base64, Variant $variant = Variant::Default, bool $explicit_padding = true): string
```

src/Psl/Encoding/Base64/encode.php
```diff
@@ -17,13 +13,12 @@
  *
  * @pure
  */
-function encode(string $binary, bool $padding = true): string
+function encode(string $binary, Variant $variant = Variant::Default, bool $padding = true): string
```
azjezz, line 12: the charset documentation can be removed from this function as well, and moved to the variant enum.

src/Psl/Encoding/Base64/Internal/Base64.php
```diff
@@ -0,0 +1,213 @@
...
+ * The following class was derived from code of Paragon Initiative Enterprises.
+ *
+ * https://github.com/paragonie/constant_time_encoding/blob/198317fa6db951dd791be0740915dae878f34b3a/src/Base64.php
+ *
+ * Code subject to MIT License (https://github.com/paragonie/constant_time_encoding/blob/198317fa6db951dd791be0740915dae878f34b3a/LICENSE.txt)
+ *
+ * Copyright (c) 2016 - 2022 Paragon Initiative Enterprises
+ */
+abstract class Base64
+{
```
azjezz:
```suggestion
 * Copyright (c) 2016 - 2022 Paragon Initiative Enterprises
 *
 * @internal
 */
abstract class Base64
{
```

src/Psl/Encoding/Base64/Internal/Base64DotSlash.php
```diff
@@ -0,0 +1,79 @@
...
+ * The following class was derived from code of Paragon Initiative Enterprises.
+ *
+ * https://github.com/paragonie/constant_time_encoding/blob/198317fa6db951dd791be0740915dae878f34b3a/src/Base64DotSlash.php
+ *
+ * Code subject to MIT License (https://github.com/paragonie/constant_time_encoding/blob/198317fa6db951dd791be0740915dae878f34b3a/LICENSE.txt)
+ *
+ * Copyright (c) 2016 - 2022 Paragon Initiative Enterprises
+ */
+final class Base64DotSlash extends Base64
+{
```
azjezz:
```suggestion
 * Copyright (c) 2016 - 2022 Paragon Initiative Enterprises
 *
 * @internal
 */
final class Base64DotSlash extends Base64
{
```

src/Psl/Encoding/Base64/Internal/Base64DotSlashOrdered.php
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * The following class was derived from code of Paragon Initiative Enterprises.
+ *
+ * https://github.com/paragonie/constant_time_encoding/blob/198317fa6db951dd791be0740915dae878f34b3a/src/Base64DotSlashOrdered.php
+ *
+ * Code subject to MIT License (https://github.com/paragonie/constant_time_encoding/blob/198317fa6db951dd791be0740915dae878f34b3a/LICENSE.txt)
+ *
+ * Copyright (c) 2016 - 2022 Paragon Initiative Enterprises
+ */
+final class Base64DotSlashOrdered extends Base64
```
azjezz:
```suggestion
 *
 * Copyright (c) 2016 - 2022 Paragon Initiative Enterprises
 *
 * @internal
 */
final class Base64DotSlashOrdered extends Base64
```

src/Psl/Encoding/Base64/Internal/Base64UrlSafe.php
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * The following class was derived from code of Paragon Initiative Enterprises.
+ *
+ * https://github.com/paragonie/constant_time_encoding/blob/198317fa6db951dd791be0740915dae878f34b3a/src/Base64UrlSafe.php
+ *
+ * Code subject to MIT License (https://github.com/paragonie/constant_time_encoding/blob/198317fa6db951dd791be0740915dae878f34b3a/LICENSE.txt)
+ *
+ * Copyright (c) 2016 - 2022 Paragon Initiative Enterprises
+ */
+final class Base64UrlSafe extends Base64
```
azjezz:
```suggestion
 *
 * Copyright (c) 2016 - 2022 Paragon Initiative Enterprises
 *
 * @internal
 */
final class Base64UrlSafe extends Base64
```

src/Psl/Encoding/Base64/Internal/Base64.php
```diff
@@ -0,0 +1,213 @@
...
+     * Base64 character set:
+     *  [A-Z]      [a-z]      [0-9]      +     /
+     *  0x41-0x5a, 0x61-0x7a, 0x30-0x39, 0x2b, 0x2f
+     *
+     * @pure
+     */
+    public static function encode(string $binary, bool $padding = true): string
+    {
+        $dest = '';
+        $binaryLength = Byte\length($binary);
```
azjezz: `Binary::safeStrlen` used in `paragonie/constant_time_encoding` refers to `mb_strlen($binary, '8bit')`, the equivalent in PSL would be `Str\length($binary, encoding: Str\Encoding::ASCII_8BIT);`, not `Byte\length($binary)`.

```suggestion
        $binaryLength = Str\length($binary, encoding: Str\Encoding::ASCII_8BIT);
```

( same applies to `Binary::safeSubstr` being `Str\slice($_, $_, $_, encoding: Str\Encoding::ASCII_8BIT)` )

azjezz (changes requested):

azjezz:
@veewee do you think this should start a new major branch ( 3.x ), or should we allow this BC break to go through?

we have before release a minor with a bc break in a non major component, so that won't be new :)

veewee:
@azjezz the way I see this, is that the public signature of the method did not change in a breaking way. The only thing that changed is extra **optional** arguments and the internal implementation.
I don't think we need a major got that.

Do we need to benchmark the implementation vs PHPs implementation to make sure it's similarly performant and produces the same results?

azjezz:
> I don't think we need a major got that.

It did though, we added new `Variant $variant` argument before the old `bool $pad` argument, while both are optional, something like `Base64\encode($data, true)` will fail now.

azjezz:
> Do we need to benchmark the implementation vs PHPs implementation to make sure it's similarly performant and produces the same results?

we can, but keep in mind that this implementation while slower ( not sure by how much ), provides constant time encoding/decoding, which the PHP implementation does not provide.

veewee:
> > I don't think we need a major got that.
> 
> It did though, we added new `Variant $variant` argument before the old `bool $pad` argument, while both are optional, something like `Base64\encode($data, true)` will fail now.

Previously we did not provide any padding options though. It was just the encoding/encoded arg

azjezz:
never mind me then :) my brain is not working right.

src/Psl/Encoding/Base64/Internal/Base64.php
```diff
@@ -0,0 +1,215 @@
...
+            $byte1 = $chunk[2];
+            $byte2 = $chunk[3];
+            $dest .= static::encode6Bits($byte0 >> 2) .
+                static::encode6Bits((($byte0 << 4) | ($byte1 >> 4)) & 63) .
+                static::encode6Bits((($byte1 << 2) | ($byte2 >> 6)) & 63) .
+                static::encode6Bits($byte2 & 63);
+        }
+        if ($i < $binaryLength) {
+            /** @var array<int, int> $chunk */
+            $chunk = unpack('C*', Str\slice($binary, $i, $binaryLength - $i, encoding: Str\Encoding::ASCII_8BIT));
```
azjezz:
```suggestion
            /** 
             * @psalm-suppress InvalidArgument
             * @var array<int, int> $chunk
             */
            $chunk = unpack('C*', Str\slice($binary, $i, $binaryLength - $i, encoding: Str\Encoding::ASCII_8BIT));
```

src/Psl/Encoding/Base64/Internal/Base64.php
```diff
@@ -0,0 +1,215 @@
...
+                'CCC',
+                ((($char0 << 2) | ($char1 >> 4)) & 0xff),
+                ((($char1 << 4) | ($char2 >> 2)) & 0xff),
+                ((($char2 << 6) | $char3) & 0xff),
+            );
+            $err |= ($char0 | $char1 | $char2 | $char3) >> 8;
+        }
+        if ($i < $base64Length) {
+            /** @var array<int, int> $chunk */
+            $chunk = unpack('C*', Str\slice($base64, $i, $base64Length - $i, encoding: Str\Encoding::ASCII_8BIT));
```
azjezz:
```suggestion
            /**
             * @psalm-suppress InvalidArgument
             * @var array<int, int> $chunk
             */
            $chunk = unpack('C*', Str\slice($base64, $i, $base64Length - $i, encoding: Str\Encoding::ASCII_8BIT));
```

src/Psl/Encoding/Base64/Variant.php
```diff
@@ -0,0 +1,33 @@
...
+     * 0x2e-0x2f, 0x41-0x5a, 0x61-0x7a, 0x30-0x39
+     */
+    case DotSlash;
+    /**
+     * Base64dotSlashOrdered character set:
+     * [.-9]      [A-Z]      [a-z]
+     * 0x2e-0x39, 0x41-0x5a, 0x61-0x7a
+     */
+    case DotSlashOrdered;
+}
```
azjezz:
```suggestion
{
    /**
     * Character set:
     *
     *  [A-Z]      [a-z]      [0-9]      +     /
     *  0x41-0x5a, 0x61-0x7a, 0x30-0x39, 0x2b, 0x2f
     */
    case Default;

    /**
     * Character set:
     *
     * [A-Z]      [a-z]      [0-9]      -     _
     * 0x41-0x5a, 0x61-0x7a, 0x30-0x39, 0x2d, 0x5f
     */
    case UrlSafe;

    /**
     * Character set:
     *
     * [A-Z]      [a-z]      [0-9]      .     /        
     * 0x41-0x5a, 0x61-0x7a, 0x30-0x39, 0x2e, 0x2f
     */
    case DotSlash;

    /**
     * Character set:
     *
     * [.-9]      [A-Z]      [a-z]
     * 0x2e-0x39, 0x41-0x5a, 0x61-0x7a
     */
    case DotSlashOrdered;
}
```

azjezz (changes requested):

src/Psl/Encoding/Base64/Internal/Base64.php
```diff
@@ -0,0 +1,215 @@
...
+        return pack('C', $bin + $diff);
+    }
+
+    /**
+     * Uses bitwise operators instead of table-lookups to turn 6-bit integers
+     * into 8-bit integers.
+     *
+     * Base64 character set:
+     * [A-Z]      [a-z]      [0-9]      +     /
+     * 0x41-0x5a, 0x61-0x7a, 0x30-0x39, 0x2b, 0x2f
```
azjezz:
```suggestion
```

repetitive, and not needed as this internal.

src/Psl/Encoding/Base64/Internal/Base64DotSlash.php
```diff
@@ -0,0 +1,81 @@
...
+        return pack('C', $bin);
+    }
+
+    /**
+     * Uses bitwise operators instead of table-lookups to turn 6-bit integers
+     * into 8-bit integers.
+     *
+     * Base64dotSlash character set:
+     * ./         [A-Z]      [a-z]     [0-9]
+     * 0x2e-0x2f, 0x41-0x5a, 0x61-0x7a, 0x30-0x39
```
azjezz:
```suggestion
```

same applies to other variants decoders.

src/Psl/Encoding/Base64/Internal/Base64.php
```diff
@@ -0,0 +1,215 @@
...
+            return '';
+        }
+
+        static::checkRange($base64);
+
+        if ($strictPadding && $base64Length % 4 !== 0) {
+            throw new Exception\IncorrectPaddingException('The given base64 string has incorrect padding.');
+        }
+        $base64 = Str\trim_right($base64, '=');
+        $base64Length = Str\length($base64, encoding: Str\Encoding::ASCII_8BIT);
```
azjezz:
```suggestion
        if ($explicit_padding && $base64_length % 4 !== 0) {
            throw new Exception\IncorrectPaddingException('The given base64 string has incorrect padding.');
        }
        $base64 = Str\trim_right($base64, '=');
        $base64_length = Str\length($base64, encoding: Str\Encoding::ASCII_8BIT);
```

`$snake_case` 🐍

azjezz:
@Gashmob aside from the comments above, this looks great to me 💪 we can ship it in the next minor soon.

src/Psl/Encoding/Base64/Internal/Base64.php
Gashmob: yes, I was not completely sure for that

Gashmob:
Ok, I've applied your suggestions and fix errors from checks. Normally all it's done, clean, good :smile:

azjezz:
awesome work @Gashmob !

thank you for contributing 🎉

PR #410, opened 2023-06-17
fix(str): add invariant to avoid unexpected errors when parsing an invalid UTF8 string

### LINKED ISSUE #409: Return value of `Psl\Str\trim()` must be of the type string, null returned
**Describe the bug**
Hi! We have recently caught the following error in production:
image

**To Reproduce**
I really tried to replicate it in order to add a test case, but I have not been able to replicate it as I don't have the complete data.

**Expected behavior**
It should return an empty string instead of a null. As `preg_replace` looks like it returns `array|string|null`, maybe a guard clause should work?

If you confirm that this could be a bug I can send a PR

devnix:
Fixes #409


devnix:
I took the approach of adding an invariant checking that the problematic parameter is a valid utf8 before doing anything else. 

Maybe an error handling around `preg_replace` would feel safer but also this approach felt cleaner and more efficient to my eyes, let me know about anything I can improve or rewrite!

src/Psl/Type/Internal/PositiveIntType.php
```diff
@@ -47,6 +47,7 @@ public function coerce(mixed $value): int
                 return $int;
             }
 
+            /** @psalm-suppress MissingThrowsDocblock */
```
veewee: This should result in a `CoerceException` instead of a `InvariantViolationException` I suppose

src/Psl/Str/trim.php
```diff
@@ -14,9 +16,13 @@
  * be stripped: space, tab, newline, carriage return, NUL byte, vertical tab.
  *
  * @pure
+ *
+ * @throws Psl\Exception\InvariantViolationException if $string is not a valid UTF-8 string.
  */
 function trim(string $string, ?string $char_mask = null): string
 {
+    Psl\invariant(is_utf8($string), 'Expected $string to be a valid UTF-8 string.');
```
veewee: I like this approach since it validates the input.
@azjezz  mentioned using `Psl\Regex` would introduce some overhead. 
Pretty sure this `mb_detect_encoding` will cause some sort of overhead as well though, since the code will now validate twice for valid UTF-8 encoding: once in `is_utf8()` - and once in `preg_replace` because it contains the `u` modifier.

Should we benchmark this, or assume we're ok with this downside?

veewee (commented):
Looks good, thanks! 
I've added a few small remarks

src/Psl/Str/trim.php
devnix: I was thinking too if it was a safer/faster option to catch the error, even if it is cumbersome.

src/Psl/Type/Internal/PositiveIntType.php
devnix: Nice catch, changing it!

veewee:
@devnix I took the liberty of performing some benchmarks here:

```php
#[Groups(['str'])]
final class Utf8Bench
{
    public function benchValidUtf8CheckThroughPreg(): void
    {
        preg_match('/./u', "HELLO");
    }

    public function benchValidUtf8CheckThroughEncodingAndPreg(): void
    {
        Psl\invariant(Str\is_utf8("HELLO"), 'Expected $string to be a valid UTF-8 string.');
        preg_match('/./u', "HELLO");
    }
}
```

The results are rather painful:

```
./vendor/bin/phpbench run --config config/phpbench.json --group=str --report=benchmark                                                                                                                  [13:37:38]
PHPBench (1.2.7) running benchmarks... #standwithukraine
with configuration file: /Users/verweto/Projects/github/azjezz/psl/config/phpbench.json
with PHP version 8.1.10, xdebug ✔, opcache ❌

\Psl\Tests\Benchmark\Str\Utf8Bench

    benchValidUtf8CheckThroughPreg..........R1 I2 - Mo0.700μs (±0.00%)
    benchValidUtf8CheckThroughEncodingAndPr.R4 I4 - Mo4.103μs (±1.18%)

Subjects: 2, Assertions: 0, Failures: 0, Errors: 0
Utf8Bench
=========

Average iteration times by variant

4.1μs     │   █
3.6μs     │   █
3.1μs     │   █
2.6μs     │   █
2.1μs     │   █
1.5μs     │   █
1.0μs     │ ▃ █
0.5μs     │ █ █
          └─────
            1 2

[█ <current>]

1: benchValidUtf8Check᠁ 2: benchValidUtf8Check᠁

Memory by variant

6.1mb     │ █ █
5.3mb     │ █ █
4.6mb     │ █ █
3.8mb     │ █ █
3.1mb     │ █ █
2.3mb     │ █ █
1.5mb     │ █ █
762.8kb   │ █ █
          └─────
            1 2

[█ <current>]

1: benchValidUtf8Check᠁ 2: benchValidUtf8Check᠁

<current>
+----------------------------------------------+---------+---------+--------+---------+
| subject                                      | memory  | mode    | rstdev | stdev   |
+----------------------------------------------+---------+---------+--------+---------+
| benchValidUtf8CheckThroughPreg ()            | 6.065mb | 0.700μs | ±0.00% | 0.000μs |
| benchValidUtf8CheckThroughEncodingAndPreg () | 6.102mb | 4.103μs | ±1.18% | 0.049μs |
+----------------------------------------------+---------+---------+--------+---------+
```

As you can see, this way is close to 6 times slower for `success`cases in comparison with dealing with the `null` return of `preg_x`.
It might not be the best of ideas - so I let @azjezz  decide on what to do with this :)

azjezz:
I think using is_utf8 is not the right solution indeed, we could either manually check for null return, or use `Psl\Regex` if it doesn't introduce as much of an overhead.

veewee:
@azjezz Psl\Regex is about 4.6x slower than regular `preg_`  - yet it is `4.600μs` so not really that big of a time :)
It seems like in this run, that the utf8 invariant is performing similarely to the Psl\Regex, but can't beat a simple null check.

```php
#[Groups(['str'])]
final class Utf8Bench
{
    public function benchValidUtf8CheckThroughPreg(): void
    {
        if (null === preg_replace('/./u', '', "HELLO")) {
            throw new \RuntimeException('xxx');
        }
    }

    public function benchValidUtf8CheckThroughEncodingAndPreg(): void
    {
        Psl\invariant(Str\is_utf8("HELLO"), 'Expected $string to be a valid UTF-8 string.');
        preg_replace('/./u', '', "HELLO");
    }

    public function benchValidUtfWithPslRegex(): void
    {
        Regex\replace('HELLO', '/./u', '');
    }
}

```

```
PHPBench (1.2.7) running benchmarks... #standwithukraine
with configuration file: /Users/verweto/Projects/github/azjezz/psl/config/phpbench.json
with PHP version 8.1.10, xdebug ✔, opcache ❌

\Psl\Tests\Benchmark\Str\Utf8Bench

    benchValidUtf8CheckThroughPreg..........R1 I3 - Mo1.000μs (±0.00%)
    benchValidUtf8CheckThroughEncodingAndPr.R1 I4 - Mo4.580μs (±2.58%)
    benchValidUtfWithPslRegex...............R1 I3 - Mo4.697μs (±1.05%)

Subjects: 3, Assertions: 0, Failures: 0, Errors: 0
Utf8Bench
=========

Average iteration times by variant

4.7μs     │   ▇ █
4.1μs     │   █ █
3.5μs     │   █ █
2.9μs     │   █ █
2.3μs     │   █ █
1.8μs     │   █ █
1.2μs     │ ▆ █ █
0.6μs     │ █ █ █
          └───────
            1 2 3

[█ <current>]

1: benchValidUtf8Check᠁ 2: benchValidUtf8Check᠁ 3: benchValidUtfWithPs᠁

Memory by variant

6.1mb     │ █ █ █
5.3mb     │ █ █ █
4.6mb     │ █ █ █
3.8mb     │ █ █ █
3.1mb     │ █ █ █
2.3mb     │ █ █ █
1.5mb     │ █ █ █
762.9kb   │ █ █ █
          └───────
            1 2 3

[█ <current>]

1: benchValidUtf8Check᠁ 2: benchValidUtf8Check᠁ 3: benchValidUtfWithPs᠁

<current>
+----------------------------------------------+---------+---------+--------+---------+
| subject                                      | memory  | mode    | rstdev | stdev   |
+----------------------------------------------+---------+---------+--------+---------+
| benchValidUtf8CheckThroughPreg ()            | 6.066mb | 1.000μs | ±0.00% | 0.000μs |
| benchValidUtf8CheckThroughEncodingAndPreg () | 6.103mb | 4.580μs | ±2.58% | 0.117μs |
| benchValidUtfWithPslRegex ()                 | 6.066mb | 4.697μs | ±1.05% | 0.049μs |
+----------------------------------------------+---------+---------+--------+---------+
```

azjezz:
Yep, i guess we just do a null check here then, we should try to minimise the performance difference between Psl and PHP stdlib, even though in this case, PHP does not have an alternative for these string functions ( at least not mb variants ).

src/Psl/Str/replace_ci.php
```diff
@@ -12,12 +14,16 @@
...
  */
 function replace_ci(string $haystack, string $needle, string $replacement, Encoding $encoding = Encoding::UTF_8): string
 {
     if ('' === $needle || null === search_ci($haystack, $needle, 0, $encoding)) {
         return $haystack;
     }
 
+    Psl\invariant(is_utf8($needle), 'Expected $needle to be a valid UTF-8 string.');
+
     return join(preg_split('{' . preg_quote($needle, '/') . '}iu', $haystack), $replacement);
```
devnix: Nice, I can easily rewrite the PR to null checks, but this function will output a warning: https://3v4l.org/citBH

Should we keep there the utf8 invariant, or we would wrap the call to `preg_split` in some way?

veewee: You might want to take a look at `Internal\suppress()` or `Internal\box()` - which are probably the best way to go.

But if I benchmark this, the fastest way is 'silencing' the error:

```php
#[Groups(['str'])]
final class Utf8Bench
{
    public function benchValidUtf8CheckThroughPreg(): void
    {
        if (null === preg_replace('/./u', '', "HELLO")) {
            throw new \RuntimeException('xxx');
        }
    }

    public function benchValidUtf8CheckThroughPregSuppressed(): void
    {
        if (null === Internal\suppress(static fn () => preg_replace('/./u', '', "HELLO"))) {
            throw new \RuntimeException('xxx');
        }
    }

    public function benchValidUtf8CheckThroughPregSilenced(): void
    {
        if (null === @preg_replace('/./u', '', "HELLO")) {
            throw new \RuntimeException('xxx');
        }
    }
}

```

```
\Psl\Tests\Benchmark\Str\Utf8Bench

    benchValidUtf8CheckThroughPreg..........I4 - Mo1.000μs (±0.00%)
    benchValidUtf8CheckThroughPregSuppresse.R2 I3 - Mo2.300μs (±1.72%)
    benchValidUtf8CheckThroughPregSilenced..R2 I4 - Mo0.800μs (±0.00%)

Subjects: 3, Assertions: 0, Failures: 0, Errors: 0
Utf8Bench
=========

Average iteration times by variant

2.3μs     │   █
2.0μs     │   █
1.7μs     │   █
1.4μs     │   █
1.2μs     │ ▄ █
0.9μs     │ █ █ ▇
0.6μs     │ █ █ █
0.3μs     │ █ █ █
          └───────
            1 2 3
<current>
+---------------------------------------------+---------+---------+--------+---------+
| subject                                     | memory  | mode    | rstdev | stdev   |
+---------------------------------------------+---------+---------+--------+---------+
| benchValidUtf8CheckThroughPreg ()           | 6.068mb | 1.000μs | ±0.00% | 0.000μs |
| benchValidUtf8CheckThroughPregSuppressed () | 6.068mb | 2.300μs | ±1.72% | 0.040μs |
| benchValidUtf8CheckThroughPregSilenced ()   | 6.068mb | 0.800μs | ±0.00% | 0.000μs |
+---------------------------------------------+---------+---------+--------+---------+
```

Not sure how strict @azjezz policy is on using silencing operators - since it will scream if `boolean xdebug.scream = true`

devnix: Probably the best tradeoff would be `Internal\suppress()`? I'll wait for @azjezz's opinion ^_^

azjezz: we can go with `Internal\suppress()`, the thing that bothers me about `@` is as you said, it would be ignored in case `xdebug,scream` is set to `true`, but other than that, i have no problem using `@` really here.

src/Psl/Str/replace_ci.php
```diff
@@ -12,12 +14,21 @@
...
         return $haystack;
     }
 
-    return join(preg_split('{' . preg_quote($needle, '/') . '}iu', $haystack), $replacement);
+    /** @psalm-suppress ImpureFunctionCall */
+    $pieces = Psl\Internal\suppress(
+        static fn () => preg_split('{' . preg_quote($needle, '/') . '}iu', $haystack)
+    );
+
+    Psl\invariant($pieces !== false, 'Expected $needle to be a valid UTF-8 string.');
```
veewee: got to love these little inconsistencies in PHP's core right?! :)

src/Psl/Str/replace_ci.php
```diff
@@ -12,12 +14,21 @@
...
  */
 function replace_ci(string $haystack, string $needle, string $replacement, Encoding $encoding = Encoding::UTF_8): string
 {
     if ('' === $needle || null === search_ci($haystack, $needle, 0, $encoding)) {
         return $haystack;
     }
 
-    return join(preg_split('{' . preg_quote($needle, '/') . '}iu', $haystack), $replacement);
+    /** @psalm-suppress ImpureFunctionCall */
+    $pieces = Psl\Internal\suppress(
```
veewee: I can't seem to figure out why you need to suppress:
https://3v4l.org/toK8g

I'm not getting any warnings from PHP (even with error reporting on `E_ALL`)

veewee (commented):
Looks good. Added a small question about the suppress here - but apart from that it looks to do what it needs to do :)

src/Psl/Str/replace_ci.php
devnix: I replicated it on this example https://3v4l.org/citBH

veewee: Ah yes I see, it's when the expression contains issues.
This also aplies to preg_replace though:
https://3v4l.org/bXP6u

So both the needle and/or haystack can fail in this case. 
The same goes for the trim functions : both the string and char_mask might fail.

I suggest adding suppress for every function and generalizing the error message so that either the needle and/or haystack may contain invalid chars?

devnix: > So both the needle and/or haystack can fail in this case.

Indeed, I had serious doubts about the invariant message. Even about the valid UTF-8 string. What if the function failed because of another reason? I could play with [`preg_last_error`](https://www.php.net/manual/es/function.preg-last-error.php) here

> I suggest adding suppress for every function and generalizing the error message so that either the needle and/or haystack may contain invalid chars?

As I tested it, no warnings appeared when throwing an invalid UTF-8. Anyway, it may be wise to silence all kind of error and use the same `preg_last_error` function. I'll work on it!

src/Psl/Str/trim.php
```diff
@@ -14,11 +16,17 @@
...
  */
 function trim(string $string, ?string $char_mask = null): string
 {
     $char_mask ??= " \t\n\r\0\x0B\x0C\u{A0}\u{FEFF}";
     $char_mask   = preg_quote($char_mask, null);
 
-    return preg_replace("{^[{$char_mask}]++|[{$char_mask}]++$}uD", '', $string);
+    $string = preg_replace("{^[{$char_mask}]++|[{$char_mask}]++$}uD", '', $string);
+
+    Psl\invariant(null !== $string, 'Expected $string to be a valid UTF-8 string.');
```
devnix: Silence and throw invariant depending on `preg_last_error`

src/Psl/Str/trim_left.php
```diff
@@ -14,11 +16,17 @@
...
  */
 function trim_left(string $string, ?string $char_mask = null): string
 {
     $char_mask ??= " \t\n\r\0\x0B\x0C\u{A0}\u{FEFF}";
     $char_mask   = preg_quote($char_mask, null);
 
-    return preg_replace("{^[{$char_mask}]++}uD", '', $string);
+    $string = preg_replace("{^[{$char_mask}]++}uD", '', $string);
+
+    Psl\invariant(null !== $string, 'Expected $string to be a valid UTF-8 string.');
```
devnix: Silence and throw invariant depending on `preg_last_error`

src/Psl/Str/trim_right.php
```diff
@@ -14,11 +16,17 @@
...
  */
 function trim_right(string $string, ?string $char_mask = null): string
 {
     $char_mask ??= " \t\n\r\0\x0B\x0C\u{A0}\u{FEFF}";
     $char_mask   = preg_quote($char_mask, null);
 
-    return preg_replace("{[{$char_mask}]++$}uD", '', $string);
+    $string = preg_replace("{[{$char_mask}]++$}uD", '', $string);
+
+    Psl\invariant(null !== $string, 'Expected $string to be a valid UTF-8 string.');
```
devnix: Silence and throw invariant depending on `preg_last_error`

src/Psl/Str/replace_ci.php
veewee: Well yeah, but then we're basically back to using the `Regex\x` functions  - because that is what they are doing internally :)

https://github.com/azjezz/psl/blob/next/src/Psl/Regex/Internal/call_preg.php

devnix: Well, after thinking about it, wouldn't be the same as using `Psl\Regex`, probably with the same overhead? 😕 

EDIT: sorry, your comment was not appearing while I was writing mine 😄 
Right now I can't find a faster/better solution. Probably `Psl\Regex` would be the best option with the cost of the overhead, as @azjezz said:
> [...] even though in this case, PHP does not have an alternative for these string functions ( at least not mb variants ).

devnix: Here's a funny thing I'm not sure how to solve, because I don't know if the problem is on the test's data set.

If I call `Regex\split`, the following assertion fails:
```php
function replace_ci(string $haystack, string $needle, string $replacement, Encoding $encoding = Encoding::UTF_8): string
{
    if ('' === $needle || null === search_ci($haystack, $needle, 0, $encoding)) {
        return $haystack;
    }

    return join(Regex\split($haystack, '{' . preg_quote($needle, '/') . '}iu'), $replacement);
}
```

```
1) Psl\Tests\Unit\Str\ReplaceCiTest::testReplaceCi with data set #2 ('مرحبا بكم', 'مرحبا سيف', 'سيف', 'بكم')
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'مرحبا بكم'
+'مرحبا '
```

As I'm debugging, I see that with the old `preg_split` returns the following array:

```php
array(2) {
  [0]=>
  string(11) "مرحبا "
  [1]=>
  string(0) ""
}
```

While calling `Psl\Regex` returns the following array, it looks like it is because of the `PREG_SPLIT_NO_EMPTY` flag:

```php
array(1) {
  [0]=>
  string(11) "مرحبا "
}
```

So in this case, if the dataset is correct, I will probably propose the following implementation:

```php
function replace_ci(string $haystack, string $needle, string $replacement, Encoding $encoding = Encoding::UTF_8): string
{
    if ('' === $needle || null === search_ci($haystack, $needle, 0, $encoding)) {
        return $haystack;
    }

    /** @var list<string> */
    $pieces = Regex\Internal\call_preg(
        'preg_split',
        static fn() => preg_split('{' . preg_quote($needle, '/') . '}iu', $haystack, $limit ?? -1),
    );

    return join($pieces, $replacement);
}
```

WDYT @veewee  @azjezz?

veewee: I don't know this language either. I'll keep it up to @azjezz

veewee (approved):
Using the Regex component is probably the best option - even though it's less performant than other options.

src/Psl/Str/replace_ci.php
azjezz: sorry for the delayed response, the test is correct, and should pass.

azjezz (approved):

PR #415, opened 2023-07-17
fix(option): return `Option<never>` for `Option::none()`

devnix:
This change would make PHPStan generics work with these kinds of examples:

```php
/**
 * @return Option<string>
 */
public function test(): Option
{
    return none();
}
```

Which right now throws an error like
```
Method Test::test() should return Psl\Option\Option<string> but returns Psl\Option\Option<mixed>. 
```

### DIFF
--- a/docs/component/option.md
+++ b/docs/component/option.md
@@ -13,7 +13,7 @@
 #### `Functions`
 
 - [from_nullable](./../../src/Psl/Option/from_nullable.php#L16)
-- [none](./../../src/Psl/Option/none.php#L14)
+- [none](./../../src/Psl/Option/none.php#L12)
 - [some](./../../src/Psl/Option/some.php#L16)
 
 #### `Classes`
--- a/src/Psl/Option/Option.php
+++ b/src/Psl/Option/Option.php
@@ -38,12 +38,11 @@ final class Option
     /**
      * Create an option with none value.
      *
-     * @template Tn
-     *
-     * @return Option<Tn>
+     * @return Option<never>
      */
     public static function none(): Option
     {
+        /** @var Option<never> */
         return new self(null);
     }
 
--- a/src/Psl/Option/none.php
+++ b/src/Psl/Option/none.php
@@ -7,9 +7,7 @@ namespace Psl\Option;
 /**
  * Create an option with none value.
  *
- * @template T
- *
- * @return Option<T>
+ * @return Option<never>
  */
 function none(): Option
 {


azjezz (approved):

azjezz:
I'm okay with this change, but i fear it might be a BC break.

wdyt @veewee?

veewee:
I don't think this is a breaking change because of:
https://psalm.dev/r/9715a63a4f

If you create an union of `T | never`, it seems to infer to `T` internally.
Validated this change on my projects who use `Option` quite intensively and got 0 errors reported by psalm either.

However, I do think we also need to add this never return to the regular `none()` function:

```php
namespace Psl\Option;

/**
 * Create an option with none value.
 *
 * @return Option<never>
 */
function none(): Option
{
    return Option::none();
}

```

devnix:
It _may_ be a breaking change in some cases.

```php
/**
 * @return Option<string>
 */
public function test(): Option
{
    /** @var Option<string> */ 
    return none();
}
```

I'm not currently using Psalm so I'm unsure if this is a linting error or just a feature to remove "dead" code that does not belong to the static analysis part, but there is a redundant `@var` now: https://github.com/vimeo/psalm/issues/2150 

```php
/**
 * @return Option<string>
 */
public function test(): Option
{
    /** @phpstan-ignore-next-line */ 
    return none();
}
```

In this example, PHPStan would throw an `No error to ignore is reported [...]` .

In both cases, I'm not sure if I would consider these BC changes in static analysis, or just regular fixes, but it will change a couple of baselines too.

Edit: 
> However, I do think we also need to add this never return to the regular none() function:

I think it would not be necessary, at least in PHPStan. Maybe Psalm would require it, idk

veewee:
In psalm, you have to define it because of this:
https://psalm.dev/r/068ab0cbf1

If you don't specify the template, it will take `mixed`. 
Maybe phpstan infers it from the caller instead of the callee or maybe it defaults to never?

I've reported an issue before in psalm regarding this 'optional' generic:
https://github.com/vimeo/psalm/issues/8487
To me, it also makes more sense to default to never instead of mixed.

You might be able to use `@psalm-var` instead of `@var`?
https://psalm.dev/docs/annotating_code/supported_annotations/#psalm-var-psalm-param-psalm-return-psalm-property-psalm-property-read-psalm-property-write-psalm-method

> However, I do think we also need to add this never return to the regular none() function:
> I think it would not be necessary, at least in PHPStan. Maybe Psalm would require it, idk

It would rather be for consistency than for fixing a SAT complaint. If `Option::none()` returns `Option<never>`, then `none()` should also return `Option<never>`.
This should work just fine in your SAT:

https://psalm.dev/r/c6ab6659f3

devnix:
You are right @veewee, the `Option<never>` in `Psl\Option\none` is necessary in PHPStan! Updating the PR.

I'm not scanning your project with PHPStan as I guess you are not trying to be PHPStan compliant. If you think it can be interesting to pass both static analysis, I could start a draft if you are ok with that 😊

azjezz:
@devnix would welcome that, there's no harm in having it here too.

PR #423, opened 2023-09-28
Bring over perf improvement from hhvm/hsl

lexidor:
X-Ref: https://github.com/hhvm/hsl/commit/61c4b4346e25d7f80814c1f004067a568b0b4a03

### DIFF
--- a/src/Psl/SecureRandom/float.php
+++ b/src/Psl/SecureRandom/float.php
@@ -15,10 +15,5 @@ use Psl\Math;
  */
 function float(): float
 {
-    /**
-     * @psalm-suppress MissingThrowsDocblock
-     */
-    $result = namespace\int(0, Math\INT53_MAX) / Math\INT53_MAX;
-
-    return (float) $result;
+    return (float) namespace\int(0, Math\INT53_MAX) / Math\INT53_MAX;
 }


azjezz:
this doesn't bring any performance improvement, PHP opcache will align the variable later on. the reason for using a separate variable here is to inform psalm that the exceptions thrown `int` function won't be thrown in this particular case.

lexidor:
Sorry azjezz, you've missed the reason for this change. The engineers at Meta weren't concerned about an intermediary variable.

Doing the float cast before the division makes this an fdiv, which doesn't check for clean (without remainder) division first. By having two integer operands, this division is performed twice. Once in the integral domain, to determine if the remainder would be zero. And if the remainder is not zero, the division is performed again in the floating point domain.

Since the only clean divisions in the range (0..2^53) over 2^53 are when the first operand are zero or one, this double division is very unlikely to result in a clean division.

I should not have removed the suppression comment and the intermediate variable. I assumed it was to indicate that a division by zero exception was impossible. My removal of the intermediate variable has cluttered the reason for this change.

If PHP's JIT is able to reason that performing an fdiv here is semantically equivalent, hats off for PHP and it's JIT. I naively assumed that, if Facebook didn't optimize the JIT for this, neither would the PHP Group.

PR #425, opened 2023-10-18
feat(option): new `Option::merge()` method

devnix:
The idea, as shown, is easy.

I propose a versatile `merge` method to make several operations easier between two `Option`s:

```php
Option\some(1)->merge(Option\some(2), static fn (int $a, int $b) => $a + $b)); // Option<3>
Option\none()->merge(Option\some(2), static fn (int $a, int $b) => $a + $b)); // Option<*never*>
Option\some(1)->merge(Option\none(), static fn (int $a, int $b) => $a + $b)); // Option<*never*>

Option\some('Hello')->merge(
    Option\some('World'), 
    static fn (string $a, string $b) => $a.' '.$b)
); // Option<"Hello World">
```

This is what triggered this particular need in my case:

```php
$value1 = Option\some(false);
$value2 = Option\some(true);

$value1->merge($value2, static fn (bool $a, bool $b) => $a && $b)->unwrapOr(false); // false

$value1 = Option\some(Money::EUR(100));
$value2 = Option\some(Money::USD(100));

$value1->merge($value2, static fn (Money $a, Money $b) => $a->equals($b))->unwrapOr(false); // false
```

Probably there is a better name for this method, for sure!

### DIFF
--- a/src/Psl/Option/Option.php
+++ b/src/Psl/Option/Option.php
@@ -279,4 +279,23 @@ final class Option
 
         return some($default());
     }
+
+    /**
+     * Merges an `Option<T>` and other `Option<T>` to `Option<Tu>` by applying a function to both contained values.
+     *
+     * @template Tu
+     *
+     * @param Option<T> $other
+     * @param (Closure(T, T): Tu) $closure
+     *
+     * @return Option<Tu>
+     */
+    public function merge(Option $other, Closure $closure)
+    {
+        if ($this->option !== null && $other->option !== null) {
+            return some($closure($this->option[0], $other->option[0]));
+        }
+
+        return none();
+    }
 }
--- a/tests/unit/Option/NoneTest.php
+++ b/tests/unit/Option/NoneTest.php
@@ -108,4 +108,25 @@ final class NoneTest extends TestCase
 
         static::assertNull($option->andThen(static fn($i) => Option\some($i + 1))->unwrapOr(null));
     }
+
+    /**
+     * @param Option\Option<int> $value1
+     * @param Option\Option<int> $value2
+     *
+     * @dataProvider provideMerge
+     */
+    public function testMerge(Option\Option $value1, Option\Option $value2): void
+    {
+        $this->expectException(NoneException::class);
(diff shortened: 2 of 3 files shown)


devnix:
I'm unsure about those mutation test errors. Any change in my branch does not cause them, am I right?

veewee:
Looks like a nice feature. Haven't jumped in the code yet, but was thinking about the name in line of `$option->combineWith($other, $closure)`. WDYT?

devnix:
I was thinking along the lines of using a short name, `join` was a candidate I was thinking of too. To me, `and`/`andThen` was the nicer options here

veewee:
> I'm unsure about those mutation test errors. Any change in my branch does not cause them, am I right?

I was looking at the mutation errors in #427. Gonna continue somewhere next week most likely.

veewee:
@devnix  can you rebase against the `next` branch to get the github actions green again?
I was thinking about naming it `compose()` this morning. What do you think about that one?

devnix:
Hi @veewee! I've been playing with Rust for the last couple of weeks, and I've found that `Option` has a [`zip`](https://doc.rust-lang.org/std/option/enum.Option.html#method.zip) method that may be useful for this purpose. Instead of requiring a callback to join two `Some` values into one like my idea, it creates a tuple if both are `Some`. Check this example: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bf6a95e1e06b98093c114a274a6e0fe6

My original proposal would be something like `zip_with` I guess, which is currently experimental, and equivalent to `.zip().map()`. I could provide the implementation of both if you are ok with that, but I'm unsure about how to represent this tuple in php/phpdoc. Also, maybe the `unzip` method would be interesting too?

veewee:
Just looked at the rust documentation and that indeed seems the way to go and the naming to use here. Feel free to provide all 3 of those functions.
Type-hinting tuples can be done with `array{0: A, 1: B}` iirc. 

https://psalm.dev/r/06049ede28

devnix:
Superseded by https://github.com/azjezz/psl/pull/434

PR #426, opened 2023-10-18
feat(option): new `Option::apply()` method

devnix:
I guess it's not an important method because you can return the same value with `map`, but I'll leave the proposal 😄

### DIFF
--- a/tests/unit/Option/NoneTest.php
+++ b/tests/unit/Option/NoneTest.php
@@ -101,12 +101,12 @@ final class NoneTest extends TestCase
         $spy = new Ref(1);
 
         $option = Option\none();
-        $option->apply(static function (int $value) use ($spy) {
+        $actual = $option->apply(static function (int $value) use ($spy) {
             $spy->value += $value;
         });
 
         static::assertSame(1, $spy->value);
-        static::assertTrue($option->isNone());
+        static::assertSame($option, $actual);
     }
 
     public function testMap(): void
--- a/tests/unit/Option/NoneTest.php
+++ b/tests/unit/Option/NoneTest.php
@@ -10,8 +10,8 @@ use Psl\Comparison\Equable;
 use Psl\Comparison\Order;
 use Psl\Option;
 use Psl\Option\Exception\NoneException;
-use Psl\Str;
 use Psl\Ref;
+use Psl\Str;
 
 final class NoneTest extends TestCase
 {
--- a/tests/unit/Option/NoneTest.php
+++ b/tests/unit/Option/NoneTest.php
@@ -95,6 +95,18 @@ final class NoneTest extends TestCase
         static::assertSame('There is no value', $result);
     }
 
+    public function testApply(): void
+    {
+        $i = 1;
+
+        $option = Option\none();
+        $option->apply(static function (int $value) use (&$i) {
+            $i += $value;
+        });
+
+        static::assertSame(1, $i);
+    }
+
     public function testMap(): void
     {
         $option = Option\none();
--- a/tests/unit/Option/NoneTest.php
+++ b/tests/unit/Option/NoneTest.php
@@ -11,6 +11,7 @@ use Psl\Comparison\Order;
 use Psl\Option;
 use Psl\Option\Exception\NoneException;
 use Psl\Str;
+use Psl\Ref;
 
 final class NoneTest extends TestCase
 {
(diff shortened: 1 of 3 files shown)


veewee:
What would be the big benifit of this over using `map()`? 
Would you consume it and chain it back into doing another operation? 

I don't see how I'dd use it personally at this moment, but I'm open for opinions :)

devnix:
Just syntactic sugar for operations if it holds `some`, but avoiding worrying about the verbosity of returning the same value if you want to chain methods.

Code like this (probably not the best example)
```php
$stock
    ->map(function (int $value) use ($logger) {
        $logger->info('New stock requested', ['stock' => $value]);
        return $value;
    })
    ->map(Stock::create(...))
    ->map(function (Stock $value) use ($product) {
        $product->updateStock($value);
        return $value;
    });
```

could be much more readable, and probably more expressive about intention:

```php
$stock
    ->apply(fn (int $value) => $logger->info('New stock requested', ['stock' => $value]))
    ->map(Stock::create(...))
    ->apply(fn (Stock $value) => $product->updateStock($stock));
```

veewee:
@devnix can you rebase against the next branch to get the github actions green again and add some tests?
It could make sense for people to have this option.

tests/unit/Option/NoneTest.php
```diff
@@ -84,6 +84,18 @@ public function testContains(): void
         static::assertFalse($option->contains(4));
     }
 
+    public function testApply(): void
+    {
+        $i = 1;
```
veewee: Can you use `\Psl\Ref` instead of the PHP reference here?
(it's designed to do just that)

tests/unit/Option/NoneTest.php
```diff
@@ -84,6 +84,18 @@ public function testContains(): void
...
+
+        $option = Option\none();
+        $option->apply(static function (int $value) use (&$i) {
+            $i += $value;
+        });
+
+        static::assertSame(1, $i);
```
veewee: can you add a check that the result of `apply()` is the same as `$option`?

tests/unit/Option/SomeTest.php
```diff
@@ -81,6 +81,18 @@ public function testContains(): void
         static::assertTrue($option->contains(2));
     }
 
+    public function testApply(): void
+    {
+        $i = 1;
```
veewee: (same comments here)

veewee (changes requested):
Looks good thanks. I've added a small couple of nitpicks to the code review.
After those it should be good to go!

tests/unit/Option/NoneTest.php
```diff
@@ -84,6 +85,19 @@ public function testContains(): void
         static::assertFalse($option->contains(4));
     }
 
+    public function testApply(): void
+    {
+        $spy = new Ref(1);
+
+        $option = Option\none();
+        $option->apply(static function (int $value) use ($spy) {
```
veewee: You are not validating the returned result at the moment:

```php
$actual = $option->apply(...);

static::assertSame($option, $actual);
```

devnix: I would say that this line would be the same assertion if I'm not mistaken: https://github.com/azjezz/psl/pull/426/files/89ed717b7d6801c77710f697b54a25ea1a0f959d#diff-1b95ed4d6bb75bfe486ce416a6a87705dc3dfffa381712a06f065d0425c9c749R98

Do you want it to be replaced by your suggestion?

EDIT: Sorry, forget it, the `$option` is immutable 🤦‍♂️

veewee: No problem :)
The some() version requires a similar approach though.

tests/unit/Option/SomeTest.php
```diff
@@ -81,6 +82,19 @@ public function testContains(): void
         static::assertTrue($option->contains(2));
     }
 
+    public function testApply(): void
+    {
+        $spy = new Ref(1);
+
+        $option = Option\some(2);
+        $option->apply(static function (int $value) use ($spy) {
```
veewee: can you add the `actual` check here as well?

devnix: Not 100% sure if I understood what you mean, I've moved those `2` values to a variable called `$actual`

veewee: No, I meant checking the return value of apply here as well. As how you did it in the None test.

devnix: Sorry, gotcha! 👍🏻

devnix:
@veewee I think this should be ready too, conflicts are resolved

veewee (approved):

veewee:
Cool, Thanks again!

PR #427, opened 2023-10-22
Fix mutations + math float tests

veewee:
This PR fixes:

* Some mutation errors that are popping up
* Float related errors in PHPunit that are bugging me on mac for some while

(See #425)

### DIFF
--- a/config/infection.json.dist
+++ b/config/infection.json.dist
@@ -21,6 +21,48 @@
         }
     },
     "timeout": 10,
-    "minMsi": 98,
-    "minCoveredMsi": 100
+    "minMsi": 100,
+    "minCoveredMsi": 100,
+    "mutators": {
+        "@default": true,
+        "Break_": {
+            "ignore": [
+                "Psl\\Collection\\*Map::zip"
+            ]
+        },
+        "DecrementInteger": {
+            "ignore": [
+                "Psl\\DataStructure\\PriorityQueue::peek"
+            ]
+        },
+        "FunctionCallRemoval": {
+            "ignore": [
+                "Psl\\Result\\Success::getThrowable"
+            ]
+        },
+        "IncrementInteger": {
+            "ignore": [
+                "Psl\\DataStructure\\PriorityQueue::peek"
+            ]
+        },
+        "LogicalNot": {
+            "ignore": [
+                "Psl\\Hash\\Context::update"
+            ]
+        },
+        "MethodCallRemoval": {
+            "ignore": [
+                "Psl\\Iter\\Iterator::seek",
+                "Psl\\Iter\\Iterator::count"
+            ]
+        },
+        "Throw_": {
+            "ignore": [
+                "Psl\\File\\ReadHandle::__construct",
+                "Psl\\File\\WriteHandle::__construct",
+                "Psl\\File\\ReadWriteHandle::__construct",
+                "Psl\\Hash\\Context::update"
+            ]
+        }
+    }
 }
--- a/tests/unit/Math/TanTest.php
+++ b/tests/unit/Math/TanTest.php
@@ -9,12 +9,14 @@ use Psl\Math;
 
 final class TanTest extends TestCase
 {
+    use FloatAsserts;
+
     /**
      * @dataProvider provideData
      */
-    public function testTan(float $expected, float $number): void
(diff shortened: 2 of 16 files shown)


config/infection.json.dist
```diff
@@ -21,6 +21,48 @@
...
     "timeout": 10,
-    "minMsi": 98,
-    "minCoveredMsi": 100
+    "minMsi": 100,
+    "minCoveredMsi": 100,
+    "mutators": {
+        "@default": true,
+        "Break_": {
+            "ignore": [
+                "Psl\\Collection\\*Map::zip"
```
veewee, line 30: When the break gets transformed into a continue, it will produce the same output - yet slower.
So this is a false positive:

https://github.com/azjezz/psl/blob/4affb62545080af0457f78a2bdcfc33cc4057f47/src/Psl/Collection/Map.php#L378-L380

config/infection.json.dist
```diff
@@ -21,6 +21,48 @@
...
+            ]
+        },
+        "DecrementInteger": {
+            "ignore": [
+                "Psl\\DataStructure\\PriorityQueue::peek"
```
veewee, line 35: This fallback to `?? 0` will never occur since we first check the array size.
This is a false positive:

https://github.com/azjezz/psl/blob/4affb62545080af0457f78a2bdcfc33cc4057f47/src/Psl/DataStructure/PriorityQueue.php#L53

I could solve it by doing an additional `invariant` check, but this seems slower for no real reason.

config/infection.json.dist
```diff
@@ -21,6 +21,48 @@
...
+            ]
+        },
+        "FunctionCallRemoval": {
+            "ignore": [
+                "Psl\\Result\\Success::getThrowable"
```
veewee, line 40: This one is actually covered - yet infection doesnt seem to be able to link the test to the mutation.

See
https://github.com/azjezz/psl/blob/4affb62545080af0457f78a2bdcfc33cc4057f47/tests/unit/Result/SuccessTest.php#L34-L42

config/infection.json.dist
```diff
@@ -21,6 +21,48 @@
...
+            ]
+        },
+        "IncrementInteger": {
+            "ignore": [
+                "Psl\\DataStructure\\PriorityQueue::peek"
+            ]
+        },
+        "LogicalNot": {
+            "ignore": [
+                "Psl\\Hash\\Context::update"
```
veewee, line 50: Hard to cover this one:

https://github.com/azjezz/psl/blob/4affb62545080af0457f78a2bdcfc33cc4057f47/src/Psl/Hash/Context.php#L74-L79

(the code coverage was already added as well.)

config/infection.json.dist
```diff
@@ -21,6 +21,48 @@
...
+            ]
+        },
+        "MethodCallRemoval": {
+            "ignore": [
+                "Psl\\Iter\\Iterator::seek",
```
veewee, line 55: These functions were added to speed up the tests : it results in time-outs cause the iterator does not iterate anymore because it removes:

https://github.com/azjezz/psl/blob/4affb62545080af0457f78a2bdcfc33cc4057f47/src/Psl/Iter/Iterator.php#L181

config/infection.json.dist
```diff
@@ -21,6 +21,48 @@
...
+                "Psl\\Iter\\Iterator::count"
+            ]
+        },
+        "Throw_": {
+            "ignore": [
+                "Psl\\File\\ReadHandle::__construct",
```
veewee, line 61: I've added some tests as well - but didnt want to go into chmodding files and testing for readability / writability at this moment. 

See
https://github.com/azjezz/psl/blob/4affb62545080af0457f78a2bdcfc33cc4057f47/src/Psl/File/ReadHandle.php#L25-L35

tests/unit/Math/TanTest.php
```diff
@@ -9,20 +9,23 @@
...
     }
 
     public function provideData(): array
     {
         return [
             [
                 -3.380515006246586,
-                5.0
+                5.0,
+                0.00000000000001
```
veewee, line 28: it is off by one on my system for some odd reason. Therefor I decided to check only the first 14 digits

Issue #412, opened 2023-06-27
Improve error messages for shape and vec -like types


veewee:
Given:

```php
use Psl\Type;
$shape = Type\shape([
    'name' => Type\string(),
    'articles' => Type\vec(Type\shape([
        'title' => Type\string(),
        'content' => Type\string(),
        'likes' => Type\int(),
        'comments' => Type\optional(Type\vec(Type\shape([
            'user' => Type\string(),
            'comment' => Type\string()
        ]))),
    ]))
]);

try {
    $shape->coerce([
        'name' => 'ok',
        'articles' => [
            [
                'title' => 'ok',
                'content' => new stdClass,
            ]
        ]
    ]);
} catch (Type\Exception\CoercionException $e) {
    echo $e->getMessage().PHP_EOL;
    var_dump($e->getTypeTrace()->getFrames());
}

```

This results in:

```php
Could not coerce "stdClass" to type "string".

array(3) {
  [0] =>
  string(20) "array{'articles': _}"
  [1] =>
  string(120) "vec<array{'title': string, 'content': string, 'likes': int, 'comments'?: vec<array{'user': string, 'comment': string}>}>"
  [2] =>
  string(19) "array{'content': _}"
}
```

Both the path and frames are rather confusing and it is unclear what failed exactly without processing this information.

Suggestion:

Keep track of an additional `path` on the `TypeTrace`so that we can build errors like:

> Could not coerce "stdClass" to type "string" at path articles.0.content

The `path` can contain a list of strings to identify the path - which will be set for shape and vec-like types:

* Shapes will mark the property: `prop`
* Vecs will mark the index: `0`
* Dicts will mark the named index: `key`

The list of paths will be `.`-separated

❗Consider the performance/memory impact, because PSL assertions are already quite runtime-heavy

PR #429, opened 2023-10-27
Improve error messages for shape and vec -like types

vrielsa:
Given:
```php
use Psl\Type;
$shape = Type\shape([
    'name' => Type\string(),
    'articles' => Type\vec(Type\shape([
        'title' => Type\string(),
        'content' => Type\string(),
        'likes' => Type\int(),
        'comments' => Type\optional(Type\vec(Type\shape([
            'user' => Type\string(),
            'comment' => Type\string()
        ]))),
    ])),
    'dictionary' => Type\dict(Type\string(), Type\vec(Type\shape([
        'title' => Type\string(),
        'content' => Type\string(),
    ]))),
    'pagination' => Type\Shape([
        'currentPage' => Type\uint(),
        'totalPages' => Type\uint(),
        'perPage' => Type\uint(),
        'totalRows' => Type\uint(),
    ])
]);

try {
    $shape->coerce([
        'name' => 'ok',
        'articles' => [
            [
                'title' => 'ok',
                'content' => 'ok',
                'likes' => 1,
                'comments' => [
                    [
                        'user' => 'ok',
                        'comment' => 'ok'
                    ],
                    [
                        'user' => 'ok',
                        'comment' =>'ok',
                    ]
                ]
            ]
        ],
       'dictionary' => [
           'key' => [
               [
                   'title' => 'ok',
                   'content' => 'ok',
               ]
           ]
       ]
    ]);
} catch (Type\Exception\CoercionException $e) {
    echo $e->getMessage().PHP_EOL;
}
``` 
Results in:
```php
Could not coerce "array" to type "array{'name': string, 'articles': vec<array{'title': string, 'content': string, 'likes': int, 'comments'?: vec<array{'user': string, 'comment': string}>}>, 'dictionary': dict<string, vec<array{'title': string, 'content': string}>>, 'pagination': array{'currentPage': uint, 'totalPages': uint, 'perPage': uint, 'totalRows': uint}}": at path pagination
``` 

The changed I made do have quite an impact on performance, but tried to keep it at a minimum. Mainly vecs and dicts are suffering. Not sure what the accepted range is.
image

image

The impact on shapes are minimal:
image

(See #412)

### DIFF
--- a/src/Psl/Type/Internal/DictType.php
+++ b/src/Psl/Type/Internal/DictType.php
@@ -44,8 +44,6 @@ final class DictType extends Type\Type
 
         $trace = $this->getTrace();
         $key_type = $this->key_type->withTrace($trace->withFrame('dict<' . $this->key_type->toString() . ', _>'));
-        $value_type = $this->value_type->withTrace($trace->withFrame('dict<_, ' . $this->value_type->toString() . '>'));
-
         $result = [];
 
         /**
@@ -53,6 +51,12 @@ final class DictType extends Type\Type
          * @var Tv $v
          */
         foreach ($value as $k => $v) {
+            $value_type = $this->value_type->withTrace(
+                $trace->withFrameAtPath(
+                    'dict<_, ' . $this->value_type->toString() . '>',
+                    (string) $k
+                )
+            );
             $result[$key_type->coerce($k)] = $value_type->coerce($v);
         }
 
--- a/src/Psl/Type/Internal/VecType.php
+++ b/src/Psl/Type/Internal/VecType.php
@@ -58,19 +58,21 @@ final class VecType extends Type\Type
             throw CoercionException::withValue($value, $this->toString(), $this->getTrace());
         }
 
-        /** @var Type\Type<Tv> $value_type */
-        $value_type = $this->value_type->withTrace(
-            $this->getTrace()
-                ->withFrame($this->toString())
-        );
-
         /**
          * @var list<Tv> $entries
          */
         $result = [];
 
-        /** @var Tv $v */
-        foreach ($value as $v) {
+        /**
+         * @var array-key $i
+         * @var Tv $v
+         */
+        foreach ($value as $i => $v) {
+            /** @var Type\Type<Tv> $value_type */
+            $value_type = $this->value_type->withTrace(
+                $this->getTrace()
+                    ->withFrameAtPath($this->toString(), (string) $i)
+            );
             $result[] = $value_type->coerce($v);
         }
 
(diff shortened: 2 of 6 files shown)


src/Psl/Type/Internal/DictType.php
```diff
@@ -44,15 +44,19 @@ public function coerce(mixed $value): array
...
         $result = [];
 
         /**
          * @var Tk $k
          * @var Tv $v
          */
         foreach ($value as $k => $v) {
+            $value_type = $this->value_type->withTrace(
+                $trace->withFrameAtPath(
+                    'dict<_, ' . $this->value_type->toString() . '>',
```
veewee, line 56: string concatenations are fast, but maybe it helps to build the frame name outside of the loop and use it instead of always recalculating the frame name?
It will always be the same value.

src/Psl/Type/Internal/VecType.php
```diff
@@ -58,19 +58,21 @@ public function coerce(mixed $value): iterable
...
-        foreach ($value as $v) {
+        /**
+         * @var array-key $i
+         * @var Tv $v
+         */
+        foreach ($value as $i => $v) {
+            /** @var Type\Type<Tv> $value_type */
+            $value_type = $this->value_type->withTrace(
+                $this->getTrace()
+                    ->withFrameAtPath($this->toString(), (string) $i)
```
veewee, line 74: Also here It might help to build the frame name outside of the loop.

tests/unit/Type/Exception/TypeCoercionExceptionTest.php
```diff
@@ -94,4 +94,29 @@ public function testConversionFailure(): void
             static::assertCount(0, $frames);
         }
     }
+
+    public function testCoercionPath(): void
```
veewee, line 98: It's a start, but we'll need tests for every changed type individually.

veewee (commented):
I'm +1 for this change!

To me, It is (at first feeling acceptibally) slower but at least it will give you helpful error messages.
Not sure how other people see this.

Some additional comments on the code:

* You only cover `coerce()`. We could have this on `assert()` as well?
* You currently don't cover paths for the types: `map`, `mutable_map`, `vector`, `mutable_vector`, `non_empty_vec`, `non_empty_dict`, `mixed_dict`, `mixed_vec`

veewee:
This might become a bit too slow on big data structures.
An alternative approach could be to re-run the assertions / coercions and collect that more advanced path information on failure only.

This issue needs some more thoughts. I'm gonna keep it open for now.

PR #433, opened 2023-12-11
feat(option): add `Option::proceed()` method

devnix:
This is another idea of implementing a `match` method that I had after learning more bits of Rust: https://doc.rust-lang.org/rust-by-example/std/option.html

WDYT?

### DIFF
--- a/tests/unit/Option/NoneTest.php
+++ b/tests/unit/Option/NoneTest.php
@@ -10,6 +10,7 @@ use Psl\Comparison\Equable;
 use Psl\Comparison\Order;
 use Psl\Option;
 use Psl\Option\Exception\NoneException;
+use Psl\Str;
 
 final class NoneTest extends TestCase
 {
@@ -86,22 +87,12 @@ final class NoneTest extends TestCase
 
     public function testProceed(): void
     {
-        $checked_divisor = static function (int $dividend, int $divisor): Option\Option {
-            if ($divisor === 0) {
-                return Option\none();
-            }
+        $result = Option\none()->proceed(
+            static fn ($i) => Str\format('Value is %d', $i),
+            static fn () => 'There is no value',
+        );
 
-            return Option\some($dividend / $divisor);
-        };
-
-        $try_division = static function (int $dividend, int $divisor) use ($checked_divisor): string {
-            return $checked_divisor($dividend, $divisor)->proceed(
-                some: static fn ($value) => sprintf('%s / %s = %s', $dividend, $divisor, $value),
-                none: static fn() => sprintf('%s / %s failed!', $dividend, $divisor),
-            );
-        };
-
-        static::assertSame('2 / 0 failed!', $try_division(2, 0));
+        static::assertSame('There is no value', $result);
     }
 
     public function testMap(): void
--- a/tests/unit/Option/NoneTest.php
+++ b/tests/unit/Option/NoneTest.php
@@ -84,7 +84,7 @@ final class NoneTest extends TestCase
         static::assertFalse($option->contains(4));
     }
 
-    public function testMatch(): void
+    public function testProceed(): void
     {
         $checked_divisor = static function (int $dividend, int $divisor): Option\Option {
             if ($divisor === 0) {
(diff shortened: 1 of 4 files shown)


veewee:
Hello,

Thanks for the PR.

For Result, we have `proceed`, which acts similar to what you want to achieve here in option:

https://github.com/azjezz/psl/blob/0aa7117b2e281e2fd855f41976bca4b59b1877f0/src/Psl/Result/ResultInterface.php#L132-L145

I've mentioned something similar in the original PR:

https://github.com/azjezz/psl/pull/356/files#r912345716


The conclusion in there was this one:

The same thing can be achieved by using `mapOrElse()` in combination with `unwrap()`.
For your example implementation, this would mean:

```php
$try_division = static function (int $dividend, int $divisor) use ($checked_divisor): string {
    return $checked_divisor($dividend, $divisor)->mapOrElse(
        static fn ($value) => sprintf('%s / %s = %s', $dividend, $divisor, $value),
        static fn() => sprintf('%s / %s failed!', $dividend, $divisor),
    )->unwrap();
};
```

From my point of view, this seems sufficient in line with your implementation as well.

If you are OK with calling `mapOrElse` with a final `unwrap()` in your codebase, I assume we don't need this and can close this?

devnix:
For sure this can be done in userland! I was just imagining a way to make some operations easier after learning Rust's `match` syntactic sugar. For sure it can be done instead with `mapOrElse()` and `unwrap()`, it's just that it's not immediately obvious 😄

We can close it if you already concluded that you don't want a function just for this

veewee:
I've given it some second thoughts and do agree that this shortcut is more than welcome. I've skimmed through some of my packages and I frequently see me using the combination of `mapOrElse` with `unwrap`. There is no alternative for this combination in rust, most likely because they have the match operator and the `?` operator as shortcuts already.

But to keep things consistent throughout the codebase, I suggest to rename it to `proceed`. That way, it's similar to what we do in `ResultInterface`.

The signature should be:

```php
    /*
     * @template Ts
     *
     * @param (Closure(T): Ts) $some
     * @param (Closure(): Ts) $none
     *
     * @return Ts
     */
    public function proceed(Closure $some, Closure $none): mixed;
```

That way, proceed will be used to map any possible scenario and unwind the result in one fixed type. That way, we can keep a consistent API throughout all these monad-like classes and interfaces.


The example will look like this:

```php
$try_division = static function (int $dividend, int $divisor) use ($checked_divisor): string {
    return $checked_divisor($dividend, $divisor)->proceed(
        some: static fn ($value) => sprintf('%s / %s = %s', $dividend, $divisor, $value),
        none: static fn() => sprintf('%s / %s failed!', $dividend, $divisor),
    );
};
```

WDYT?

devnix:
I think this one is ready too! 😄

tests/unit/Option/NoneTest.php
```diff
@@ -84,6 +84,26 @@ public function testContains(): void
         static::assertFalse($option->contains(4));
     }
 
+    public function testProceed(): void
+    {
+        $checked_divisor = static function (int $dividend, int $divisor): Option\Option {
```
veewee: Just wondering : do we need these large tests? 
The things you put in here seem rather like examples.

Maybe it could make sense to just do something like `$actual = none()->proceed(...)` and verify the correct return? That way, it makes sure there is no logic issue in the example and on top of that: one does not need to grasp the example first before seeing what is wrong with the test-case.

For other components, we are adding README.md files to the component root with examples and more information about the way it works. Maybe it makes more sense to go that way?

(for example: https://github.com/azjezz/psl/tree/next/src/Psl/Async#readme)

devnix: Indeed I've taken the examples from the Rust documentation to be familiar, but I agree they would better be in a markdown file.

I'm not confident in my English and I also don't know in depth a lot of components, so I'm somewhat scared of writing documentation. I guess I can try it in another pull request and get corrected in the reviews, but I don't want to spam a lot of them too 😊

veewee:
Looks good, can you resolve conflicts so that it can be merged,

devnix:
Ready to go @veewee!

veewee (approved):

veewee:
Nice, thanks!

PR #434, opened 2023-12-12
feat(option): add `Option::zip()`, `Option::zipWith()` and `Option::unzip()` methods

devnix:
Typing `unzip` generics has been a huge PITA. I've been able to make them work in Psalm, but not for PHPStan: https://github.com/phpstan/phpstan/discussions/10285

Eager to get feedback 😄

### DIFF
--- a/src/Psl/Option/Option.php
+++ b/src/Psl/Option/Option.php
@@ -311,7 +311,7 @@ final class Option implements Comparison\Comparable, Comparison\Equable
      * Combines two `Option` values into a single `Option` containing a tuple of the two inner values.
      * If either of the `Option`s is `None`, the resulting `Option` will also be `None`.
      *
-     * @note: If an element is `None`, the corresponding element in the resulting tuple will be `None`.
+     * @note: If an element is `None`, both elements in the resulting tuple will be `None`.
      *
      * @template Tu
      *
--- a/src/Psl/Option/Option.php
+++ b/src/Psl/Option/Option.php
@@ -315,11 +315,7 @@ final class Option implements Comparison\Comparable, Comparison\Equable
      *
      * @param Option<Tu> $other The other `Option` to zip with.
      *
-     * @return (
-     *     T is never
-     *     ? Option<never>
-     *     : (Tu is never ? Option<never> : Option<array{T, Tu}>)
-     * )
+     * @return Option<array{T, Tu}> The resulting `Option` containing the combined tuple or `None`.
      */
     public function zip(Option $other): Option
     {
--- a/src/Psl/Option/Option.php
+++ b/src/Psl/Option/Option.php
@@ -305,4 +305,77 @@ final class Option implements Comparison\Comparable, Comparison\Equable
     {
         return Comparison\equal($this, $other);
     }
+
+    /**
+     * Combines two `Option` values into a single `Option` containing a tuple of the two inner values.
+     * If either of the `Option`s is `None`, the resulting `Option` will also be `None`.
+     *
+     * @note: If an element is `None`, the corresponding element in the resulting tuple will be `None`.
+     *
+     * @template U
+     *
+     * @param Option<U> $other The other `Option` to zip with.
+     *
+     * @return Option<array{T, U}> The resulting `Option` containing the combined tuple or `None`.
+     */
+    public function zip(Option $other): Option
+    {
+        return $this->andThen(static function ($a) use ($other) {
+            return $other->map(static fn($b) => [$a, $b]);
+        });
+    }
+
+    /**
+     * Applies the provided closure to the value contained in this `Option` and the value contained in the $other `Option`,
+     * and returns a new `Option` containing the result of the closure.
+     *
+     * @template U
+     * @template Tv
+     *
+     * @param Option<U> $other The Option to zip with.
+     * @param (Closure(T, U): Tv) $closure The closure to apply to the values.
+     *
+     * @return Option<Tv> The new `Option` containing the result of applying the closure to the values,
+     *                    or `None` if either this or the $other `Option is `None`.
+     */
+    public function zipWith(Option $other, Closure $closure): Option
+    {
+        return $this->andThen(
+            /** @param T $a */
+            static function ($a) use ($other, $closure) {
+                return $other->map(
+                    /** @param U $b */
+                    static fn ($b) => $closure($a, $b)
+                );
+            }
+        );
+    }
+
+    /**
+     * @template OValue1
+     * @template OValue2
+     *
+     * @psalm-if-this-is Option<array{OValue1, OValue2}>
+     *
+     * @return array{Option<OValue1>, Option<OValue2>}
+     */
+    public function unzip(): array
+    {
+        if ($this->option === null) {
+            return [none(), none()];
+        }
+
+        /** @psalm-suppress DocblockTypeContradiction we want this check in runtime */
+        if (!is_array($this->option[0])) {
+            return [none(), none()];
+        }
+
+        if (!array_key_exists(0, $this->option[0]) || !array_key_exists(1, $this->option[0])) {
+            return [none(), none()];
+        }
+
+        [$a, $b] = $this->option[0];
+
+        return [some($a), some($b)];
+    }
 }
--- a/src/Psl/Option/Option.php
+++ b/src/Psl/Option/Option.php
@@ -6,6 +6,7 @@ namespace Psl\Option;
 
 use Closure;
 use Psl\Comparison;
+use Psl\Type;
 
 /**
  * @template T
@@ -312,11 +313,15 @@ final class Option implements Comparison\Comparable, Comparison\Equable
      *
      * @note: If an element is `None`, the corresponding element in the resulting tuple will be `None`.
      *
-     * @template U
+     * @template Tu
      *
-     * @param Option<U> $other The other `Option` to zip with.
+     * @param Option<Tu> $other The other `Option` to zip with.
      *
-     * @return Option<array{T, U}> The resulting `Option` containing the combined tuple or `None`.
+     * @return (
+     *     T is never
+     *     ? Option<never>
+     *     : (Tu is never ? Option<never> : Option<array{T, Tu}>)
+     * )
      */
     public function zip(Option $other): Option
     {
@@ -329,13 +334,13 @@ final class Option implements Comparison\Comparable, Comparison\Equable
      * Applies the provided closure to the value contained in this `Option` and the value contained in the $other `Option`,
      * and returns a new `Option` containing the result of the closure.
      *
-     * @template U
-     * @template Tv
+     * @template Tu
+     * @template Tr
      *
-     * @param Option<U> $other The Option to zip with.
-     * @param (Closure(T, U): Tv) $closure The closure to apply to the values.
+     * @param Option<Tu> $other The Option to zip with.
+     * @param (Closure(T, Tu): Tr) $closure The closure to apply to the values.
      *
-     * @return Option<Tv> The new `Option` containing the result of applying the closure to the values,
+     * @return Option<Tr> The new `Option` containing the result of applying the closure to the values,
      *                    or `None` if either this or the $other `Option is `None`.
      */
     public function zipWith(Option $other, Closure $closure): Option
@@ -344,7 +349,7 @@ final class Option implements Comparison\Comparable, Comparison\Equable
             /** @param T $a */
             static function ($a) use ($other, $closure) {
                 return $other->map(
-                    /** @param U $b */
+                    /** @param Tu $b */
                     static fn ($b) => $closure($a, $b)
                 );
             }
@@ -352,12 +357,14 @@ final class Option implements Comparison\Comparable, Comparison\Equable
     }
 
     /**
-     * @template OValue1
-     * @template OValue2
+     * @template Tv
+     * @template Tr
      *
-     * @psalm-if-this-is Option<array{OValue1, OValue2}>
+     * @psalm-if-this-is Option<array{Tv, Tr}>
      *
-     * @return array{Option<OValue1>, Option<OValue2>}
+     * @throws Type\Exception\AssertException
+     *
+     * @return array{Option<Tv>, Option<Tr>}
      */
     public function unzip(): array
     {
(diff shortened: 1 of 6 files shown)


src/Psl/Option/Option.php
```diff
@@ -305,4 +305,77 @@ public function equals(mixed $other): bool
...
+     *
+     * @return array{Option<OValue1>, Option<OValue2>}
+     */
+    public function unzip(): array
+    {
+        if ($this->option === null) {
+            return [none(), none()];
+        }
+
+        /** @psalm-suppress DocblockTypeContradiction we want this check in runtime */
```
veewee: Maybe better to remove these 2 ifs and rather assert the type this way:

```php
Type\shape([Type\mixed(), Type\mixed()])->assert($this->option);
```

It makes more sense to me to trigger a type's Assertion error than a pair of none's in this case.

In PSL v2+, we are going for better static analyzer errors meaning we will deal less with runtime errors. However, in this case I suppose it makes a lot of sense to make sure that the given value can be unzipped.

Since phpstan does not have a equivalence for `psalm-if-this-is` ATM, it must be added to https://github.com/php-standard-library/phpstan-extension. However I don't see that as a blocking issue at the moment. Rather something that needs to be done for phpstan to infer the types correctly.


Can you provide a static analysis test-case in this folder as well to make sure it works (and keeps on working) as expected:
tests/static-analysis/Option

tests/fixture/Point.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Tests\Fixture;
+
+final class Point
+{
+    public function __construct(
+        public readonly float $x,
```
veewee: asserting floats is error-prone on different systems with differend precision. Maybe better to just make it an int?

tests/unit/Option/NoneTest.php
```diff
@@ -130,4 +130,44 @@ public function testEquality(): void
...
+        static::assertTrue($x->zipWith($y, static fn($a, $b) => $a + $b)->isNone());
+        static::assertTrue($y->zipWith($x, static fn($a, $b) => $a + $b)->isNone());
+    }
+
+    /**
+     * @dataProvider provideTestUnzip
+     */
+    public function testUnzip(Option\Option $option): void
+    {
+        [$x, $y] = $option->unzip();
```
veewee, line 157: If you implement the change requested above, you'll only need to validate the none() case.

tests/unit/Option/SomeTest.php
```diff
@@ -129,4 +130,41 @@ public function testEquality()
...
+     */
+    public function testUnzip(Option\Option $option, mixed $expectedX, mixed $expectedY): void
+    {
+        [$x, $y] = $option->unzip();
+
+        static::assertSame($expectedX, $x->unwrap());
+        static::assertSame($expectedY, $y->unwrap());
+    }
+
+    private function provideTestUnzip(): iterable
```
veewee, line 165: You'll have to split this one up in 2 different cases: the case where you are able to unzip some([a, b]) and the case where the data is not a tuple of [A, B]

veewee (changes requested):

src/Psl/Option/Option.php
devnix: > Can you provide a static analysis test-case in this folder as well to make sure it works (and keeps on working) as expected:
tests/static-analysis/Option

On the way!

devnix: >Maybe better to remove these 2 ifs and rather assert the type this way:
>
>```php
>Type\shape([Type\mixed(), Type\mixed()])->assert($this->option);
>```

There is a problem with this approach: it gives a false `PossiblyNullArrayAccess` after converting `list{list{OValue1:fn-psl\option\option::unzip as mixed, OValue2:fn-psl\option\option::unzip as mixed}}` (inferred by `@psalm-if-this-is`) back into `list{array{0: mixed, 1: mixed}}|null`, and is also assigning `mixed` to `$a` and `$b`.

While I was doing different cases for the tests, I tested and discarded a version in which I would manually verify the tuple, and throw a `Psl\Option\Exception\UnzipException`. Would you be OK with that, @veewee?

tests/fixture/Point.php
devnix: Or a `numeric-string`, but I'm not sure how this could be a problem

src/Psl/Option/Option.php
veewee: This seems to work psalm-wise:

https://psalm.dev/r/9c6fa21644

I must admit it's a bit "hacky", but it's an error that shouldn't happen in runtime if you have your SAT doing the work for you. So I'm not sure here if it's a good idea to introduce an exception that should never occur cause it's being caught up front. WDYT?

tests/fixture/Point.php
veewee: see https://www.php.net/manual/en/language.types.float.php

> So never trust floating number results to the last digit, and do not compare floating point numbers directly for equality. If higher precision is necessary, the [arbitrary precision math functions](https://www.php.net/manual/en/ref.bc.php) and [gmp](https://www.php.net/manual/en/ref.gmp.php) functions are available.

src/Psl/Option/Option.php
devnix: That's interesting! The other way I see also is to keep the "manual" type checking that I have already implemented and throw an `invariant_violation()`. We would miss the structure of the type error in the exception message.

Both are doable, one requiring a private method to perform that smart but hacky assertion that would give a complete `AssertionException`, and the other one would not add a private method and would throw an `InvariantViolationException`. Do you have any preference?

veewee: > one requiring a private method to perform that smart but hacky assertion

You don't need a private function for that. You could do `Type\vec(Type\shape([Type\mixed(), Type\mixed()]))->assert($options);` 

Because that is what the `assertTuppleOption` is basically doing. (I've added it there to see how psalm behaves - but it is essentially the same). So I guess that would be the preferred solution for me ATM. Hope it won't bite us when a psalm update is performed though :)

devnix: About this
> You don't need a private function for that. You could do Type\vec(Type\shape([Type\mixed(), Type\mixed()]))->assert($options);

There's this problem I mentioned above. Probably it's behaving differently than a normal psalm assertion because of the Psalm plugin, but if I'm not wrong just erases the types of the tuple inside the function

> >Maybe better to remove these 2 ifs and rather assert the type this way:
> >
> >```php
> >Type\shape([Type\mixed(), Type\mixed()])->assert($this->option);
> >```
> 
> There is a problem with this approach: it gives a false `PossiblyNullArrayAccess` after converting `list{list{OValue1:fn-psl\option\option::unzip as mixed, OValue2:fn-psl\option\option::unzip as mixed}}` (inferred by `@psalm-if-this-is`) back into `list{array{0: mixed, 1: mixed}}|null`, and is also assigning `mixed` to `$a` and `$b`.
> 
> While I was doing different cases for the tests, I tested and discarded a version in which I would manually verify the tuple, and throw a `Psl\Option\Exception\UnzipException`. Would you be OK with that, @veewee?

src/Psl/Option/Option.php
```diff
@@ -305,4 +305,77 @@ public function equals(mixed $other): bool
...
         return Comparison\equal($this, $other);
     }
+
+    /**
+     * Combines two `Option` values into a single `Option` containing a tuple of the two inner values.
+     * If either of the `Option`s is `None`, the resulting `Option` will also be `None`.
+     *
+     * @note: If an element is `None`, the corresponding element in the resulting tuple will be `None`.
+     *
+     * @template U
```
veewee: Let's try to keep the template names consistent throughout the class as well:
`Tu` seems better.

src/Psl/Option/Option.php
```diff
@@ -305,4 +305,77 @@ public function equals(mixed $other): bool
...
+        return $this->andThen(static function ($a) use ($other) {
+            return $other->map(static fn($b) => [$a, $b]);
+        });
+    }
+
+    /**
+     * Applies the provided closure to the value contained in this `Option` and the value contained in the $other `Option`,
+     * and returns a new `Option` containing the result of the closure.
+     *
+     * @template U
```
veewee: Let's try to keep the template names consistent throughout the class (and with rust) as well:

```
* @template Tu  
* @template Tr
```

src/Psl/Option/Option.php
```diff
@@ -305,4 +305,77 @@ public function equals(mixed $other): bool
...
+                return $other->map(
+                    /** @param U $b */
+                    static fn ($b) => $closure($a, $b)
+                );
+            }
+        );
+    }
+
+    /**
+     * @template OValue1
```
veewee: Let's try to keep the template names consistent throughout the class (and with rust) as well:

```
* @template Tv
* @template Tr
```

tests/unit/Option/SomeTest.php
devnix: I'm not sure what you mean:

The `Option\some(null)->zip(Option\some('hi'))` returns an `Option<array{null, 'hi'}>`, which would be a valid tuple.

`Option\some(1)->zip(Option\some('hi'))` similarly returns an `Option<array{1, 'hi'}>`, another valid tuple.

Also, `Option\some([true, false])` is another way to build the same valid tuple to unzip, `Option<array{true, false}>`.

There are cases where the data is not a valid tuple covered in `NoneTest`. For example, `Option\none()->zip(Option\some(1))` is a valid tuple of `Psl\Option\Option<list{never, 1}>` that will correctly unzip into two `None`s

veewee: I meant adding tests for the situation where you want to unzip e.g. some(1) - which is not possible by the tuple type assertion that is added in the code.

devnix: Ow, I was just facing a mutation right now
```diff
$option = $this->option[0];
-Type\shape([Type\mixed(), Type\mixed()])->assert($option);
+
[$a, $b] = $option;
return [some($a), some($b)];
```
and I remembered that I had in a non-pushed version exactly what you mean, sorry!

devnix:
Sorry for having to review it so many times, I hope it's ready now! 😄

src/Psl/Option/Option.php
```diff
@@ -305,4 +306,78 @@ public function equals(mixed $other): bool
...
+     * Combines two `Option` values into a single `Option` containing a tuple of the two inner values.
+     * If either of the `Option`s is `None`, the resulting `Option` will also be `None`.
+     *
+     * @note: If an element is `None`, the corresponding element in the resulting tuple will be `None`.
+     *
+     * @template Tu
+     *
+     * @param Option<Tu> $other The other `Option` to zip with.
+     *
+     * @return (
```
devnix: `T is never` is still not supported by PHPStan (I would not consider it a blocker): https://github.com/phpstan/phpstan/issues/9634

veewee: This is also quite "edgy" in psalm tbh.
See: https://github.com/vimeo/psalm/issues/8487

You can't really trust it to return either never or null.
I'dd say returning a non-conditional `Option<array{T, Tu}>` is the way to go here.
(as we do in all other places as well)

This will be valid cause psalm checks it this way:
https://psalm.dev/r/9715a63a4f
(as we discovered in https://github.com/azjezz/psl/pull/415 )

src/Psl/Option/Option.php
```diff
@@ -305,4 +306,78 @@ public function equals(mixed $other): bool
     {
         return Comparison\equal($this, $other);
     }
+
+    /**
+     * Combines two `Option` values into a single `Option` containing a tuple of the two inner values.
+     * If either of the `Option`s is `None`, the resulting `Option` will also be `None`.
+     *
+     * @note: If an element is `None`, the corresponding element in the resulting tuple will be `None`.
```
veewee: I'm a bit confused here. This is not true right?
If either of them is None, the result is None.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=b59993df008ad0c70f0a4c979f1e5dfb

veewee (commented):
> Sorry for having to review it so many times, I hope it's ready now! 😄

No worries :) I rather review 100 times and make sure to ship something good than to rush in changes.

I think we're almost there. Added 2 remarks and then it should be good to go I guess.

src/Psl/Option/Option.php
devnix: goddamn-right-yes

src/Psl/Option/Option.php
```diff
@@ -305,4 +306,78 @@ public function equals(mixed $other): bool
     {
         return Comparison\equal($this, $other);
     }
+
+    /**
+     * Combines two `Option` values into a single `Option` containing a tuple of the two inner values.
+     * If either of the `Option`s is `None`, the resulting `Option` will also be `None`.
+     *
+     * @note: If an element is `None`, both elements in the resulting tuple will be `None`.
```
veewee: I don't think this is true either : it will just be none(). Not a tuple of [none(), none()].
You can probably just drop this line, cause you mention how it works the line above.

src/Psl/Option/Option.php
devnix: >You can't really trust it to return either never or null.

Not sure about this. As I've been testing, when it _knows_ that you have a `None` value, it's going to be a consistent `Option<never>`. It's covered in 
https://github.com/azjezz/psl/pull/434/files/4eddf3143fe5eb9c72ce665a0748abad17a97cbb#diff-bda3d7e943382d0856488cf5330399ae97aa009d528b8b896df1ce183014dde4

> I'dd say returning a non-conditional Option<array{T, Tu}> is the way to go here.

By doing this, the result of `Option\none()->zip(Option\some(1))` would be `Psl\Option\Option<list{never, 1}>`, which is wrong as far as I understand, it should be `Option<never>`

> (as we do in all other places as well)

Indeed if we keep a conditional return here, I would be missing it in `zipWith()` 🤦‍♂️

veewee: The thing is that it will mostly never be `never` imo.
Cause `never` means `none()`, and you always want to work with `some(T)` === `option<T>`.
Making psalm think it is all about the `T` and `never` is just if it's not there.

So you'll probably have:

`Option<int>::zip(Option<string>)` ===  `Option<array{int, string}>`.

(of which one could actually be `Option<never>` - but it doesn't matter cause psalm knows `never|string` is `string`.

In more readable words: you won't typehint userland code with never, but rather with what you expect to be in there.

src/Psl/Option/Option.php
devnix: Oof, right. I've thought with the logic of the `unzip()` function, that should return the tuple I mention (if I'm not mistaken again 😄 )

src/Psl/Option/Option.php
devnix: >The thing is that it will mostly never be `never` IMO.

I feel the same, knowing beforehand that you are working with a `none` is going to be found on a bit of an absurd code or simple examples, but having the possibility that static analysis thinks that you can have a `Psl\Option\Option<list{never, int}>` triggers my OCD if I can infer the type better beforehand 🤣

The main problem is later with the `unzip` function. If we don't type it a conditional return and we unzip `Psl\Option\Option<list{never, int}>`, we get a `list{Psl\Option\Option<never>, Psl\Option\Option<int>}`, which is not the real behavior too. 

image

veewee: If it's typed that way, i'dd say it's a userland issue. You can perfectly provide the correct types.
Or it could ad well actually be never. Anyways, I think removing the condition should be fine here

devnix:
This should do it, happy holidays! 😄

veewee:
Thanks

PR #435, opened 2023-12-14
feat(type): add class_string types (#432)

### LINKED ISSUE #432: Type\implements_interface
There is a `Assert::implementsInterface($className, $interface)` method that is implemented as 

```php
    /**
     * @psalm-pure
     * @psalm-template ExpectedType of object
     * @psalm-param class-string<ExpectedType> $interface
     * @psalm-assert class-string<ExpectedType> $value
     *
     * @param mixed  $value
     * @param mixed  $interface
     * @param string $message
     *
     * @throws InvalidArgumentException
     */
    public static function implementsInterface($value, $interface, $message = '')
    {
        if (!\in_array($interface, \class_implements($value))) {
            static::reportInvalidArgument(\sprintf(
                $message ?: 'Expected an implementation of %2$s. Got: %s',
                static::valueToString($value),
                static::valueToString($interface)
            ));
        }
    }
```

`Psl\invariant` solution for the similar problem (to check whether a class name implements an interface) does not provide the same psalm type assertion still.

```php
Type\implements_interface($interface)->asset($className);
```

`Psl\invariant`, but it does not narrow the type.

**Additional context**

Happy to provide a PR :-)

zerkms:
fixes #432

This PR creates a new `Type\class_string(T)` that allows to assert that the value is of a type `class-string<T>`.

Example of use:

```php
$value = Type\class_string(Psl\Collection\CollectionInterface::class)->assert('foo');
/** @psalm-trace $value */;

// $value: class-string<Psl\Collection\CollectionInterface<array-key, mixed>>
```

### DIFF
--- a/tests/unit/Type/ClassStringTypeTest.php
+++ b/tests/unit/Type/ClassStringTypeTest.php
@@ -20,6 +20,8 @@ final class ClassStringTypeTest extends TypeTest
         yield [$_ = Collection\MutableVector::class, $_];
         yield [$_ = Collection\Map::class, $_];
         yield [$_ = Collection\MutableMap::class, $_];
+        yield [$_ = Collection\MutableMapInterface::class, $_];
+        yield [$_ = Collection\CollectionInterface::class, $_];
     }
 
     public function getInvalidCoercions(): iterable
--- a/tests/unit/Type/ClassStringTypeTest.php
+++ b/tests/unit/Type/ClassStringTypeTest.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Tests\Unit\Type;
+
+use Psl\Collection;
+use Psl\Type;
+
+final class ClassStringTypeTest extends TypeTest
+{
+    public function getType(): Type\TypeInterface
+    {
+        return Type\class_string(Collection\CollectionInterface::class);
+    }
+
+    public function getValidCoercions(): iterable
+    {
+        yield [$_ = Collection\Vector::class, $_];
+        yield [$_ = Collection\MutableVector::class, $_];
+        yield [$_ = Collection\Map::class, $_];
+        yield [$_ = Collection\MutableMap::class, $_];
+    }
+
+    public function getInvalidCoercions(): iterable
+    {
+        yield [null];
+        yield [STDIN];
+        yield ['UnknownClass'];
+        yield [$this->stringable('foo')];
+        yield [new class {
+        }];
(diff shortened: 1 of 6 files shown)


tests/unit/Type/ClassStringTypeTest.php
```diff
@@ -0,0 +1,42 @@
...
+    {
+        return Type\class_string(Collection\CollectionInterface::class);
+    }
+
+    public function getValidCoercions(): iterable
+    {
+        yield [$_ = Collection\Vector::class, $_];
+        yield [$_ = Collection\MutableVector::class, $_];
+        yield [$_ = Collection\Map::class, $_];
+        yield [$_ = Collection\MutableMap::class, $_];
```
veewee, line 22: For completeness : could you add a valid coercion that checks for subclasses of a given class / interface type?

It behaves like this on psalm and given your code I think it behaves the same. But just to be sure here.

See inheritance cases in:
https://psalm.dev/r/03a91eba69

tests/unit/Type/ClassStringTypeTest.php
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public function getInvalidCoercions(): iterable
+    {
+        yield [null];
+        yield [STDIN];
+        yield ['hello'];
```
veewee: Maybe instead of 'hello' -> Could you set the value to something like : 'UnknownClass' ?
Cause that makes it clear why this fails.

veewee (commented):
Thanks! This looks very good. 
I've added 2 little remarks to improve the tests a bit. But I think it's fine to go after this.

tests/unit/Type/ClassStringTypeTest.php
zerkms: I don't understand :-( `Psl\Tests\Unit\Type\TypeTest` only allows to return one value from the `getType` and test against it.

So currently I check that coercion from specific class name to an interface works fine. What exactly do you suggest?

tests/unit/Type/ClassStringTypeTest.php
zerkms: Indeed, done.

tests/unit/Type/ClassStringTypeTest.php
veewee: Oh yeah indeed. I was assuming you could also vary in `getType`, but that's not how it works.
Thanks for pointing it out. It seems sufficient this way.

PR #437, opened 2024-01-16
Added support for coercing `stdClass` to `array<TKey, TValue>`

### LINKED ISSUE #436: Corece an `array` out of an `stdClass`?
Just throwing out ideas, but in the context of JSON -> Domain, the `ShapeType` could be useful when dealing with JSON structures where the difference between hashmap and array is sometimes to be preserved:

```php
$input = <<<'JSON'
{
  "foo": "hello world!"
}
JSON;

$data = Json\decode($input, false);

echo Type\shape([
    'foo' => Type\non_empty_string(),
])->coerce($data)['foo'];
```

Ocramius:
Fixes #436

This change is useful when decoding JSON hashmaps into PHP hashmaps, since PHP keeps JSON-decoded data as an `stdClass` unless explicitly told to do so.

Also note that preserving JSON structures as `stdClass` is sometimes very much required, since an empty hashmap is a lossy conversion to `[]` performed by `json_decode()`, when using `associative: true`. This patch makes preserving JSON data structures a bit easier.

This adds some very minimal overhead to `ShapeType#coerce()`, which should be imperceptible.

### DIFF
--- a/src/Psl/Type/Internal/ShapeType.php
+++ b/src/Psl/Type/Internal/ShapeType.php
@@ -8,6 +8,7 @@ use Psl\Iter;
 use Psl\Type;
 use Psl\Type\Exception\AssertException;
 use Psl\Type\Exception\CoercionException;
+use stdClass;
 
 use function array_diff_key;
 use function array_filter;
@@ -51,6 +52,10 @@ final class ShapeType extends Type\Type
      */
     public function coerce(mixed $value): array
     {
+        if ($value instanceof stdClass) {
+            $value = (array) $value;
+        }
+
         // To whom reads this: yes, I hate this stuff as passionately as you do :-)
         if (! is_array($value)) {
             // Fallback to slow implementation - unhappy path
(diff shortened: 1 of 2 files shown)


src/Psl/Type/Internal/ShapeType.php
```diff
@@ -51,6 +52,10 @@ public function __construct(
      */
     public function coerce(mixed $value): array
     {
+        if ($value instanceof stdClass) {
```
Ocramius, line 55: Question: perhaps `get_class($value) === stdClass::class` too? :thinking: 

I don't know if people are crazy enough to extend `stdClass` in 2023, but given that the PHP community has regressed to architectural styles from 2007 over time...

veewee: Even if people are crazy enough to extend `stdClass` in 2024 😉 , it could still be coerced into the specified shape.
So I don't see a big issue with just using the `instanceof` check.

Unless you have a different opinion, this one looks good to go to me!

veewee (approved):

veewee:
Thanks!

PR #441, opened 2024-01-27
`json_encode` never returns an empty string

gsteel:
… at least I don't think it can:

https://3v4l.org/8G6ln


gsteel:
Sorry, not sure how the docs work. I've amended the referenced line number for `encode` but it's still red.

veewee:
Seems valid indeed:
https://psalm.dev/r/2106809556

Thanks for adding.

The documentation is auto-generated and can be altered by the command: `make docs-generate`

veewee:
Thanks!

PR #445, opened 2024-03-19
feat(locale): introduce 'Locale' component

azjezz:


src/Psl/Locale/Locale.php
```diff
@@ -0,0 +1,896 @@
...
+declare(strict_types=1);
+
+namespace Psl\Locale;
+
+use Locale as NativeLocale;
+
+/**
+ * Represents a locale identifier.
+ */
+enum Locale: string
```
azjezz, line 12: The cases for this enum were mostly generated using a script, we could add more in the future.

veewee: Would it make sense to include said script and check from time to time that new locales are added automatically?

veewee (approved):
LGTM 🚢

src/Psl/Locale/Locale.php
azjezz: This is what I used actually, but had to go over the result manually and update some names as they contained non-ASCII characters.

I think it would be nice to have a nice script for this in the future, but I don't see the value on having this in the repo.

```php
foreach (\ResourceBundle::getLocales('') as $locale) {
    $name = \Locale::getDisplayName($locale, 'en');
    $name = \Psl\Str\replace_every($name, [
        '(' => '',
        ')' => '',
        '[' => '',
        ']' => '',
        ',' => '',
        '.' => '',
        '&' => '',
        '-' => '',
        ' ' => '',
    ]);

    echo 'case ' . $name . ' = "' . $locale . '";' . PHP_EOL;
}
```

PR #446, opened 2024-03-21
feat(datetime): Introduce `DateTime` component

### LINKED ISSUE #27: DateTime API

azjezz:
~This is the first draft of the upcoming `DateTime` component, and would act as a base, not a final propsal.~

Please let me know your thoughts, what do you like, what do you dislike, what is there to add/remove.

closes #27 

~NOTE: this PR is lacking tests, and does not pass static analysis.~


azjezz:
@veewee let me know what you think of this inital draft :)

src/Psl/DateTime/Era.php
```diff
@@ -0,0 +1,40 @@
...
+ * Represents the two eras in the Gregorian and Julian calendars as an enum.
+ *
+ * This enum distinguishes between the Anno Domini (AD) era, denoting years after the birth of Jesus Christ,
+ * and the Before Christ (BC) era, denoting years before the birth of Jesus Christ. It provides a type-safe way
+ * to represent and work with these two divisions of historical time.
+ */
+enum Era: string
+{
+    case AnnoDomini = 'AD';
+    case BeforeChrist = 'BC';
```
azjezz, line 17: wasn't sure here, should we use AD/BC or CE/BCE? i personally lean more towards AD/BC.

but again, do we even need `Era` enum? is this even useful? ( ref `Psl\DateTime\DateTimeInterface::getEra()` ).

src/Psl/Async/Scheduler.php
```diff
@@ -114,35 +115,43 @@ public static function defer(Closure $callback): string
     /**
      * Delay the execution of a callback.
      *
-     * @param float $delay The amount of time, to delay the execution for in seconds.
+     * @param DateTime\Duration|float $delay The amount of time, to delay the execution for in seconds.
```
veewee: Would it make sense to deprecate the usage of `float` and drop it in next major release?
(same for all of those similar changes in this PR)

src/Psl/DateTime/DateFormat.php
```diff
@@ -0,0 +1,24 @@
...
+ * This enum provides a collection of standardized date format strings for various protocols
+ * and standards, such as RFC 2822, ISO 8601, HTTP headers, and more.
+ */
+enum DateFormat: string
+{
+    case Rfc2822 = 'EEE, dd MMM yyyy HH:mm:ss Z';
+    case Iso8601 = 'yyyy-MM-dd\'T\'HH:mm:ssXXX';
+    case Http = 'EEE, dd MMM yyyy HH:mm:ss zzz';
+    case Cookie = 'EEEE, dd-MMM-yyyy HH:mm:ss zzz';
+    case Sql = 'yyyy-MM-dd HH:mm:ss';
```
veewee: Maybe split it up in SQL date and Sql DateTime ?

src/Psl/DateTime/Era.php
veewee: We use AD/BC, but I haven't ever been in a coding situation where this would add any value.
For me, it's not a must to have this : The check is pretty simple `Year > 0` if you need this.
However, its here - we could as well keep it.

src/Psl/DateTime/Timezone.php
```diff
@@ -0,0 +1,546 @@
...
+     *
+     * @param TemporalInterface $temporal The temporal object to calculate the time zone offset for.
+     *
+     * @return Duration The offset from UTC as a Duration instance.
+     *
+     * @mutation-free
+     */
+    public function getOffset(TemporalInterface $temporal): Duration
+    {
+        return Internal\zone_override($this, static function () use ($temporal): Duration {
```
veewee: I have the feeling that this one is being used as a hack here.
Wouldn't it be less hacky to use `DateTimeZone::getOffset()` instead?

src/Psl/DateTime/Timestamp.php
```diff
@@ -0,0 +1,266 @@
...
+     *
+     * Returns the current time with precision up to nanoseconds, adjusted to maintain accuracy.
+     *
+     * @return self The current timestamp.
+     *
+     * @pure
+     */
+    public static function now(): self
+    {
+        $hrTime = hrtime();
```
veewee: is hrtime the right tool to use here over microtime?
As far as I understand, it should solely be used for performance metrics (since it's counted from an arbitrary point in time) rather than for timestamps.

You are playing with the offset, which I don't fully understand.

src/Psl/DateTime/Timestamp.php
```diff
@@ -0,0 +1,266 @@
...
+     * The timezone is required to accurately represent the timestamp in the desired locale.
+     *
+     * @mutation-free
+     */
+    public function format(Timezone $timezone, null|DateFormat|string $format = null, ?Locale\Locale $locale = null): string
+    {
+        return Internal\zone_override($timezone, function () use ($timezone, $locale, $format): string {
+            $obj = new NativeDateTime();
+            $obj->setTimestamp($this->getSeconds());
+            $obj->setTimezone(new NativeDateTimeZone($timezone->value));
```
veewee: I've seen you using the zone_override a couple of times for this and it seems a bit odd.
Wouldn't it be possible to set the timezone through the constructor instead of relying on the system's timezone?

src/Psl/DateTime/Internal/zone_override.php
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @param (Closure(): T) $callback
+ *
+ * @return T
+ *
+ * @internal
+ *
+ * @pure - This technically a lie.
+ */
+function zone_override(?Timezone $tz, Closure $callback): mixed
```
veewee: I don't really like this function and have the feeling it is used as some kind of workaround in a lot of scenarios.

src/Psl/DateTime/DateTime.php
```diff
@@ -0,0 +1,417 @@
...
+     *
+     * @return int<1, 12>
+     *
+     * @mutation-free
+     */
+    public function getMonth(): int
+    {
+        return $this->month;
+    }
+
```
veewee, line 391: maybe something like `getWeekOfYear()` could be handy here.

veewee (commented):
Hello,

What a great addition!
I went through the code and added some initial comments here and there.

Most of the functionalities seem to be covered already, I don't have a wishlist of missing features at this moment. They will surely come once I start using this library though.

My head is pretty full at the moment, so I'll come back later to hopefully give some more thoughts on the component from my end. But I already wanted to share these initial comments.

Great job BTW! :)

src/Psl/DateTime/Timezone.php
azjezz: `DateTimeZone::getOffset()` requires a `DateTime` parameter, and internally its basically doing what we are doing here.

src/Psl/DateTime/Timestamp.php
azjezz: Hrtime allows us to calculate the nanoseconds between every call to `now()`, so when you retrieve two points in time, you will be able to calculate the difference between them down to the nanosecond.

Psl is not the only lib that does this, see: https://github.com/symfony/clock/blob/7.0/MonotonicClock.php

<details>
<summary>How hrtime/offset works</summary>
Yes, hrtime is from an arbitry point of time, but its precious!

to explain what we do here, lets say time() returns `100`, while hrtime() returns `[2, 4000]`

after 2 seconds, and 1000 ns have passed, `time()` would return `102`, while `hrtime()` would return `[4, 5000]`.

so the first time `now()` is called, we get the hrtime:

```
$hrtime = hrtime(); // [2, 4000]
```

check if we have previously set the different between `time()` and `hrtime()`, if not, calculate the different
```
$time = time(); // 100
$different = [
  'seconds' => $time - $hrtime[0], // 100 - 2 = 98
  'nanoseconds' => $hrtime[1],
];
```

then, we calcualte the time in both seconds an nanoseconds, like this:

```
$seconds = $hrtime[0] + $different['seconds']; // 2 + 98 = 100
$nanoseconds = $hrtime[1] + $different['nanoseconds]; // 4000 + 4000 = 8000;

```

so, the first call to `now()` results in `[100, 8000]`

after `1000` ns have passed, if we call `now()` again, what happens is:

```
$hrtime = hrtime(); // [2, 5000]
```

we then add offset:

```
$seconds = $hrtime[0] + $different['seconds']; // 2 + 98 = 100
$nanoseconds = $hrtime[1] + $different['nanoseconds]; // 5000 + 4000 = 9000;
```

so the result we get is `[100, 9000]`, much more accurate than regular `time()`, allowing you to calcualte the difference between 2 points of time down to the nanosecond!

but, there is more!

lets say 10 seconds, and 999999999 nanoseconds have passed!


```
$hrtime = hrtime(); // [12, 999999999]
```

we then add offset:

```
$seconds = $hrtime[0] + $different['seconds']; // 12 + 98 = 110
$nanoseconds = $hrtime[1] + $different['nanoseconds]; // 999999999 + 4000 = 1000004000;
```

this later gets normalized to `[111, 4000]`
</details>

src/Psl/DateTime/DateTime.php
azjezz: we have `getIsoWeekNumber`, it returns [year, week_number], as according to the ISO standard, a week could belong to the previosu year, or next year, as explained in the method docblock.

src/Psl/DateTime/Internal/zone_override.php
azjezz: removed. we now use IntlTimezone/IntlCalendar/IntlDateFormatter internally, none of which need global state, so our code is not truly immutable :D

src/Psl/DateTime/Timestamp.php
azjezz: replaced by IntlTimezone.

src/Psl/Async/Scheduler.php
azjezz: I'm not sure, `Async\sleep(0.5)` IMO is still clear, while `Async\sleep(DateTime\Duration::milliseconds(500))` could be too verbose. some people might lean more toward using float for most cases.

azjezz: also, a lot of changes related to this in this PR are BC-breaking, so we need a new major for this PR, ( e.g parameter change on interfaces such as `IO\ReadHandleInterface`, and more )

src/Psl/DateTime/DateTime.php
```diff
@@ -159,96 +142,113 @@ public static function todayAt(int $hours, int $minutes, int $seconds = 0, int $
      */
     public static function fromParts(int $year, Month|int $month, int $day, int $hours = 0, int $minutes = 0, int $seconds = 0, int $nanoseconds = 0, Timezone $timezone = null): self
```
veewee: This is the only static constructor where timezone is optional.
How do you see it : always explicitly provide it (even though a lot of the times you just want the default one) or optionally provide it?

src/Psl/DateTime/Timezone.php
```diff
@@ -487,60 +485,94 @@ enum Timezone : string
...
-     * its corresponding {@see Timezone} enum instance. If the system's default time zone does not match any
-     * of the enum cases, UTC is returned as the default.
+     * @param bool $local Indicates whether the temporal object's time should be treated as local time (`true`) or as UTC time (`false`).
      *
-     * @return Timezone The system's default time zone as a {@see Timezone} enum instance or UTC if not found.
+     * @return Duration The total offset from UTC as a Duration instance, including any DST adjustments.
+     *
+     * @mutation-free
      */
-    public static function default(): Timezone
```
veewee: no more "default" (/system) timezone?
It did make sense to me to have it. Otherwise you need to explicitly set it everywhere.

src/Psl/DateTime/Timestamp.php
veewee: Thanks for the explanation. Makes sense indeed!
Do you need the `hrtime() === false` check like symfony has here as well and is there a fallback? (for example no nanoseconds will be used - or up to microseconds)?

src/Psl/DateTime/DateTime.php
azjezz: actually i missed this, it should be required.

I don't know how we can define a default timezone without touching the global state.

src/Psl/DateTime/Timezone.php
azjezz: Yes, that is the idea.

I'm still not sure about it however, we either have default timezone/locale, or not.

If yes, that means we will be accessing global state, e.g:

```php
$a = DateTime\DateTime::now()->getTimezone();

something_else();

$b = DateTime\DateTime::now()->getTimezone();
```

there will be no garuntee that both `$a` and `$b` are the same.

we could store the value interanlly ourselves to ensure future calls return the same value, but then, how do you change the default value at runtime?

Note: `Psl\Locale\Locale` also does not have 'default()` method, so if we are adding one here, we should add one for locale as well.

src/Psl/DateTime/DateFormat.php
azjezz: done

src/Psl/DateTime/Timestamp.php
azjezz: Yep, i have added a check, and i also split this function into two, `now()` and `monotonic()`, `now()` is enough for almost all cases, while `monotonic()` is useful for things like performance measurment.

note: `now()` uses `microtime()`, so while it is not precious to the nanoseconds, it is precious to the millisecond, making it a better alternative to `time()`.

src/Psl/DateTime/DateTime.php
azjezz: timezone is now optional in all static functions, we use Timezone::default(), but this means these functions can't be pure.

src/Psl/DateTime/Timezone.php
azjezz: I have added `Timezone::default()` and `Locale::default()`, neither are pure.

src/Psl/Async/Scheduler.php
azjezz: dropped float!

azjezz:
I feel like the current API is "good enough" for a first iteration now, though it's just lacking a bit more in testing.

Things I'm unsure about:

- Do we need `ZonedTemporalInterface`?
- Should `TemporalInterface` be renamed into `UnzonedTemporalInterface`?
- Should we split `DateTimeInterface` into `UnzonedDateTimeInterface` and `ZonedDateTimeInterface`? (The same goes for `DateTime` -> `UnzonedDateTime` + `ZonedDateTime`)
- Regarding the `FormatPattern` enum, are the current patterns sufficient, or should we add or remove some?
- About purity: many methods are not pure because they need to use `Timezone::default()` and `Locale::default()`, which are mutation-free but rely on system settings (INI configs) to retrieve the default values ( therefore, not pure ). Should we instead use `Timezone::UTC` / `Locale::English` to make those methods pure? Is it worth it?
  - If so, should we adjust `Timezone::default()` to return `Timezone::UTC` and introduce `Timezone::system()` to retrieve the system default?
  - Similarly, should `Locale::default()` return `Locale::English`, and should there be a `Locale::system()` to retrieve the system default?

azjezz:
additional comment by @veewee on [Discord](https://discord.gg/m3TDcP4tNP): 

> I have the feeling that splitting it up into 2 concepts introduces more issues than it solves. 
> The initial discussion was about making Timezone optional or required.
I don't see the benefit in having them both, cause that ends up in 2 different ways of using the component.
> 
> So maybe to summarize:
> 
> If we add timezones required, the component will result in less bugs given you need to specify and therefore think about the Timezone constanntly.
> 
> Which brings me to adding it optionally: results in an easier to use API by assuming the default. When writing cross Timezone code it relies on UTC (or system) and it accepts a way to override the Timezone. 
> 
> When creating cross Timezone code, the default Timezone gets repeated frequently throughout the codebase anyways. 
> 
> Which brings me to the question: what benefit does making required have over optional?
> 
>  It is very explicit and verbose, which makes it an annoying repeating thing. 
So do we want the component to be very explicit or rather implicit?

src/Psl/DateTime/Timestamp.php
```diff
@@ -0,0 +1,305 @@
...
+        null|FormatPattern|string $pattern = null,
+        null|Timezone $timezone = null,
+        null|Locale $locale = null,
+    ): static {
+        $formatter = Internal\create_intl_date_formatter($date_style, $time_style, $pattern, $timezone, $locale);
+
+        $timestamp = $formatter->parse($raw_string);
+        if ($timestamp === false) {
+            // Only show pattern in the exception if it was provided.
+            // TODO(azjezz): should we have `ParserException` which contains all passed parameters ( styles, locale, timezone, pattern .. etc)?
```
veewee: missing todo ;)

src/Psl/DateTime/Timestamp.php
```diff
@@ -0,0 +1,305 @@
...
+        /** @psalm-suppress MissingThrowsDocblock */
+        return self::fromRaw((int) $timestamp);
+    }
+
+    /**
+     * Returns this Timestamp instance itself, as it already represents a timestamp.
+     *
+     * @mutation-free
+     */
+    public function getTimestamp(): self
```
veewee, line 187: this is a bit of a weird function. What's the use-case?

src/Psl/DateTime/Timestamp.php
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    /**
+     * Returns the raw seconds and nanoseconds of the timestamp as an array.
+     *
+     * @return array{int, int<0, 999999999>}
+     *
+     * @mutation-free
+     */
+    public function toRaw(): array
```
veewee: is 'raw' the best name we could imagine here?

src/Psl/DateTime/Timestamp.php
```diff
@@ -0,0 +1,305 @@
...
+     * @param null|Timezone $timezone Determines the timezone for which the date and time are formatted. Defaults to {@see Timezone::default()}.
+     * @param null|Locale $locale Specifies the locale to be used for formatting locale-specific date and time components. Defaults to {@see Locale::default()}.
+     *
+     * @return string The formatted date and time string.
+     *
+     * @see https://unicode-org.github.io/icu/userguide/format_parse/datetime/#datetime-format-syntax
+     *
+     * @mutation-free
+     */
+    public function format(null|FormatDateStyle $date_style = null, null|FormatTimeStyle $time_style = null, null|FormatPattern|string $pattern = null, null|Timezone $timezone = null, null|Locale $locale = null): string
```
veewee: this is probably something we need to document: there are a lot of options.
Does it make sense to have both day / time styles and patterns in the same function?

azjezz: I thought about having:

```
format($pattern, $timezone, $locale)
toString($date_style, $time_style, $timezone, $locale)
__toString() ( alias for toString(null, null, null, null) ).
```

but not sure. 🤔

src/Psl/DateTime/Timestamp.php
azjezz: `toParts()` ?

src/Psl/DateTime/Timestamp.php
azjezz: the idea is that every `TemporalInterface` object should be convertable to a Timestamp, Timestamp itself is an implementation of `TemporalInterface`, so it needs to implement this.

src/Psl/DateTime/DateTime.php
```diff
@@ -0,0 +1,490 @@
...
+    {
+        if (
+            $nanoseconds < 0 || $nanoseconds >= NANOSECONDS_PER_SECOND ||
+            $seconds < 0 || $seconds >= 60 ||
+            $minutes < 0 || $minutes >= 60 ||
+            $hours < 0 || $hours >= 24 ||
+            $month < 1 || $month > 12 ||
+            $day < 1 || $day > 31 || $day > Month::from($month)->getDaysForYear($year)
+        ) {
+            throw new Exception\InvalidArgumentException('One or more components of the date-time are out of valid ranges.');
```
veewee: This error message is not very helpful though.

src/Psl/DateTime/DateTime.php
```diff
@@ -0,0 +1,490 @@
...
+                $calendar->get(IntlCalendar::FIELD_YEAR) === $year &&
+                ($calendar->get(IntlCalendar::FIELD_MONTH) + 1) === $month &&
+                $calendar->get(IntlCalendar::FIELD_DAY_OF_MONTH) === $day &&
+                $calendar->get(IntlCalendar::FIELD_HOUR_OF_DAY) === $hours &&
+                $calendar->get(IntlCalendar::FIELD_MINUTE) === $minutes &&
+                $calendar->get(IntlCalendar::FIELD_SECOND) === $seconds
+            )
+        ) {
+            throw new Exception\InvalidArgumentException(
+                'The given components do not form a valid date-time.',
```
veewee: This error message is not very helpful though.

src/Psl/DateTime/DateTime.php
```diff
@@ -0,0 +1,490 @@
...
+    }
+
+    /**
+     * Returns the month.
+     *
+     * @return int<1, 12>
+     *
+     * @mutation-free
+     */
+    public function getMonth(): int
```
veewee, line 387: Should we also have a method for grabbing the Month as the enum ?

src/Psl/DateTime/Timestamp.php
veewee: that would be more consistent with DateTime indeed

src/Psl/DateTime/Timestamp.php
veewee: Something like that could be nice indeed. Don't think we really need `__toString()`, cause that wont take any params.

src/Psl/DateTime/DateTime.php
```diff
@@ -0,0 +1,490 @@
...
+
+use IntlCalendar;
+use Psl\Locale\Locale;
+
+/**
+ * @immutable
+ *
+ * @psalm-suppress ImpureMethodCall
+ */
+final readonly class DateTime implements DateTimeInterface
```
veewee, line 13: I'm missing interoperability methods to transform from and into PHP's std `DateTimeInterface`.

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
```diff
@@ -0,0 +1,473 @@
...
+        $current_year = $this->getYear();
+
+        // Check for potential overflow when adding years
+        if ($current_year >= Math\INT64_MAX - $years || $current_year <= Math\INT64_MIN - $years) {
+            throw new Exception\OverflowException("Adding years results in a year that exceeds the representable integer range.");
+        }
+
+        $target_year = $current_year + $years;
+
+        // Handle the day and month for the target year, considering leap years
```
veewee: Just thinking out loud here : would it make sense to just call plusMonths($years * MONTHS_IN_YEAR) here? 
That way the logic does not need to get repeated in plus/minus months and years

src/Psl/DateTime/DateTime.php
```diff
@@ -0,0 +1,490 @@
...
+     * Adds the specified duration to this date-time object, returning a new instance with the added duration.
+     *
+     * @throws Exception\UnderflowException If adding the duration results in an arithmetic underflow.
+     * @throws Exception\OverflowException If adding the duration results in an arithmetic overflow.
+     *
+     * @mutation-free
+     */
+    public function plus(Duration $duration): static
+    {
+        return static::fromTimestamp($this->timezone, $this->getTimestamp()->plus($duration));
```
veewee: Every plus / minus could result in the previous / next month or leap year problematic.
We are now only solving it for months and years - but the same could really go for minutes and hours right?
Not sure what the best place is to handle it.

src/Psl/DateTime/constants.php
```diff
@@ -0,0 +1,117 @@
...
+const MINUTES_PER_WEEK = 10_080;
+
+/**
+ * The number of hours in a day.
+ *
+ * @var int
+ */
+const HOURS_PER_DAY = 24;
+
+/**
```
veewee, line 105: maybe also

MONTHS_PER_YEAR

WEEKS_PER_YEAR is probably not possible due to 51-53 range?

src/Psl/DateTime/ZonedTemporalConvenienceMethodsTrait.php
```diff
@@ -0,0 +1,68 @@
...
+declare(strict_types=1);
+
+namespace Psl\DateTime;
+
+use Psl\Locale\Locale;
+
+/**
+ * @require-implements ZonedTemporalInterface
+ */
+trait ZonedTemporalConvenienceMethodsTrait
```
veewee: Still think everything is zoned (either implicit or explicit) so i don't see a big benifit of splitting it up in Zoned or not.

src/Psl/DateTime/DateTime.php
veewee: I'm also wondering : how does it deal with daylight saving time?
As in : If I add an hour an half hour before the DST hour get shifted, it should be the same hour.

(for example : 2.30am + 1h = 2.30am when at 3.00am the time gets shifted back to 2.00am) Sure it's a weird scenario which I don't really know if it's actually valid.)

src/Psl/DateTime/Timestamp.php
azjezz: I think `__toString()` is okay to keep, as it allows for `echo $temporal;`

src/Psl/DateTime/ZonedTemporalConvenienceMethodsTrait.php
azjezz: removed!

src/Psl/DateTime/DateTime.php
azjezz: leaving those when we add `Psl\Interoperability` component :)

src/Psl/DateTime/constants.php
azjezz: ADDED MONTHS_PER_YEAR, WEEKS_PER_YEAR is not static :/

src/Psl/DateTime/DateTime.php
azjezz: It deals with it better than PHP i would say :D not perfect maybe?

```php
<?php

declare(strict_types=1);

use Psl\DateTime;
use Psl\IO;

require_once 'vendor/autoload.php';

$timestamp = DateTime\Timestamp::fromParts(seconds: 1711846740);
$datetime = DateTime\DateTime::fromTimestamp($timestamp, DateTime\Timezone::EuropeLondon);

// outputs: 31 Mar 2024, 00:59:00 / 0 second(s) offset
IO\write_error_line('%s / %s offset', $datetime->toString(), $datetime->getTimezoneOffset()->toString());

// adding 1-hour results in entering DST, as it starts 1am
$datetime = $datetime->plus(DateTime\Duration::hours(1));

// outputs: 31 Mar 2024, 02:59:00 / 1 hour(s) offset
IO\write_error_line('%s / %s offset', $datetime->toString(), $datetime->getTimezoneOffset()->toString());

// let's leave DST
$datetime = $datetime->plusMonths(7);

// output: 31 Oct 2024, 02:59:00 / 0 second(s) offset
IO\write_error_line('%s / %s offset', $datetime->toString(), $datetime->getTimezoneOffset()->toString());
```

```sh
$ php foo.php
31 Mar 2024, 00:59:00 / 0 second(s) offset
31 Mar 2024, 02:59:00 / 1 hour(s) offset
31 Oct 2024, 02:59:00 / 0 second(s) offset
```

the "problem" is that it doesn't remove the offset afterwards, but it should, we could do so maybe?

veewee: Not sure what your question is about. Would you expect the last one to be `1:59 at offset 0`?

I'm really not sure on how it should work without having used it myself. But I remember having stumbled against an issue at some moment where I for example did 'today at 20:00' + 1 week, which resulted in `21:00` or `19:00` which broke my scheduling. (the bug was in a flutter app that was written in [dart](https://medium.com/pinch-nl/datetime-and-daylight-saving-time-in-dart-9c9468633b5d)). That is basically the problem I would not like to face myself. The main edge cases of course lie in the moment of the time-switch from e.g. 3.00 -> 2.00, cause there is 2.00:+1 and 2:00+0 I suppose.

azjezz: I'll test how this works in chrono ( https://docs.rs/chrono/latest/chrono/struct.Date.html ) and other libs, and try to implement the same behavior.

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
azjezz: done

src/Psl/DateTime/DateTime.php
azjezz: agree, we also need to use `OutOfRange` exception instead of InvalidArgumentException here.

will split this check into multiple to throw more clear messages.

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
```diff
@@ -0,0 +1,595 @@
...
+     * Adds the specified years to this date-time object, returning a new instance with the added years.
+     *
+     * @throws Exception\UnderflowException If adding the years results in an arithmetic underflow.
+     * @throws Exception\OverflowException If adding the years results in an arithmetic overflow.
+     *
+     * @psalm-mutation-free
+     */
+    public function plusYears(int $years): static
+    {
+        return $this->plusMonths($years * MONTHS_PER_YEAR);
```
veewee, line 372: I was trying out a test-case that failed on leap day in one of our projects.
It seems to result in a fatal error:

```php
$at = \Psl\DateTime\DateTime::parse('2024-02-29 09:00:00', FormatPattern::SqlDateTime)->minusYears(18);
echo $at->toString(DateStyle::Full, locale: \Psl\Locale\Locale::DutchBelgium) . PHP_EOL;
```

I wanted to test a "is the user eighteen years old today" case on a leap day.



```
PHP Fatal error:  Uncaught ValueError: -1 is not a valid backing value for enum Psl\DateTime\Month in /psl/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php:456
Stack trace:
#0 /psl/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php(456): Psl\DateTime\Month::from(-1)
#1 /psl/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php(385): Psl\DateTime\DateTime->minusMonths(216)
#2 /psl/test.php(14): Psl\DateTime\DateTime->minusYears(18)
#3 {main}
  thrown in /psl/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php on line 456

Fatal error: Uncaught ValueError: -1 is not a valid backing value for enum Psl\DateTime\Month in /psl/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php:456
Stack trace:
#0 /psl/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php(456): Psl\DateTime\Month::from(-1)
#1 /psl/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php(385): Psl\DateTime\DateTime->minusMonths(216)
#2 /psl/test.php(14): Psl\DateTime\DateTime->minusYears(18)
#3 {main}
  thrown in /psl/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php on line 456
```

Instinctively, I first tried to do : `Duration::years(18)` but it seems like Duration stops at weeks.
Does it make sense to add more methods to duration (months, years) - or is it considered not possible because of dynamic days per year?  If so : Maybe the "total*" methods should be based on a timestamp? Like `getTotalSeconds(Timestamp $timestamp);` so that it can deal with these dynamic durations?

That way the implementation for plusMonths could be moved to Timestamp making it always work in exactly the same fashion?

src/Psl/DateTime/Internal/format_rfc3339.php
```diff
@@ -0,0 +1,54 @@
...
+    Timestamp $timestamp,
+    ?SecondsStyle $seconds_style = null,
+    bool $use_z = false,
+    null|Timezone $timezone = null,
+): string {
+    $seconds = $timestamp->getSeconds();
+    $nanoseconds = $timestamp->getNanoseconds();
+
+    if (null === $seconds_style) {
+        $seconds_style = match (true) {
```
veewee: could make sense to move this match expression to a static constructor on SecondsStyle maybe?

src/Psl/DateTime/Duration.php
```diff
@@ -0,0 +1,729 @@
...
+
+    /**
+     * Compiles and returns the duration's components (hours, minutes, seconds, nanoseconds) in an
+     * array, in descending order of significance.
+     *
+     * @return array{int, int, int, int}
+     *
+     * @psalm-mutation-free
+     */
+    public function getParts(): array
```
veewee, line 201: I think parts should go to years here.

src/Psl/DateTime/DateTime.php
```diff
@@ -0,0 +1,552 @@
...
+
+    /**
+     * Adds the specified duration to this date-time object, returning a new instance with the added duration.
+     *
+     * @throws Exception\UnderflowException If adding the duration results in an arithmetic underflow.
+     * @throws Exception\OverflowException If adding the duration results in an arithmetic overflow.
+     *
+     * @psalm-mutation-free
+     */
+    public function plus(Duration $duration): static
```
veewee: where did you draw the line between convenience trait and this class methods?
It somehow looks like this one could e.g. go on the convenience part as well.

src/Psl/DateTime/Timestamp.php
```diff
@@ -0,0 +1,273 @@
...
+
+    /**
+     * Adds the specified duration to this timestamp object, returning a new instance with the added duration.
+     *
+     * @throws Exception\UnderflowException If adding the duration results in an arithmetic underflow.
+     * @throws Exception\OverflowException If adding the duration results in an arithmetic overflow.
+     *
+     * @psalm-mutation-free
+     */
+    public function plus(Duration $duration): static
```
veewee, line 236: I think these should become smarter by extending the Duration and perform the calculations as you do in DateTime. Otherwise this one will suffer from the same issues as you solved in e.g. plusMonths().

src/Psl/DateTime/Duration.php
```diff
@@ -0,0 +1,729 @@
...
+     * @pure
+     */
+    private function __construct(
+        private int $hours,
+        private int $minutes,
+        private int $seconds,
+        private int $nanoseconds
+    ) {
+    }
+
```
veewee, line 50: it could make sense to add a `parse()` method  that takes a format like '+1 hours 30 minutes' that `strtotime()` currently accepts or a format that `DateInterval::createFromDateString('1 day')` accepts. Or maybe also a compatibility layer with `DateInterval`.

src/Psl/DateTime/DateTime.php
```diff
@@ -0,0 +1,552 @@
...
+     *
+     * @psalm-mutation-free
+     */
+    public static function parse(string $raw_string, null|FormatPattern|string $pattern = null, ?Timezone $timezone = null, null|Locale $locale = null): static
+    {
+        $timezone ??= Timezone::default();
+
+        return self::fromTimestamp(Timestamp::parse($raw_string, $pattern, $timezone, $locale), $timezone);
+    }
+
```
veewee, line 311: Currently you have both parse which takes a pattern and fromString which takes some kind of less enforced date/time-style. However it always has to be in some kind of format. Would it make sense to add one more - in line of [strtotime](https://www.php.net/manual/en/function.strtotime.php). The alternative is to use `DateTime::fromTimestamp(Timestamp::fromParts(strtotime(...))` which maybe becomes a bit much? A similar method could be added to `Timestamp`

src/Psl/DateTime/DateTime.php
azjezz: previous because `fromTimestamp` was not part of the interface and was specfic to this class, i think we can move those to the trait now, trying to have as many things there as possible

src/Psl/DateTime/Timestamp.php
azjezz: could you explain further? i don't see what issue might happen here? 🤔

src/Psl/DateTime/Duration.php
azjezz: >'+1 hours 30 minutes' that strtotime() currently accepts or a format that DateInterval::createFromDateString('1 day')

we could 🤔 

>maybe also a compatibility layer with DateInterval

that would be added after this PR as part of the `Interp` component.

src/Psl/DateTime/DateTime.php
azjezz: the reason i didn't add it is that it isn't a think really outside of PHP as far as i seen, most other languages do not have such features, and thinking about it, i couldn't find a case where you would have a datetime string that you are unaware of its format.

src/Psl/DateTime/Duration.php
azjezz: getParts and fromParts here tend to reflect the properties stored internally, we could have a function to return the number of days maybe, but since it is not an internal property, it shouldn't be returned here.

src/Psl/DateTime/Internal/format_rfc3339.php
azjezz: makes sense 👍

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
azjezz: fixed.

azjezz: > Duration::years(18) but it seems like Duration stops at weeks.
Does it make sense to add more methods to duration (months, years)

Yes, that is not possible due to leap years, and months have different length, e.g we don't know if Duration::month(1) should return 28 days, 29, 30, or 31.

azjezz: same for years.

veewee: I tried the fix with leap days:

```
$at = \Psl\DateTime\DateTime::parse('2024-02-29 09:00:00', FormatPattern::SqlDateTime)->minusYears(18);
echo $at->toRfc3339();
```

This results in:

> 2006-03-01T09:00:00+00:00

Which is also what PHP is giving you with regular DateTime.
However,
I'dd rather expect it to be `2006-02-28T09:00:00+00:00` - since march did not start yet.

This is how brick/date-time does it:

https://github.com/brick/date-time/blob/4682c4e1c3cd0b39cce6ac2d6c5eee60de27bde2/src/LocalDate.php#L817-L819

Example:

```php
$today = new DateTimeImmutable('2024-02-29');
$nextYear = $today->modify('+1 year');
echo $nextYear->format('Y-m-d') . PHP_EOL; // 2025-03-01

$today = \Brick\DateTime\LocalDate::fromNativeDateTime($today);
$nextYear = $today->plusYears(1);

echo $nextYear->__toString() . PHP_EOL;  // 2025-02-28
```

(It's based on java's : JSR 310 (Date and Time API))

Java example:

```java
import java.time.*;
import java.time.format.DateTimeFormatter;
import java.time.format.FormatStyle;
import java.time.temporal.ChronoUnit;
import java.time.temporal.TemporalAdjusters;

LocalDate leapDay = LocalDate.of(2024, Month.FEBRUARY, 29);
Period twoYears = Period.ofYears(2);
LocalDate added = leapDay.plus(twoYears);

System.out.println("👋 Hello, " + added);

```
> 👋 Hello, 2026-02-28
(can be tested here : https://dev.java/playground/ )

src/Psl/DateTime/Duration.php
veewee: Just pasting this here:
https://github.com/brick/date-time/blob/324425632ac5933b7253bde65520ca13de850f7a/src/Duration.php#L53-L131

Brick uses this parser for parsing its duration.

src/Psl/DateTime/Timestamp.php
veewee: I think it comes down to this:

Currently, it is only possible to plus() or minus() a duration on Timestamp, but there is no way for adding years, months, ... since this is not part of Duration.

So currently that logic lives inside DateTime (which uses a Timestamp internally). On top of DateTime, there are added extra methods to add years and months specifically. At that location, the additional logic for calculating days in months / years are added. 

So I was thinking : If DateTime uses a Timestamp internally: why not make adding years / months part of timestamp instead so that is possible too. That way, the logic in DateTime could be a proxy to the methods in Timestamp.


Optionally, the Duration could be extended to years / months if the total seconds logic took a Timestamp as parameter to base calculations upon. I also see that's not part of brick-time, so I suppose that is not the best idea there. However it is a bit strange that it's not possible to `plus(Duration::years(1)` but it is possible to `plusYears(1)` from a user's perspective.


Example from java  in which it is possible to specify a period (duration) of 2 years and add that:

```java
import java.time.*;
import java.time.format.DateTimeFormatter;
import java.time.format.FormatStyle;
import java.time.temporal.ChronoUnit;
import java.time.temporal.TemporalAdjusters;

LocalDate leapDay = LocalDate.of(2024, Month.FEBRUARY, 29);
Period twoYears = Period.ofYears(2);
LocalDate added = leapDay.plus(twoYears);

System.out.println("👋 Hello, " + added);
```
(can be ran here https://dev.java/playground/)

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
azjezz: this is really depends on the library/api, there is no standard telling us to go a day forward, or a day back.

I suppose going back a day makes sense as it leaves you within the same month 🤔

src/Psl/DateTime/Timestamp.php
azjezz: we won't be able to do `Duration::years(1)`, the main difference between `Period` in java and `Duration` in psl is that `Duration` could be converted to a total number of seconds ( or milliseconds, nanoseconds .. etc ),. while the `Period` object in java can't, it is mearly a data object, so if you do +500 days on a two year period, it won't be 3 years and 135 days, but 2 years and 500 days, as it won't be able to convert the days to proper months / years without having a reference timestamp. e.g:

```java
import java.time.*;
import java.time.format.DateTimeFormatter;
import java.time.format.FormatStyle;
import java.time.temporal.ChronoUnit;
import java.time.temporal.TemporalAdjusters;


Period period = Period.ofYears(2).withDays(500);

System.out.println("👋 Period: " + period);
```

and the reason we want `Duration` to be convertable to seconds and fractions of seconds is that we want to use it for timeouts and such, `Period` in java cannot be used for this purpose, unless you have a reference point.

---

as for moving the logic of adding/removing years/months to `Timestamp`, that is also not possible, the reason being `Timestamp` could represent a monotomic time, adding / removing years from it makes no sense, as we do not have an actual timestamp reference to decide when a leap year should happen.

veewee: Thanks for the explanation. Maybe a few remarks til I stop about this :) : 

* Brick's date-time package come shipped with `Duration` which is pretty much the same as what we got here. On top of that, it has a `Period` which can be used for date-ranges like years months. There the  difference is made between plusPeriod and plusDuration. it might make sense to prepare the classes for that as well?


*  About the monotomic timestamps : Does it make sense to have both monotomic and reguler Timestamps in the same class? I can imagine if you are doing performance testing, you don't really want to add days to the results there? Cause this would add days to unix epoch most likely?

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
veewee: Most bugs I've seen passing by on leap day were pretty much the example above : adding / subtracting years to leap day resulting in an unexpected day in march. I think it might be a better default. Can't really find much information about why this behaviour was implemented though.

azjezz: case for 28-02: we are still in february, the last day.
case for 01-03: 28 days have already passed in february, we are in the day after the 28th, since its not a leap year, its 01-03.

veewee: It's only an issue if you add a year or more on a leap day though.

So a better question to ask is: what should one expect if you add a year to a leap day, just like how should adding a month on the 31th of January behave

veewee: ... which brings me to this:

```php
$d = new \DateTimeImmutable('2024-01-31 09:00:00');
echo $d->add(\DateInterval::createFromDateString('1 month'))->format('Y-m-d');
```

> 2024-03-02

This is the approach your implementation follows as well:

```php
$at = \Psl\DateTime\DateTime::parse('2024-01-31 09:00:00', FormatPattern::SqlDateTime)->plusMonths(1);
echo $at->toRfc3339();
```

> 2024-03-02T09:00:00+00:00%

**BUT**

Java / brick:
```java
LocalDate leapDay = LocalDate.of(2024, Month.JANUARY, 31);
Period oneMonth = Period.ofMonths(1);
LocalDate added = leapDay.plus(oneMonth);
```

> 2024-02-29


To me, it makes sense to take the java path here as well.

azjezz:
I believe this is enough as is. we could certainly improve it in the future, but i don't see any blockers that prevent us from shipping it for v3.

PR #447, opened 2024-03-23
chore(locale): add `Locale::default()` function to retrieve default locale

azjezz:


src/Psl/Locale/Locale.php
```diff
@@ -791,6 +795,48 @@ enum Locale: string
...
+
+        $locale = Str\lowercase($language);
+        if ($script) {
+            $locale .= '_' . Str\capitalize($script);
+        }
+
+        if ($region) {
+            $locale .= '_' . Str\uppercase($region);
+        }
+
```
azjezz, line 834: wondering if we should do more attempts here, example, if the configured locale is `ar_Arab_TN` ( langauge = ar, script = Arab, region = TN ), we won't be able to match the locale with anything, as we do not have `ar_Arab_TN` in the enum, so we will fallback to English, however, we do have `ar_TN`, or even just `ar`, that we can fallback to instead. this would require us to create multiple combinations of `language+region` ( e.g. `ar_TN` ), `language+script` ( e.g. `zh_Hans` ), `language+script+region` ( e.g `sr_Latn` ) and just `langauge` ( e.g. `en` ) and try matching against them from the most verbose ( `lang+script+region` ) to the least verbose ( `lang` ).

personally i think what we have now is enough, and we can add this feature in the future.

wdyt?

src/Psl/Locale/Locale.php
```diff
@@ -791,6 +795,48 @@ enum Locale: string
...
+        // Attempt to match the system-configured locale with a supported enum instance,
+        // defaulting to English if a precise match is unavailable.
+        return self::tryFrom($locale) ?: self::English;
+    }
```
azjezz, line 838: wondering if we should do more attempts here, example, if the configured locale is `ar_Latn_TN` ( langauge = Arabic, script = Latin, region = Tunisia ), we won't be able to match the locale with anything, as we do not have `ar_Latn_TN` in the enum, so we will fallback to English, however, we do have `ar_TN`, or even just `ar`, that we can fallback to instead.

This would require us to create multiple combinations of `language+region` ( e.g. `ar_TN` ), `language+script` ( e.g. `zh_Hans` ), `language+script+region` ( e.g `sr_Latn` ) and just `langauge` ( e.g. `en` ) and try matching against them from the most verbose ( `lang+script+region` ) to the least verbose ( `lang` ).

personally i think what we have now is enough, and we can add this feature in the future.


wdyt?

azjezz: wondering if we should do more attempts here, example, if the configured locale is `ar_Arab_TN` ( langauge = Arabic, script = Arabic, region = Tunisia ), we won't be able to match the locale with anything, as we do not have `ar_Arab_TN` in the enum, so we will fallback to English, however, we do have `ar_TN`, or even just `ar`, that we can fallback to instead ( note, the default script for `ar` language is `Arab`, so it is extremely unlikely that you would encounter `ar_Arab` or `en_Latn`, as those are the "default" scripts for the language, but, they are still valid ).

This would require us to create multiple combinations of `language+region` ( e.g. `ar_TN` ), `language+script` ( e.g. `zh_Hans` ), `language+script+region` ( e.g `sr_Latn_RS` ) and just `langauge` ( e.g. `en` ) and try matching against them from the most verbose ( `lang+script+region` ) to the least verbose ( `lang` ).

personally i think what we have now is enough, and we can add this feature in the future.


wdyt?

veewee: It might be sufficient indeed.


Maybe an easier approach instead of trying to combine many values, is to: try full, try language, fallback.
Something like:

```php
self::tryFrom($local) ?? self::tryFrom($language) : self::English;
```

That would at least translate into `Local::Arabic` `ar` then?

azjezz: A good idea indeed! we could improve it further later.

PR #451, opened 2024-03-26
feat(tcp): add support for TLS/SSL in TCP

### LINKED ISSUE #342: Add TLS/SSL support
The TCP component is incapable of connecting over TLS/SSL (ex. to any HTTPS site)

The simplest solution would be to add a flag to `Psl\TCP\Connect` which would set the protocol to `tls` (tls requested) or `tcp` (default/no tls requested) here: https://github.com/azjezz/psl/blob/9f9ddfe5eff68d39de7b5a6bf2337fdb77cace47/src/Psl/TCP/connect.php#L32.  However there are also several settings in the SSL context https://www.php.net/manual/en/context.ssl.php so I could see this going into ConnectOptions (or a separate SSLConnectOptions within ConnectOptions).

https://www.php.net/manual/en/function.stream-socket-enable-crypto.php appears to allow you to enable TLS after starting a stream, however I was not able to get it to work with HTTPS, so it looks like there's a slightly different effect of `stream_context_create("tls://")` vs `$stream = stream_context_create("tcp://"); stream_socket_enable_crypto($stream);` - this would allow you to do something like `$stream->enableCrypto();`.

**Additional context**
Running a TLS server is an entirely different beast, because that is going to require loading the certificate in (client certificates also exist but they're far less used 😢).  With the complexity here it might make sense to make an entirely separate TLS-over-TCP module.

azjezz:
closes #342 

- [x] TCP/TLS client support
- [x] TCP/TLS server support


src/Psl/TCP/TLS/Certificate.php
```diff
@@ -0,0 +1,47 @@
...
+     * @param non-empty-string $certificateFile The path to the certificate file.
+     * @param non-empty-string|null $keyFile The path to the private key file associated with the certificate, if any.
+     * @param non-empty-string|null $passphrase The passphrase for the private key file, if the file is encrypted.
+     *
+     * @pure
+     */
+    public function __construct(
+        public readonly string  $certificateFile,
+        public readonly ?string $keyFile = null,
+        public readonly ?string $passphrase = null,
```
veewee, line 30: #[\SensitiveParameter]

azjezz: Ah! we actually have many places where we should add the attribute, mainly `Psl\Password` and `Psl\Hash` components.

PR #456, opened 2024-04-02
Fix PHP84 implicit nullable types

veewee:
See https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

The added cs-fixer rule will enforce that nullable types are always explicitly set.

This PR is targeted to 2.9.x to make that PHP 8.4 compatible as well.
Once merged, it should be merged to next (3.x) as well.


azjezz (approved):

veewee:
@azjezz  shall I merge + tag as 2.9.1 and then merge it back in next?

azjezz:
@veewee yes please 🙏

PR #458, opened 2024-04-03
chore: miscellaneous changes

azjezz:
- marked some classes as `readonly`
- fixes invalid psalm annotations ( mainly related to pure/mutation-free )
- remove the use of promoted properties, this allows us to declare all properties using `camelCase` while still using `snake_case` for arguments.
- declare some classes as immutable/mutation-free
- declare some methods as mutation-free/pure
- declare some functions as mutation-free/pure
- remove redudnant type checks in `Option::unzip` as it is verified using `@psalm-if-this-is`
- introduced `Class\is_readonly` function.

### DIFF
--- a/src/Psl/Class/is_readonly.php
+++ b/src/Psl/Class/is_readonly.php
@@ -1,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Class;
+
+use ReflectionClass;
+
+/**
+ * Checks if class is read only.
+ *
+ * @param class-string $class_name
+ */
+function is_readonly(string $class_name): bool
+{
+    /** @psalm-suppress MissingThrowsDocblock */
+    return (new ReflectionClass($class_name))->isReadOnly();
+}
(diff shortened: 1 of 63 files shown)


src/Psl/Class/is_readonly.php
```diff
@@ -0,0 +1,18 @@
...
+
+/**
+ * Checks if class is read only.
+ *
+ * @param class-string $class_name
+ */
+function is_readonly(string $class_name): bool
+{
+    /** @psalm-suppress MissingThrowsDocblock */
+    return (new ReflectionClass($class_name))->isReadOnly();
```
veewee, line 17: oh, did not know about the internals of the Class component : looks like you were already taking the functional approach there (regarding the reflection component) ;)

veewee (approved):

src/Psl/Class/is_readonly.php
azjezz: `Class`, `Trait`, and `Interface` are considered much simpler than the reflection API. they only provide a handful of fucntions.

Issue #130, opened 2021-02-16
`Psl\Type\non_empty_string()` is not marked as `@psalm-pure`?


Ocramius:
In the context of following snippet:

```php

    /**
     * @throws InvalidPackageName
     *
     * @psalm-pure
     */
    public static function fromName(string $name): self
    {
        if (preg_match('{^[a-z0-9](?:[_.-]?[a-z0-9]+)*/[a-z0-9](?:(?:[_.]?|-{0,2})[a-z0-9]+)*$}iD', $name) !== 1) {
            throw InvalidPackageName::fromInvalidName($name);
        }

        return new self(
            non_empty_string()
                ->coerce($name)
        );
    }
```

psalm reports:

```
... Cannot call an impure function from a mutation-free context (see https://psalm.dev/202)
            non_empty_string()
```

I wonder if all type constructors can/should be declared pure? Is this where we got stuck on conditional type purity?

azjezz:
back to https://github.com/vimeo/psalm/issues/4145 😛 

the issue here is that we can't declare `Type::assert`/`Type::coerce` as "pure", because not all implementations are pure ( for example IterableType::assert iterates over an iterable, which might not be immutable ).

so i suggest you add `@psalm-suppress` here.

azjezz:
Also, i attempted to improve this by having the psalm plugin report the call as pure if the argument is pure/immutable, but that is not possible via the plugin system, and it seems psalm have a special case for internal php functions ( e.g: https://github.com/vimeo/psalm/blob/bd6efd7cf2b7e6fdf01b343be0c5bbb9aacdbbbb/src/Psalm/Internal/Codebase/Functions.php#L499 )

Ocramius:
Wow that's ugly :D 

Guess we'll have to improve upstream before being able to declare it here: still like vimeo/psalm#4145 as an idea, but hard to bring it to life :-\

azjezz:
yea, i will trying to play around with psalm to see if i can make it possible to have special "purity checkers" for function/method calls, this will allow to mark things like `Type\string()->assert()` pure without the need for vimeo/psalm#4145 using the plugin

PR #461, opened 2024-04-05
feat(type): Introduce type constructor purity

veewee:
Partially solves https://github.com/azjezz/psl/issues/130 : The Type\xxx() functions are now considered pure.
`coerce()` and `assert()` can still not be marked as pure for every type (given impure iterable issues)


azjezz:
could we also mark all classes as readonly?

azjezz (approved):
will leave `readonly` for another PR. this is okay on its own.

veewee:
oh yeah I forgot about that ! :) Maybe next week

PR #464, opened 2024-04-06
feat: narrow return type for `min()`, `max()`, `median()` and `mean()`

simPod:
Consider whether the input is empty or not


azjezz (approved):

azjezz:
Thank you @simPod 🙏

PR #465, opened 2024-04-09
feat(type): Notnull type optimizations

### LINKED ISSUE #357: Type\not_null()

Quite often I find myself using `Type\instance_of` to assert values are not null, for example...

```php
/** @var null|Foo $x */
$x = getSomeValueOrAccessAPropertyOrMethodEtc();

$x = Type\instance_of(Foo::class)->assert($x);
```

Here I'm essentially duplicating the expected type (`Foo::class`), because it's usually already declared on the property/method etc...  but is nullable. 

It would be more concise to be able to say...

```php
$x = Type\not_null()->assert($x);
```

Which would assert that `$x` is `Foo` and not `null|Foo`.

I've presented my current solution for this above, I'm not aware of any other features of PSL which could be used here (apologies if I've missed something).

veewee:
Fixes https://github.com/azjezz/psl/issues/357

There is still one situation, where you use `nonnull()` inside a `shape()`.
In this situation, the type will resolve to mixed (inluding null).
This can only be solved once psalm has a `TNonNull` type.

Thanks @devnix !

### DIFF
--- a/src/Psl/Type/Internal/NonNullType.php
+++ b/src/Psl/Type/Internal/NonNullType.php
@@ -18,9 +18,15 @@ use Psl\Type\Exception\CoercionException;
 final readonly class NonNullType extends Type\Type
 {
     /**
-     * @psalm-assert-if-true mixed $value
+     * @template T of mixed
+     *
+     * @param T|null $value
+     *
+     * @psalm-assert-if-true T $value
      *
      * @ara-assert-if-true nonnull $value
+     *
+     * @return ($value is null ? false : true)
      */
     public function matches(mixed $value): bool
     {
@@ -28,9 +34,13 @@ final readonly class NonNullType extends Type\Type
     }
 
     /**
+     * @template T of mixed
+     *
+     * @param T|null $value
+     *
      * @ara-return nonnull
      *
-     * @return mixed
+     * @return ($value is null ? never : T)
      */
     public function coerce(mixed $value): mixed
     {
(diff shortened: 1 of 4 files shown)


src/Psl/Type/Internal/NonNullType.php
```diff
@@ -42,13 +52,17 @@ public function coerce(mixed $value): mixed
...
 
     /**
+     * @template T
+     *
+     * @param T $value
+     *
      * @ara-assert nonnull $value
      *
-     * @psalm-assert mixed $value
+     * @psalm-assert T $value
```
veewee, line 61: The assertion part will also still think it's mixed BTW

azjezz (approved):

azjezz:
I don't like that we are exposing an internal type, but i think its okay in this case.

another option is to introduce a new interface as part of the public API, like `EliminiationTypeInterface` and have nonnull return `EliminiationTypeInterface<null>` .

example:

```hack
interface EliminiationTypeInterface<TEliminated, TGroup> extends TypeInterface<TGroup>
{
    public function matches<TInput: mixed>(
        TInput|TEliminated $value
    ): ($value is TEliminated ? false : true);

    public function coerce<TInput: mixed>(
        TInput|TEliminated $value,
    ): ($value is TEliminated ? never : TGroup);

    public function assert<TInput: mixed>(
        TInput|TEliminated $value,
    ): ($value is TGroup ? ($value is TEliminated ? never : TGroup) : never);
}
```

veewee:
@azjezz That could work - but the interface would be broken by design:

```
extends TypeInterface<TGroup>
```

Since there is no way in psalm for doing an Exclude, it will always tell the upstream that it is of type `TGroup` - yet it is of type `Exclude<TGroup, TEliminated>`.

Not sure if that is a better solution than leaking one internal that most likely wouldn't affect much applications.
I could make a comment on the "why we did this" there ?

PR #466, opened 2024-04-18
feat: narrow ResultInterface to concrete implementation after calling `isSucceeded()` and `isFailed()`

simPod:


tests/static-analysis/Result/result_interface.php
```diff
@@ -0,0 +1,42 @@
...
+use Psl\Result\Failure;
+use Psl\Result\ResultInterface;
+use Psl\Result\Success;
+
+/**
+ * @param ResultInterface<int> $result
+ *
+ * @throws Exception
+ *
+ * @return Failure<int, Exception>
```
veewee, line 17: Stange : Shouldn't psalm complain about the second generic parameter `Exception` that does not exist here?

veewee (commented):
Thanks for the PR. 
Seems like a nice addition!

veewee (approved):

azjezz:
I think we need to revert this 😅 `ResultInterface` is part of the public API, and while it is quite unlikey, someone could re-implement `Success`/`Failure`, e.g `MySuccess`/`MyFailure`, in which case, `MySuccess::isSucceeded` would return true, but that does not mean its an instance of `Success` ( that is what psalm would think ).

repro:

```
function foo(ResultInterface $result): Success {
  if ($result->isSucceeded()) {
    return $result;
  }

  exit('no');
}

foo($mySuccessImplementation); // throws an exception
```

[*CRIES*](https://wiki.php.net/rfc/sealed_classes)

simPod:
Hm, that's true. 😢<sup>2</sup>

Should I send revert PR?

veewee:
Woops, merged a bit too fast - sorry.
A revert PR is welcome. 😢 **3

simPod:
There u go https://github.com/azjezz/psl/pull/469

simPod:
BTW is there any chance the RFC is accepted in the future? (like it was missing something that could be fixed)

PR #467, opened 2024-04-18
feat: introduce `first_opt()`, `first_key_opt()`, `last_opt()`, `last_key_opt()` and `search_opt()`

simPod:
I'd like to modify search operations over `iterables` by wrapping the result in Option.

Currently, it is not possible to use these functions to find `null` value since `null` represents "no result".

This applies to all `first*`, `last*` and `search` functions.

WDYT?


azjezz:
I like the idea of returning an `Option`, but i dislike having this type of BC, what about adding new functions that result in an `Option`? e.g `first_opt<T>(iterable<T> $i): Option<T>` vs `first<T>(iterable<T> $I): null|T`?

`_opt` suffix is one way to go about this ( this is a popular choice in the rust community, e.g: https://docs.rs/chrono/latest/chrono/struct.Date.html#method.and_hms_micro VS https://docs.rs/chrono/latest/chrono/struct.Date.html#method.and_hms_micro_opt), but i'm open to suggestions if you have any, maybe we can find a more suitable naming convention for those functions.

cc @veewee this might interest you.

simPod:
I'll go for it. Thanks for input.

veewee:
@azjezz 

The migration path might be better this way indeed.
However you'll probably end up to something that looks like this in the long run, which might be something we'dd want to avoid?

* v3 -> deprecate old API in favour for `_opt`
* v4 -> remove old API
* v5 -> Recreate original function without the `_opt` suffix and deprecate the `_opt` suffix functions
* v6 -> remove the `_opt` functions again

azjezz:
@veewee personally i would like to keep both, i don't see why we need to deprecate/remove the old behaviour.

src/Psl/Iter/last_key_opt.php
```diff
@@ -0,0 +1,28 @@
...
+ * if the iterable is empty, {@see Option::none} will be returned.
+ *
+ * @template Tk
+ * @template Tv
+ *
+ * @param iterable<Tk, Tv> $iterable
+ *
+ * @return Option<Tk>
+ */
+function last_key_opt(iterable $iterable)
```
azjezz:
```suggestion
function last_key_opt(iterable $iterable): Option
```

src/Psl/Iter/last_key_opt.php
```diff
@@ -0,0 +1,28 @@
...
+{
+    $last = Option::none();
+    foreach ($iterable as $k => $_) {
+        $last = $k;
+    }
+
+    return $last instanceof Option ? $last : Option::some($last);
```
azjezz: this breaks static analysis.

if the given type is `iterable<string, Option<bool>>` psalm assume the return type is `Option<Option<bool>>`, you are breaking that here by making it `Option<bool>`. meaning, regardless of the type of `$last` it should be returned as is.

```suggestion
    $last = Option::none();
    foreach ($iterable as $k => $_) {
        $last = Option::some($k);
    }

    return $last;
```

src/Psl/Iter/last_opt.php
```diff
@@ -0,0 +1,27 @@
...
+ * @return Option<Tv>
+ */
+function last_opt(iterable $iterable)
+{
+    $last = Option::none();
+    foreach ($iterable as $v) {
+        $last = $v;
+    }
+
+    return $last instanceof Option ? $last : Option::some($last);
```
azjezz: same as above.

azjezz (changes requested):

src/Psl/Iter/last_key_opt.php
simPod: true. i did not want to initialize Option for each value but you're right

src/Psl/Iter/last_opt.php
```diff
@@ -0,0 +1,27 @@
...
+ * Returns the last element of an iterable wrapped in {@see Option::some},
+ * if the iterable is empty, {@see Option::none} will be returned.
+ *
+ * @template Tv
+ *
+ * @param iterable<Tv> $iterable
+ *
+ * @return Option<Tv>
+ */
+function last_opt(iterable $iterable)
```
azjezz: missing return type.
```suggestion
function last_opt(iterable $iterable): Option
```

src/Psl/Iter/search_opt.php
```diff
@@ -0,0 +1,39 @@
...
+ *      => Null
+ *
+ * @template T
+ *
+ * @param iterable<T> $iterable The iterable to search
+ * @param (Closure(T): bool) $predicate
+ *
+ * @return Option<T>
+ */
+function search_opt(iterable $iterable, Closure $predicate)
```
azjezz: same

azjezz (changes requested):

src/Psl/Iter/search_opt.php
simPod:
```suggestion
function search_opt(iterable $iterable, Closure $predicate): Option
```

src/Psl/Iter/search_opt.php
```diff
@@ -0,0 +1,39 @@
...
+ * the value of the matching element wrapped in {@see Option::some}.
+ * If a predicate never returns true, {@see Option::none} will be returned.
+ *
+ * Examples:
+ *
+ *      Iter\search(['foo', 'bar', 'baz'], fn($v) => 'baz' === $v)
+ *      => Str('baz')
+ *
+ *      Iter\search(['foo', 'bar', 'baz'], fn($v) => 'qux' === $v)
+ *      => Null
```
azjezz: exampless are outdated

azjezz (changes requested):
one last thing 😅

src/Psl/Iter/search_opt.php
simPod: good catch. is it correct now? Not sure what `Str('baz')` was, replaced.

azjezz (approved):

veewee (approved):
LGTM. Thanks!

PR #470, opened 2024-04-19
feat(Result): introduce `Result::unwrapOr()`

simPod:
I'd like to introduce a function that allows to get inner value from Result if success and allows to bypass throwing an exception from Failure by providing a default value.

### DIFF
--- a/src/Psl/Result/Failure.php
+++ b/src/Psl/Result/Failure.php
@@ -45,6 +45,20 @@ final readonly class Failure implements ResultInterface
     }
 
     /**
+     * Unwrap the Result if it is succeeded or return $default value.
+     *
+     * @param D $default
+     *
+     * @return T|D
+     *
+     * @template D
+     */
+    public function unwrapOr(mixed $default): mixed
+    {
+        return $default;
+    }
+
+    /**
      * Since this is a failed result wrapper, this always returns the `Throwable` thrown during the operation.
      *
      * @return Te - The `Throwable` thrown during the operation.
--- a/src/Psl/Result/Failure.php
+++ b/src/Psl/Result/Failure.php
@@ -47,11 +47,11 @@ final readonly class Failure implements ResultInterface
     /**
      * Unwrap the Result if it is succeeded or return $default value.
      *
+     * @template D
+     *
      * @param D $default
      *
-     * @return T|D
-     *
-     * @template D
+     * @return D
      */
     public function unwrapOr(mixed $default): mixed
     {
--- a/src/Psl/Result/ResultInterface.php
+++ b/src/Psl/Result/ResultInterface.php
@@ -96,6 +96,17 @@ interface ResultInterface extends Psl\Promise\PromiseInterface
     public function getResult();
 
     /**
+     * Unwrap the Result if it is succeeded or return $default value.
+     *
+     * @param D $default
+     *
+     * @return T|D
+     *
+     * @template D
+     */
+    public function unwrapOr(mixed $default): mixed;
+
+    /**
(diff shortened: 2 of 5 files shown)


src/Psl/Result/unwrapResultOr.php
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * @param ResultInterface<T> $r
+ * @param F $f
+ *
+ * @return T|F
+ *
+ * @template T
+ * @template F
+ */
+function unwrapResultOr(ResultInterface $r, mixed $f)
```
veewee: It would be better to introduce `ResultInterface::unwrapOr(mixed $default): mixed` instead of adding a function. This would be in line with how the `Option` class does it.

veewee (changes requested):

src/Psl/Result/unwrapResultOr.php
azjezz: Agree here. this should be a method in `ResultInterface`.

side note: we use `snake_case` for functions :)

src/Psl/Result/Failure.php
```diff
@@ -44,6 +44,11 @@ public function getResult(): void
         throw $this->throwable;
     }
 
+    public function unwrapOr(mixed $default)
```
veewee: Can you add the (inherited) docblock types here as well?
Also the `: mixed` php return type.


(this should be done on the interface and the 2 implementing classes)

simPod: I considered those docblock implicit in implementations but I see it's copied everywhere. Done.

src/Psl/Result/Failure.php
```diff
@@ -44,6 +44,20 @@ public function getResult(): void
...
     }
 
+    /**
+     * Unwrap the Result if it is succeeded or return $default value.
+     *
+     * @param D $default
+     *
+     * @return T|D
+     *
+     * @template D
```
azjezz: This function always returns the default.

```suggestion
     * @template D
     *
     * @param D $default
     *
     * @return D
 ```

File: src/Psl/Result/ResultInterface.php [hunk 4]
```diff
@@ -95,6 +95,17 @@ public function always(Closure $always): ResultInterface;
...
 
+    /**
+     * Unwrap the Result if it is succeeded or return $default value.
+     *
+     * @param D $default
+     *
+     * @return T|D
+     *
+     * @template D
+     */
```
azjezz: template comes first

```suggestion
     * Unwrap the Result if it is succeeded or return $default value.
     *
     * @template D 
     *
     * @param D $default
     *
     * @return T|D
     */
```

src/Psl/Result/Success.php
```diff
@@ -44,6 +44,20 @@ public function getResult(): mixed
...
+     * @param D $default
+     *
+     * @return T|D
+     *
+     * @template D
+     */
+    public function unwrapOr(mixed $default): mixed
+    {
+        return $this->value;
+    }
```
azjezz: this always returns `T`.

```suggestion
     * @template D
     *
     * @param D $default
     *
     * @return T
     */
    public function unwrapOr(mixed $default): mixed
    {
        return $this->value;
    }
```

azjezz (changes requested):
just a few changes needed :)

src/Psl/Result/Failure.php
simPod: thx for review, good points

veewee (approved):
LGTM

azjezz (approved):

PR #477, opened 2024-06-07
feat(DateTime) Provide a human-like months substract and add system

veewee:
As mentioned in 

* https://github.com/azjezz/psl/pull/446#discussion_r1558855675
* https://github.com/azjezz/psl/pull/446#discussion_r1560977706

This PR introduces a way of adding and subsctracting months that is more in line with how you would do it as a human, rather than a computer.

I know you told me a few times:

> The behaviour there is still the same, i personally think it is okay. its not broken. as there is no "standard" way of doing things here, every lib / lang does something that is completely different

However, I hope you consider taking this approach when it's just a "merge" button click away :)

### DIFF
--- a/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
+++ b/src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
@@ -362,8 +362,7 @@ trait DateTimeConvenienceMethodsTrait
     /**
      * Adds the specified years to this date-time object, returning a new instance with the added years.
      *
-     * @throws Exception\UnderflowException If adding the years results in an arithmetic underflow.
-     * @throws Exception\OverflowException If adding the years results in an arithmetic overflow.
+     * @throws Exception\UnexpectedValueException If adding the years results in an arithmetic issue.
      *
      * @psalm-mutation-free
      */
@@ -375,8 +374,7 @@ trait DateTimeConvenienceMethodsTrait
     /**
      * Subtracts the specified years from this date-time object, returning a new instance with the subtracted years.
      *
-     * @throws Exception\UnderflowException If subtracting the years results in an arithmetic underflow.
-     * @throws Exception\OverflowException If subtracting the years results in an arithmetic overflow.
+     * @throws Exception\UnexpectedValueException If subtracting the years results in an arithmetic issue.
      *
      * @psalm-mutation-free
      */
@@ -388,10 +386,11 @@ trait DateTimeConvenienceMethodsTrait
     /**
      * Adds the specified months to this date-time object, returning a new instance with the added months.
      *
-     * @throws Exception\UnderflowException If adding the months results in an arithmetic underflow.
-     * @throws Exception\OverflowException If adding the months results in an arithmetic overflow.
+     * @throws Exception\UnexpectedValueException If adding the months results in an arithmetic issue.
      *
      * @psalm-mutation-free
+     *
+     * @psalm-suppress MissingThrowsDocblock - The Math exceptions from Math\div do not result in any error.
      */
     public function plusMonths(int $months): static
     {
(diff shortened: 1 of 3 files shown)


veewee:
(The mutation tests nor failing windows unit test seem related to this PR)

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
```diff
@@ -7,6 +7,9 @@
 use Psl\Locale\Locale;
 use Psl\Math;
 
+use function intdiv;
+use function min;
+
```
azjezz: should use `Math\minva` and `Math\div` instead.

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
```diff
@@ -403,28 +403,29 @@ public function plusMonths(int $months): static
...
+            $target_month = $target_month - 12;
         }
 
-        return $this->plus(Duration::days($days_to_add));
+        return $this->withDate(
+            $target_year = $this->getYear() + $plus_years,
+            $target_month = Month::from($target_month)->value,
+            min([
+                $this->getDay(),
+                Month::from($target_month)->getDaysForYear($target_year)
```
azjezz: we could create the enum in the line before to avoid having two calls to `Month::from`

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
```diff
@@ -438,25 +439,23 @@ public function minusMonths(int $months): static
...
+            $target_month = 12 - Math\abs($target_month);
         }
 
-        return $this->minus(Duration::days($days_to_subtract));
+        return $this->withDate(
+            $target_year = $this->getYear() - $minus_years,
+            $target_month = Month::from($target_month)->value,
+            min([
+                $this->getDay(),
+                Month::from($target_month)->getDaysForYear($target_year)
```
azjezz: same here

azjezz (changes requested):
LGTM! just nitpicking 😅

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
veewee: added `minva`

However `div` throws:

>  * throws Exception\ArithmeticException If the $numerator is Math\INT64_MIN and the $denominator is -1.
> * throws Exception\DivisionByZeroException If the $denominator is 0.

Which is not true, cause de denominator is always 12. 
Because I don't want to be leaking those exceptions on these methods (cause they don't make sense), I just used `\intdiv` directly.

azjezz: @psalm-supress is your friend :P

azjezz (approved):

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
```diff
@@ -438,25 +442,25 @@ public function minusMonths(int $months): static
...
-                $current_year--; // Adjust year when wrapping
-            }
-
-            $target_month = ($total_months % MONTHS_PER_YEAR) ?: MONTHS_PER_YEAR;
-            $target_year = $current_year + Math\div($total_months - 1, MONTHS_PER_YEAR);
-
-            // Subtract days of the month we are moving into
-            $days_to_subtract += Month::from($target_month)->getDaysForYear($target_year);
+        $minus_years = Math\div($months, 12);
+        $months_left = $months - ($minus_years * 12);
```
azjezz, line 446: todo: `12` should be replaced with `MONTHS_PER_YEAR` ( just in case humanity decided to change the calender, so we have less work todo 😛 )

PR #478, opened 2024-06-09
Move nonnull type to public namespace

veewee:
Currently, the hack in the nonnull type results in a psalm issue when it's being used:

nonnullissue

This PR moves the NonNull type out of the internal namespace to overcome this problem.

### DIFF
--- a/src/Psl/Type/nonnull.php
+++ b/src/Psl/Type/nonnull.php
@@ -4,21 +4,19 @@ declare(strict_types=1);
 
 namespace Psl\Type;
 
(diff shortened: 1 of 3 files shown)


src/Psl/Type/nonnull.php
```diff
@@ -17,8 +15,8 @@
  */
 function nonnull(): TypeInterface
```
azjezz:
```suggestion
function nonnull(): NonNull
```

shouldn't this be updated?

azjezz (approved):

PR #482, opened 2024-07-01
feat(collections): introduce `Set`, `SetInterface`, `MutableSet`, and `MutableSetInterface`

azjezz:
introduce the following new types:

- `Set`
- `MutableSet`
- `SetInterface`
- `MutableSetInterface`

`map` and `mapWithKeys` methods have been removed from collection interfaces, they are now present only in `Map`, `MutableMap`, `MapInterface`, `MutableMapInterface`, `Vector`, `MutableVector`, `VectorInterface`, `MutableVectorInterface`, `Set`, `MutableSet`, `SetInterface`, and `MutableSetInterface`. ( the reason is that both map and vector map using the signature `map<U>(c: f(V): U) -> Self<K, U>` while sets map using `map<U>(c: f(V): U) -> Self<U, U>` signature ).


azjezz:
@veewee i need `Set`/`MutableSet` in another project, hope the change regarding the other interfaces does not break anything for you :)

azjezz:
note: this change will be part of v3, but in v4, I plan on removing the current `Psl\Collection` component, and replacing it with a new implementation that is more rust-ish ( see https://doc.rust-lang.org/std/collections/index.html ) that makes use of `Psl\Option` and `Psl\Result`, as the current implementation was initially supposed to be ( a sort of still is ) 
 a PHP port `HH\` collections in HHVM ( https://docs.hhvm.com/hack/arrays-and-collections/object-collections )

src/Psl/Collection/MutableAccessibleCollectionInterface.php
```diff
@@ -4,11 +4,11 @@
 
 namespace Psl\Collection;
 
+use ArrayAccess;
 use Closure;
 
 /**
- * The base interface implemented for a collection type that you are able set and remove its values.
- * keys.
+ * The base interface implemented for a collection type that you are able remove its values.
```
joshrai: This still supports setting, right? Not just removing?
```suggestion
 * The base interface implemented for a collection type whose values you are able to set and remove.
```

src/Psl/Collection/MutableSet.php
```diff
@@ -0,0 +1,716 @@
...
+     * @param array<T, mixed> $elements
+     *
+     * @psalm-mutation-free
+     */
+    public function __construct(array $elements)
+    {
+        $set = [];
+        foreach ($elements as $key => $_) {
+            $set[$key] = $key;
+        }
```
joshrai: Per our conversation offline, I feel it would be more intuitive to use the values of `$elements`, rather than the keys, to construct a MutableSet – something like this:
```suggestion
     * @param array<array-key, T> $elements
     *
     * @psalm-mutation-free
     */
    public function __construct(array $elements)
    {
        $set = [];
        foreach ($elements as $_ => $value) {
            $set[$value] = $value;
        }
```

This would also have implications for other method implementations below, like `map()`.

src/Psl/Collection/MutableSet.php
```diff
@@ -0,0 +1,716 @@
...
+
+    /**
+     * Returns a `MutableSet` after an operation has been applied to each value
+     * in the current `MutableSet`.
+     *
+     * Every value in the current Map is affected by a call to `map()`, unlike
+     * `filter()` where only values that meet a certain criteria are affected.
+     *
+     * The keys will remain unchanged from the current `MutableSet` to the
+     * returned `MutableSet`.
```
joshrai, line 431: From the point of view of the implementation (the private `elements` array), I think this isn't true; and more abstractly, "key" doesn't really have a clear meaning, because a set doesn't exactly have keys. I think the comments in the HHVM codebase are helpful:
https://github.com/facebook/hhvm/blob/HHVM-3.27.8/hphp/hack/hhi/collections/interfaces.hhi#L1447-L1465

src/Psl/Collection/SetInterface.php
```diff
@@ -0,0 +1,329 @@
...
+
+/**
+ * @template T of array-key
+ *
+ * @extends AccessibleCollectionInterface<T, T>
+ */
+interface SetInterface extends AccessibleCollectionInterface
+{
+    /**
+     * Returns the value at the specified key in the current set.
```
joshrai: Just a nit, similar to my comment in MutableSet::map(), but I think that abstractly, it makes sense to describe a set as having only values, not keys. It does mean that some of these methods are a little awkward to describe or don't have much value (and are being defined just for consistency with other interfaces or conventions), but it might still be clearer to say that keys don't exist, conceptually.

tests/unit/Collection/AbstractSetTest.php
```diff
@@ -0,0 +1,534 @@
...
+     * @var class-string<SetInterface>
+     */
+    protected string $setClass = SetInterface::class;
+
+    public function testIsEmpty(): void
+    {
+        static::assertTrue($this->default()->isEmpty());
+        static::assertTrue($this->createFromKeys([])->isEmpty());
+        static::assertFalse($this->createFromKeys(['foo', 'bar'])->isEmpty());
+        static::assertEmpty($this->createFromKeys([null])->isEmpty());
```
joshrai: Should this be `assertFalse` instead of `assertEmpty`?

tests/unit/Collection/AbstractSetTest.php
joshrai, line 1: Not sure where the best place would be to put this, but do you think it would be helpful to have a test related to type juggling? E.g., show that `$this->createFromKeys([123])->contains('123')` throws an exception? (Actually, I'm not 100% sure that it does, but I think so. Either way, have a test to clarify the expectations.)

joshrai (commented):
Looks good! Just left a few minor comments.

src/Psl/Collection/MutableAccessibleCollectionInterface.php
azjezz: indeed, the comment was modified after removing `set()` methods from the interface, but before adding `ArrayAccess`.

src/Psl/Collection/SetInterface.php
```diff
@@ -0,0 +1,329 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Collection;
+
+use Closure;
+
+/**
+ * @template T of array-key
```
veewee, line 10: Limiting to `array-key` does not seem like a good idea to me.

In JS, you can use a `Set` to store anything.
See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#value_equality

```
> new Set([{}, {}, {}]);
Set(3) { {}, {}, {} }
```

src/Psl/Type/vector.php
```diff
@@ -9,7 +9,7 @@
 /**
  * @psalm-pure
  *
- * @template T
+ * @template T of array-key
```
veewee: vector should not of type `array-key` right?

src/Psl/Type/Internal/MutableSetType.php
```diff
@@ -0,0 +1,120 @@
...
+            $key_type = $this->key_type;
+            /** @var list<T> $keys */
+            $keys = [];
+            $key = null;
+            $iterating = true;
+            try {
+                /**
+                 * @var T $key
+                 */
+                foreach ($value as $key => $_) {
```
veewee: Using the key doesn't feel very natural here.
Wouldnt it make more sense to use the values instead? If one wanted to use the keys, they could first grab them and pass those along to the type? (same for the other type)

src/Psl/Type/Internal/MutableSetType.php
```diff
@@ -0,0 +1,120 @@
...
```
veewee: also here it might make more sense to assert the values instead of the key.
(assuming we'dd allow something broader than array-key to be contained in the set)

src/Psl/Type/Internal/MutableSetType.php
```diff
@@ -0,0 +1,120 @@
...
+                    $iterating = false;
+                    $keys[] = $key_type->assert($key);
+                    $iterating = true;
+                }
+            } catch (Throwable $e) {
+                throw match (true) {
+                    $iterating => AssertException::withValue(null, $this->toString(), PathExpression::iteratorError($key), $e),
+                    default => AssertException::withValue($key, $this->toString(), PathExpression::iteratorKey($key), $e),
```
veewee: For other types, we are using `PathExpression::path` to indicate where things go wrong. 
(assuming we wont be using the key to identify)

src/Psl/Collection/SetInterface.php
veewee: I agree : the keys are not relevant for a Set. They are rather a unique identifier to assure the set values are unique.

src/Psl/Collection/SetInterface.php
```diff
@@ -0,0 +1,329 @@
...
+     *
+     * @param positive-int $size The size of each chunk.
+     *
+     * @return VectorInterface<static<T>> A `VectorInterface` containing the original
+     *                                    `SetInterface` split into chunks of the given size.
+     *
+     * @psalm-mutation-free
+     */
+    public function chunk(int $size): VectorInterface;
+}
```
veewee, line 337: Maybe some other usefull methods that could be added (or in another PR)

* diff
* union
* intersect

veewee (commented):
Looks like a good place to start from.
I don't think the changes will break anything from my end, since I'm just using the provided classes instead of interfaces.

Added some comments here and there.

src/Psl/Type/vector.php
azjezz: nice catch!

src/Psl/Collection/SetInterface.php
azjezz: the problem is that we do not have a simple ( and more importantly fast ) way to determine what is unique.

I don't know how `Set` in js determines what is unique ( it looks like all `{}` are present, so not sure if it even does so? ), in [rust](https://doc.rust-lang.org/std/collections/struct.HashSet.html), this is done by requiring all elements to implement `Hash` https://doc.rust-lang.org/std/hash/trait.Hash.html and calculating+comparing the hash of the element.

we have 3 ways to do this in psl.

1. support only array-key, this way we can simply make `Set<T>` `array<T, T>` internally, and have PHP handle "uniqueness" for us ( by far the fastest implementation )
2. support only scalars, and use `Psl\Vec\unique_scalar`: this will make the internal table `list<T>`, it is slower than the first solution, but still fast.
3. support all types, and use `Psl\Vec\unique`: this is the slowest, and because of how slow `Psl\Vec\unique` was, we introduced `Psl\Vec\unique_scalar` a while back.

veewee: In JavaScript, objects are compared per reference and not by value.
It is using the SameValueZero comparison:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#comparing_equality_methods

So I suppose it is using something like Vec\unique internally - but only when appending items. So the slowness is in adding, not in the other methods. I don't really know how slow that acutally is and how many items you are putting in your Set, but in my cases the amount of items in a set is not considered a lot meaning it won't be very slow. But thats depending on the amount of items.

If you are aiming on big data sets, something like a unique hash might be the fastest indeed.
We could create some internal function that does something like this to generate a unique ID:

```php
assert(hash('1') === 'string<1>');
assert(hash(1) === 'int<1>');
assert(hash($object) === 'object<spl_object_hash()>');
assert(hash($array) === 'array<recursive_md5()>'); // not sure what to do here

// ...
```

(of course could be an actual hash rather than this representation)

src/Psl/Collection/MutableSet.php
```diff
@@ -0,0 +1,716 @@
...
+     *
+     * @psalm-external-mutation-free
+     *
+     * @psalm-assert null|array-key $offset
+     *
+     * @throws Exception\InvalidOffsetException If the offset is not null or the value is not the same as the offset.
+     */
+    public function offsetSet(mixed $offset, mixed $value): void
+    {
+        if (null === $offset || $offset === $value) {
```
joshrai, line 688: Per the earlier discussion about keys not really being a valid concept of sets, we might even consider disallowing the use of non-null offsets, so `$mutableSet[] = 123` is legal but `$mutableSet[123] = ...` is illegal, regardless of what `...` is.

tests/unit/Collection/AbstractSetTest.php
azjezz: we do not include tests for type juggling, `$this->createFromKeys([123])->contains('123')` will result in an error in Psalm, and we consider any use of Psl that fails static analysis to be undefined behaviour. It makes things easier here :)

src/Psl/Collection/SetInterface.php
azjezz: I think we should keep it as is for now, the use case i personally have for it does not require support for anything beyond int|string, we could add support for other types later on in v3. but as mentioned, i plan to remove the current `Psl\Collection` component in v4 and re-write it from scratch to align more with Rust `std::collections`.

src/Psl/Collection/MutableSet.php
azjezz: I think it is okay to keep support for this, it does not cause any problems here.

src/Psl/Type/Internal/MutableSetType.php
azjezz: updated.

src/Psl/Type/mutable_set.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @psalm-pure
+ *
+ * @template T of array-key
+ *
+ * @param TypeInterface<T> $key_type
+ *
+ * @return TypeInterface<Collection\MutableSetInterface<T>>
+ */
+function mutable_set(TypeInterface $key_type): TypeInterface
```
veewee: `$type`or `$value_type`
This one also requires adding documentation to the README.md

src/Psl/Type/set.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @psalm-pure
+ *
+ * @template T of array-key
+ *
+ * @param TypeInterface<T> $key_type
+ *
+ * @return TypeInterface<Collection\SetInterface<T>>
+ */
+function set(TypeInterface $key_type): TypeInterface
```
veewee: `$type`or `$value_type`
This one also requires adding documentation to the README.md

veewee (approved):

PR #483, opened 2024-08-01
Add value-of<BackedEnum> type

gsteel:
Adds a type that can assert `value-of<BackedEnum>`

If this is an acceptable improvement - perhaps an `EnumCase` type too?

There are some SA issues with the template `@extends Type<value-of<T>>` that I'm struggling to fix - help appreciated!

### DIFF
--- a/src/Psl/Type/Internal/BackedEnumValueType.php
+++ b/src/Psl/Type/Internal/BackedEnumValueType.php
@@ -1,0 +1,134 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Type\Internal;
+
+use BackedEnum;
+use Psl\Exception\InvariantViolationException;
+use Psl\Exception\RuntimeException;
+use Psl\Type\Exception\AssertException;
+use Psl\Type\Exception\CoercionException;
+use Psl\Type\Type;
+use ReflectionEnum;
+use ReflectionException;
+use ReflectionNamedType;
+
+use function is_a;
+use function is_int;
+use function is_string;
+use function Psl\invariant;
+use function Psl\Type\int;
+use function Psl\Type\string;
+
+/**
+ * @template T of BackedEnum
+ *
+ * @extends Type<value-of<T>>
+ *
+ * @internal
+ */
+final readonly class BackedEnumValueType extends Type
+{
+    private bool $isStringBacked;
+
+    /**
+     * @psalm-mutation-free
+     *
+     * @param class-string<T> $enum
+     *
+     * @throws RuntimeException If reflection fails.
+     * @throws InvariantViolationException If the given value is not class-string<BackedEnum>.
+     */
+    public function __construct(
+        private string $enum
+    ) {
+        $this->isStringBacked = $this->hasStringBackingType($this->enum);
+    }
+
+    /**
+     * @param class-string<T> $enum
+     *
+     * @throws RuntimeException If reflection fails.
+     * @throws InvariantViolationException If the given value is not class-string<BackedEnum>.
+     */
+    private function hasStringBackingType(string $enum): bool
+    {
+        invariant(is_a($enum, BackedEnum::class, true), 'A BackedEnum class-string is required');
+
+        // If the enum has any cases, detect its type by inspecting the first case found
+        $case = $enum::cases()[0] ?? null;
+        if ($case !== null) {
+            return is_string($case->value);
+        }
+
+        // Fallback to reflection to detect the backing type:
+        try {
+            $reflection = new ReflectionEnum($enum);
+            $type = $reflection->getBackingType();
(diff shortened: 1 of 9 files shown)


azjezz:
LGTM, but there seems to be a plsam failure.

src/Psl/Type/backed_enum_value.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * @psalm-pure
+ *
+ * @template T of BackedEnum
+ *
+ * @param class-string<T> $enum
+ *
+ * @return TypeInterface<value-of<T>>
+ */
+function backed_enum_value(string $enum): TypeInterface
```
veewee, line 23: You can document this new function in https://github.com/azjezz/psl/blob/next/src/Psl/Type/README.md

veewee (dismissed):

gsteel:
Psalm can't seem to deal with `value-of<T>` properly, but it _does_ infer the type

https://psalm.dev/r/bdb7fd39f0

Should the value be forced in the code with `/** @psalm-var */` or is the baseline prefereable?

gsteel:
Incidentally, PHPStan will pass this if you use something like `(value-of<T>&int)|(value-of<T>&string)`, but psalm really hates that because it only allows intersections of objects

https://phpstan.org/r/971caf90-6364-436e-811e-76f81be900f5

azjezz:
> forced in the code with /** @psalm-var */ 

Yes, that is okay.

gsteel:
I've added `@psalm-suppress` - forcing the type creates more issues and at least the suppressions will surface as `UnusedPsalmSuppress` in future psalm updates.

src/Psl/Type/Internal/BackedEnumValueType.php
```diff
@@ -0,0 +1,111 @@
...
+    /**
+     * @psalm-assert-if-true value-of<T> $value
+     */
+    public function matches(mixed $value): bool
+    {
+        if (! is_string($value) && ! is_int($value)) {
+            return false;
+        }
+
+        $values = array_map(
```
BackEndTea: Not sure about performance, but can't we simplify this code by doing:

```php 
return $this->enum::tryFrom($value) !== null;
```

gsteel: Yes. It's a `TypeError` to pass int to `tryFrom` for a string-backed enum and vice-versa, so I'll need to consult `$this->isStringBacked`

src/Psl/Type/Internal/BackedEnumValueType.php
```diff
@@ -0,0 +1,101 @@
...
+    ) {
+        $this->isStringBacked = is_string($this->enum::cases()[0]->value);
+    }
+
+    /**
+     * @psalm-assert-if-true value-of<T> $value
+     */
+    public function matches(mixed $value): bool
+    {
+        if (! is_string($value) && ! is_int($value)) {
```
BackEndTea: I think this if statement can e removed entirely, since the string/int checks are done as part of the match

src/Psl/Type/Internal/BackedEnumValueType.php
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * @psalm-mutation-free
+     *
+     * @param class-string<T> $enum
+     */
+    public function __construct(
+        private string $enum
+    ) {
+        $this->isStringBacked = is_string($this->enum::cases()[0]->value);
```
BackEndTea: This code will produce warnings if the enum has no values.

e.g.

```php
enum Bar: string {}

\Psl\Type\backed_enum_value(Bar::class);
```

Not sure if we want to validate against that, or if its not worth it.

azjezz: we should use reflections here.

gsteel: I didn't think it was worth the penalty of using reflection, but I've changed it to reflect the enum now 👍

BackEndTea: Doing some quick checks on my machine, reflection is a factor 3 slower. Maybe we could fall back to it only if there are no cases?

gsteel: OK, I've extracted detection to a private method to make the purpose of the added code obvious and added additional tests to cover the detection logic.

gsteel: I'm also unsure how to force a reflection failure here to test the uncovered lines. Any ideas?

BackEndTea (commented):
LGTM 👍

veewee (approved):

PR #485, opened 2024-08-30
Fix mutation coverage

veewee:
This PR fixes the infection errors in CI (related to the DateTime component)

### DIFF
--- a/config/infection.json.dist
+++ b/config/infection.json.dist
@@ -32,7 +32,8 @@
         },
         "DecrementInteger": {
             "ignore": [
-                "Psl\\DataStructure\\PriorityQueue::peek"
+                "Psl\\DataStructure\\PriorityQueue::peek",
+                "Psl\\DateTime\\TemporalConvenienceMethodsTrait::since"
             ]
         },
         "FunctionCallRemoval": {
@@ -42,7 +43,8 @@
         },
         "IncrementInteger": {
             "ignore": [
-                "Psl\\DataStructure\\PriorityQueue::peek"
+                "Psl\\DataStructure\\PriorityQueue::peek",
+                "Psl\\DateTime\\TemporalConvenienceMethodsTrait::since"
             ]
         },
         "LogicalNot": {
@@ -58,6 +60,7 @@
         },
         "Throw_": {
             "ignore": [
+                "Psl\\DateTime\\DateTime::__construct",
                 "Psl\\File\\ReadHandle::__construct",
                 "Psl\\File\\WriteHandle::__construct",
                 "Psl\\File\\ReadWriteHandle::__construct",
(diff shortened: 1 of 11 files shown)


config/infection.json.dist
```diff
@@ -32,7 +32,8 @@
         },
         "DecrementInteger": {
             "ignore": [
-                "Psl\\DataStructure\\PriorityQueue::peek"
+                "Psl\\DataStructure\\PriorityQueue::peek",
+                "Psl\\DateTime\\TemporalConvenienceMethodsTrait::since"
```
veewee, line 36: -> Altering since() causes timeouts in the async component.

config/infection.json.dist
```diff
@@ -58,6 +60,7 @@
         },
         "Throw_": {
             "ignore": [
+                "Psl\\DateTime\\DateTime::__construct",
```
veewee, line 63: -> It's impossible to test the InvalidArgumentExceptions that are being thrown. I'm wondering if it would make more sense to use the same checks from `fromParts` in the constructor directly.

The constructor does a range check, the fromParts does a calendar (intl) check. 
Both result in different type of exceptions with different messages.

Would it make sense to always go for the fromParts() checks instead?

src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
```diff
@@ -616,10 +616,9 @@ public function toString(null|DateStyle $date_style = null, null|TimeStyle $time
         $timestamp = $this->getTimestamp();
 
         /**
-         * @psalm-suppress InvalidOperand
          * @psalm-suppress ImpureMethodCall
          */
         return Internal\create_intl_date_formatter($date_style, $time_style, null, $timezone ?? $this->getTimezone(), $locale)
-            ->format($timestamp->getSeconds() + ($timestamp->getNanoseconds() / NANOSECONDS_PER_SECOND));
+            ->format($timestamp->getSeconds());
```
veewee, line 622: the default pattern does not represent nanoseconds.
it is not possible (to my knowledge) to get a timestamp to contain nanoseconds > NANOSECONDS_PER_SECOND - 1. 

I did not find a case to cover this calculation

src/Psl/DateTime/Duration.php
```diff
@@ -674,27 +674,24 @@ public function toString(int $max_decimals = 3): string
...
-        /** @var list<array{string, string}> $values */
-        $values = [
-            [((string) $this->hours), 'hour(s)'],
-            [((string) $this->minutes), 'minute(s)'],
-            [$sec_sign . ((string) $sec) . $decimal_part, 'second(s)'],
-        ];
 
-        // $end is the sizeof($values), use static value for better performance.
-        $end = 3;
-        while ($end > 0 && $values[$end - 1][0] === '0') {
```
veewee, line 686: mutating the while's results in timeouts. I decided to go for a couple of simple if statements instead that results in the same behaviour.

PR #490, opened 2024-09-12
feat: introduce search_opt_k_v

simPod:
Currently, there's no way to search iterable using also a key


azjezz:
we should follow existing naming convention ( e.g `Vec\chunk` / `Vec\chunk_with_keys` ).

I say we rename to `Iter\search_with_keys_opt` and also introduce `Iter\search_with_keys` for consistency.

simPod:
adapted

PR #495, opened 2024-09-25
Added different message to distinguish between two cases:

### LINKED ISSUE #494: New v3 path error message should distinguish between "lack of key" and "key equals null"

```php
Type\shape(['foo' => Type\string()], true)->assert([]);
Type\shape(['foo' => Type\string()], true)->assert(['foo' => null]);
```

Both these two evaluations result in a `AssertException` with a `Expected "array{'foo': string}", got "null" at path "foo".` message.

It's misleading as the lack of the key is definitely different beast to the key that is `= null`.

I would prefer two distinct messages (non-native speaker here, so needs polishing):

```
Expected "array{'foo': string}", path "foo" does not exist.
Expected "array{'foo': string}", got "null" at path "foo".
```

**Additional context**

zerkms:
a) the value at the path is of the wrong type
b) the path does not exist

Fixes https://github.com/azjezz/psl/issues/494

**IMPORTANT**: this is still "work-in-progress" and merely a conversation opener with the aim to find out better wording (not sure I like it like that, but it's the best I could think of at the moment).

### DIFF
--- a/src/Psl/Type/Exception/AssertException.php
+++ b/src/Psl/Type/Exception/AssertException.php
@@ -17,25 +17,8 @@ final class AssertException extends Exception
     /**
      * @param list<string> $paths
      */
-    private function __construct(?string $actual, string $expected, array $paths = [], ?Throwable $previous = null)
+    private function __construct(?string $actual, string $expected, string $message, array $paths = [], ?Throwable $previous = null)
     {
-        $first = $previous instanceof Exception ? $previous->getFirstFailingActualType() : $actual;
-
-        if ($first !== null) {
-            $message = Str\format(
-                'Expected "%s", got "%s"%s.',
-                $expected,
-                $first,
-                $paths ? ' at path "' . Str\join($paths, '.') . '"' : '',
-            );
-        } else {
-            $message = Str\format(
-                'Expected "%s", received no value at path "%s".',
-                $expected,
-                Str\join($paths, '.'),
-            );
-        }
-
         parent::__construct(
             $message,
             $actual ?? 'null',
@@ -57,9 +40,18 @@ final class AssertException extends Exception
         ?string $path = null,
         ?Throwable $previous = null
     ): self {
-        $paths = $previous instanceof Exception ? [$path, ...$previous->getPaths()] : [$path];
+        $paths = Vec\filter_nulls($previous instanceof Exception ? [$path, ...$previous->getPaths()] : [$path]);
+        $actual = get_debug_type($value);
+        $first = $previous instanceof Exception ? $previous->getFirstFailingActualType() : $actual;
 
-        return new self(get_debug_type($value), $expected_type, Vec\filter_nulls($paths), $previous);
+        $message = Str\format(
+            'Expected "%s", got "%s"%s.',
+            $expected_type,
+            $first,
+            $paths ? ' at path "' . Str\join($paths, '.') . '"' : '',
+        );
+
+        return new self($actual, $expected_type, $message, $paths, $previous);
     }
 
     public static function withoutValue(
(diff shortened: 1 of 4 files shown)


src/Psl/Type/Exception/AssertException.php
```diff
@@ -17,18 +17,28 @@ final class AssertException extends Exception
     /**
      * @param list<string> $paths
      */
-    private function __construct(string $actual, string $expected, array $paths = [], ?Throwable $previous = null)
+    private function __construct(?string $actual, string $expected, array $paths = [], ?Throwable $previous = null)
     {
         $first = $previous instanceof Exception ? $previous->getFirstFailingActualType() : $actual;
```
veewee: Wouldn't this be `"null"` on nested shapes where one of the underlying keys does not exist? (given that you pass null as a string to the parent exception?

(same comment for CoercionException)

src/Psl/Type/Exception/AssertException.php
```diff
@@ -17,18 +17,28 @@ final class AssertException extends Exception
...
                 'Expected "%s", got "%s"%s.',
                 $expected,
                 $first,
-                $paths ? ' at path "' . Str\join($paths, '.') . '"' : ''
-            ),
-            $actual,
+                $paths ? ' at path "' . Str\join($paths, '.') . '"' : '',
+            );
+        } else {
+            $message = Str\format(
```
veewee: It might make more sense to build the messages from the named constructors instead of performing if/else logic here?

zerkms: Could be, yes, I tried to make as few changes as possible - as the primary goal at this stage is to build some common grounds :-)

src/Psl/Type/Exception/AssertException.php
zerkms: I don't understand what it means, could you expand? If the "outer" shape does not match - no "nested" shapes will be checked, right?

src/Psl/Type/Exception/AssertException.php
zerkms: Okay, done, will keep the history for now, and will squash if the PR is ever ready to be accepted.

src/Psl/Type/Exception/AssertException.php
veewee: For example:

```php
shape([
    'a' => shape([
        'b' => mixed()
    ]),
])->coerce(['a' => []]);
```

This will result in:

```
Could not coerce "null" to type "array{'a': array{'b': mixed}}" at path "a.b"
 -> Previous:  Could not coerce to type "array{'b': mixed}" at path "b" as the value was not passed.
```



> If the "outer" shape does not match - no "nested" shapes will be checked, right?

The types will coerce top-down indeed. But if one of the nested children results in a failure, the exception gets wrapped and enriched with the "path" information. It's that `$first` logic inside the exception constructor which you default to `?? 'null'`.

zerkms: Oh yes, now I see that `CoercionException` and `AssertException` should be monadic. How about now?

Only for `CoercionException` yet, as a proof-of-concept.

And a mandatory reminder: this is still under heavy WIP, I understand it's not a trivial change and I will clean things up during the final git squash.

veewee: I don't think `wrap()`-ing it is the way:

```php
vec(
    shape([
    'a' => shape(['b' => mixed()])
    ])
)->coerce([['a' => []]]);
```

> Could not coerce "null" to type "vec<array{'a': array{'b': mixed}}>" at path "0.a.b"

The parent's exception construction should be more aware of those details from within the child's type exception.
Maybe an alternate suggestion would to make  the `$actual` and `$first` property nullable for those situations where the value is not available and let that decide the message instead?

zerkms:
UPD: I didn't abandon it, a little busy at work, will return when have a little spare time (unless someone wants to join and help me untangle it :-) )

PR #501, opened 2024-12-08
chore: switch to mago

azjezz:
there is over 700 linting issues reported by mago, we will not be fixing those in this PR, but i will make a following PR to address them.


azjezz:
@veewee the ci failure seems to be unrelated to this change, can you confirm? 

I have also updated psalm so that might be the reason.

veewee:
I horizontally scrolled through the code-style changes and am OK with them.

That said, I don't know if all rules we applied in php-cs-fixer and phpcs are configured the same way as in mago. 
On existing code, it's hard to tell everything is applied because it was already that way before.
Not sure if new code will also be changed according the specific rules we've set in the past. Did you base the formatter / linter rules on the existing once inside this repo or are there some of them missing?  How can we make sure this will work as expected on new code as well?

> the ci failure seems to be unrelated to this change, can you confirm?

The CI didn't fail before, so it must be the bump of the psalm dependency in here. 
It's probably best to fix those issues regardless in here or to create a separate psalm bump issue.

azjezz:
> That said, I don't know if all rules we applied in php-cs-fixer and phpcs are configured the same way as in mago.

Yes, but it seems that php-cs-fixer and phpcs both were not able to actually apply some of those rules, for example, line soft limit rule in phpcs was never applied by phpcs, it only complained if a line exceeded that limit, while mago does split the line if is too long, and inlines it if it fits, an example can be seen in [src/Psl/Async/sleep.php](https://github.com/azjezz/psl/pull/501/files#diff-465b0814d2e5ff835cd530abaf0187cf5a9ab079f5e7a61865174e0b319ebadb), where mago inlines the call because it fits into a single line, while both phpcs and php-cs-fixer ignore it.

since phpcs and php-cs-fixer are not able to split lines or inline them correctly, we previously had to set a [super high line limit in phpcs](https://github.com/azjezz/psl/blob/4a9a6698d98a7dcbd679be8100d6935ff45e622b/config/.phpcs.xml#L42) to avoid having to fix the lines manually.

i personally think mago is doing a better job at applying psr-12 than both php-cs-fixer and phpcs.

veewee:
@azjezz 

I'm sure mago is doing a good job regarding the rules it knows.

My remark was more about the rules that are not in PSR-12.

An non exhaustive list of examples:

* sorting namespaces
* adding strict_types declaration
* using static:: in unit tests
* Forcing final or abstract on all classes.
* Forcing list syntax
* ordered interfaces
* static lambdas
* ....

(It's basically the reason why we've choosen to use both php-cs-fixer and phpcs : to force multiple rules that the other tool does not know)

azjezz:
well, that's the goal of dogfeeding :D to improve it by using it on a real world project.

as for the rule you listed.


- sorting namespaces: not yet
- adding strict_types declaration: yes, linter
- using static:: in unit tests: not yet
- Forcing final or abstract on all classes: neither phpcs or php-cs-fixer support this, but we can make it possible in mago.
- Forcing list syntax: yes, linter
- ordered interfaces: in what context?
- static lambdas: not yet.

I'll make sure to add all the missing rules you mentioned in the next couple of days before merging this.

veewee:
> as for the rule you listed.
> 
>     * Forcing final or abstract on all classes: neither phpcs or php-cs-fixer support this, but we can make it possible in mago.

php-cs-fixer has https://cs.symfony.com/doc/rules/class_notation/final_class.html

>     * ordered interfaces: in what context?

php-cs-fixer uses https://cs.symfony.com/doc/rules/class_notation/ordered_interfaces.html


> 
> I'll make sure to add all the missing rules you mentioned in the next couple of days before merging this.

As mentioned, it is a non exhaustive list. If we want to fully remove both phpcs and php-cs-fixer, we should cover all the rules that we currently apply in both of those tools or agree on not using specific rules anymore.

azjezz:
> php-cs-fixer has https://cs.symfony.com/doc/rules/class_notation/final_class.html

we had this disabled because we couldn't make it work in Psl sadly.

> php-cs-fixer uses https://cs.symfony.com/doc/rules/class_notation/ordered_interfaces.html

this is easy to add, but not sure if we want to.

---

> As mentioned, it is a non exhaustive list. If we want to fully remove both phpcs and php-cs-fixer, we should cover all the rules that we currently apply in both of those tools or agree on not using specific rules anymore.

I don't see the need for us to do that, as long as the code is psr-12 compatiable, which is what we had set those two tools to do.

other rules can be implemented one-by-one, and that is the goal of having mago here, to act as an example+test case.

PR #503, opened 2025-01-20
chore: update mago configuration to match latest release

azjezz:


mago.toml
```diff
@@ -1,15 +1,43 @@
...
+name = "maintainability/excessive-parameter-list"
+threshold = 9
+
+[[linter.rules]]
+name = "maintainability/too-many-enum-cases"
+threshold = 800
+
+[[linter.rules]]
+name = "maintainability/too-many-methods"
+threshold = 43
```
azjezz, line 40: everything here is configured to ensure Psl code as is right now passes, there is a lot of places where it would make sense to lower complexity, but i don't think its worth doing right now.

veewee (approved):

Issue #502, opened 2025-01-13
`MutableMap` and similar collections are not guaranteeing the type of the collection


devnix:
**Describe the bug**
Suppose you define a map that only allows `string` as key and set a string that PHP would cast to an integer as an array key (any integer-ish string that doesn't start with zero). In that case, it will return an `integer` instead of the promised type by the static analysis.

**To Reproduce**
```php
/** @var MutableMap<string, string> $map */
$map = new MutableMap([]);

$map->add('foo', 'foo');
$map->add('123', '123');

/** @psalm-trace $values */
$values = $map->keys()->toArray(); // Trace: $values: list<string>
var_dump($values); // array(2) {[0]=>  string(3) "foo"  [1]=>  int(123)}
```
**Expected behavior**
I would expect this output:
```php
var_dump($values); // array(2) {[0]=>  string(3) "foo"  [1]=>  string(3) "123"}
```

I wonder if you think that this should be implemented differently to avoid such bugs as I do. I prefer a slightly slower/less memory-efficient but correct, runtime-safe implementation.

azjezz:
Fixing this would mean a huge performance penalty, and a BC break that requires a major release.

I personally do not use collections often, so i'm not really sure.

@veewee wdyt?

azjezz:
cc @tkquizlet

devnix:
From a static analysis POV is not a BC but a bugfix. It even breaks when you do `$map->at('123')`.

I can try a couple of test implementations and benchmark them 😊

veewee:
I must agree that this behaviour has bitten me in production as well on regular arrays and if we can fix it, that would be nice.

However I do agree that this is a BC break because people might get unexpected exceptions on code that worked just fine previously in a minor version. Also the performance consideration might be a blocker for people heavily using the maps: we should avoid that this basic building block becomes slow and therefore unusable.

Is there some smart way we could enforce this without too much of an overhead of actually parsing the types?
I was thinking in lines of storing the keys as a set and using the index as the value pointer.
But then I realized this problem also exists in sets ... :)

devnix:
I was not thinking about parsing types at all, but to maintain two inner lists, one for the key and another for the value. I think it should not be such a traumatic overhead

veewee:
That would mean you have to iterate internally to find the value for a key, which is not very performant and not what you'dd expect from a map.

devnix:
The alternative would be to use a non-cryptographic hashing algorithm and do the hash table in userland. 

[...]

PR #504, opened 2025-01-20
chore: benchmarks for `Collection` component

devnix:
Initial PR for https://github.com/azjezz/psl/issues/502 before changing the implementation.

It's the first time I use PHPBench, so I would love any early feedback while I try to give coverage to all the component 😄


azjezz:
@devnix please do a `git reset --hard origin/next` and cherry-pick your commits on top then `git push -f` to have cleaner history.

PR #505, opened 2025-01-21
Add IS0-8601 and RFC-3339 pattern without microsecond fraction

veewee:
In both ISO-8601 and RFC-3339, the microseconds part (decimal fraction) is optional.

* https://www.ietf.org/rfc/rfc3339.txt - See 5.6
* https://www.ionos.com/digitalguide/websites/web-development/iso-8601/ - See basic principles.

In this PR, I introduce 2 more options for these patterns without microseconds.
This makes it interopable with the default PHP date formats (atom, iso-8601 and expanded, DATE_RFC3339 and expanded)

The ISO standard goes even further and makes seconds optional, 2-year digits possible, ... 
That is out of scope for now.

### DIFF
--- a/src/Psl/DateTime/FormatPattern.php
+++ b/src/Psl/DateTime/FormatPattern.php
@@ -16,6 +16,7 @@ enum FormatPattern: string implements DefaultInterface
 {
     case Rfc2822 = 'EEE, dd MMM yyyy HH:mm:ss Z';
     case Iso8601 = 'yyyy-MM-dd\'T\'HH:mm:ss.SSSXXX';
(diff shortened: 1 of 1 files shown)


mago.toml
```diff
@@ -7,7 +7,6 @@ excludes = []
 null_type_hint = "pipe"
 
 [linter]
-level = "Error"
```
veewee: > ERROR Failed to build the configuration: unknown field `level`, expected one of `default_plugins`, `plugins`, `rules` for key `linter`

azjezz (dismissed):
LGTM, needs rebase.

PR #506, opened 2025-01-27
Update psalm, infection, mago, ...

veewee:
* psalm v6
* infection 0.29
* mago through composer

### DIFF
--- a/.github/workflows/coding-standards.yml
+++ b/.github/workflows/coding-standards.yml
@@ -24,11 +24,8 @@ jobs:
         run: |
           make install -j10 -O
 
-      - name: "install mago"
-        run: wget -qO- https://carthage.software/mago.sh | bash -s -- --install-dir="/usr/local/bin"
-
       - name: "formatting"
-        run: mago fmt --dry-run
+        run: ./vendor/bin/mago fmt --dry-run
 
       - name: "linting"
-        run: mago lint --reporting-format=github
+        run: ./vendor/bin/mago lint --reporting-format=github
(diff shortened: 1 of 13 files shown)


src/Psl/DateTime/DateTimeConvenienceMethodsTrait.php
```diff
@@ -346,7 +347,7 @@ public function getISOWeekNumber(): array
      */
     public function getWeekday(): Weekday
     {
-        return Weekday::from((int) $this->format(pattern: 'e'));
+        return Weekday::from((int) $this->format(pattern: 'e', locale: Locale::EnglishUnitedKingdom));
```
veewee, line 350: Turns out these expected formats might give different results for different locales.
Hardcoded the en_GB way as it was used in the tests and is considered correct for these cases.

.github/workflows/coding-standards.yml
```diff
@@ -24,11 +24,8 @@ jobs:
         run: |
           make install -j10 -O
 
-      - name: "install mago"
-        run: wget -qO- https://carthage.software/mago.sh | bash -s -- --install-dir="/usr/local/bin"
```
veewee, line 28: Small downside : since make install takes the locked mago, one needs to bump the dependencies in order te run on newer mago version. It makes the pipeline more stable though ... I'll leave this up to you to decide how to best deal with this: I'dd say a periodic update through PRs is probably the best in terms of compatibility.

azjezz:
LGTM, ship it 🚢

PR #510, opened 2025-02-26
feat(type): add `int_range` type

gsteel:
Ref #390 

Unfortunately, Psalm does not appear to support templates in int ranges, i.e. `int<TMin, TMax>`

What's the best thing to do here? reduce the templates to `int` and do something in the psalm plugin?


gsteel:
Added a rather hopeful feature request here too: https://github.com/vimeo/psalm/issues/11343

azjezz:
Could you please update the annotations to make psalm happy? either way psalm won't understand it. we can add proper templates later once psalm adds support for this.

rebase is needed as well.

gsteel:
@azjezz Looks like the windows test failure is un-related

PR #512, opened 2025-03-01
chore(collection): use a more natural JSON serialization for Sets and Maps

joshrai:
This adjusts the methods `Set::jsonSerialize()` and `MutableSet::jsonSerialize()` to return list-like arrays that `json_encode()` (with default flags) will encode as a JSON array rather than as a JSON object. This is arguably a more natural representation of a set than the classes' internal representations, which have keys that equal their corresponding values. If nothing else, it's more in line with the serialization done in Hack for the `keyset` type, and it's less redundant.

For example:

```php
$s = new Psl\Collection\Set(['foo', 'bar']);
$serialized = $s->jsonSerialize();

var_export($serialized);
echo PHP_EOL;
echo 'JSON: '.json_encode($serialized);
```

Formerly, this would have output the following:

```
array (
  'foo' => 'foo',
  'bar' => 'bar',
)
JSON: {"foo":"foo","bar":"bar"}
```

But now it will output this:

```
array (
  0 => 'foo',
  1 => 'bar',
)
JSON: ["foo","bar"]
```

### DIFF
--- a/docs/component/collection.md
+++ b/docs/component/collection.md
@@ -28,10 +28,10 @@
 #### `Classes`
 
 - [Map](./../../src/Psl/Collection/Map.php#L25)
-- [MutableMap](./../../src/Psl/Collection/MutableMap.php#L25)
-- [MutableSet](./../../src/Psl/Collection/MutableSet.php#L22)
+- [MutableMap](./../../src/Psl/Collection/MutableMap.php#L26)
+- [MutableSet](./../../src/Psl/Collection/MutableSet.php#L23)
 - [MutableVector](./../../src/Psl/Collection/MutableVector.php#L24)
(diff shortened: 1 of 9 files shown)


azjezz:
@veewee i'm merging this into `next`, but not sure if we should consider this to be a BC break or not, we can either leave `next` to be the next minor release of `3.x` or mark it as `4.x-dev` now. ( branches for all 3.x minors have already been created ).

@joshrai i updated the PR to also include similar changes for `Map` and `MutableMap` as we aslo need them to always serialize as objects.

veewee:
@azjezz it's a bc break: if people store the  JSON and rely on encoding/decoding it, this will be breaking for them.

PR #513, opened 2025-03-03
feat(type): add container type

azjezz:
This PR introduces a new `container` type, which is quite similar to `iterable`, however, unlike the `iterable` type, this type does not allow non-array-key keys, and always returns an `array` instead of `Psl\Iter\Iterator`.


veewee:
Maybe this info should be added to the docs as well?

PR #525, opened 2025-05-24
feat: make immutable collections array accessible

erikgaal:
This PR adds support of array accessing to the immutable `Psl\Collection` objects.

It does this by hoisting the `ArrayAcess` interface from the `MutableAccessibleCollectionInterface` to the `AccessibleCollectionInterface` contract.

To ensure the immutability of the collections, the immutable versions will still throw an `Error` when setting or unsetting elements via array access.

### DIFF
--- a/docs/component/collection.md
+++ b/docs/component/collection.md
@@ -12,11 +12,11 @@
 
 #### `Interfaces`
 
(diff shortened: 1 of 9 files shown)


erikgaal:
@maintainer, any clue what's up with the crashing Infection run?

PR #527, opened 2025-09-02
chore: migrate from `psalm` to `mago`

azjezz:
This PR:

1. updates mago to the latest version
2. re-formats the codebase using the latest mago
3. fixing all linting problems reported by mago
4. migrates from using `psalm` for static analysis to using `mago`
5. removes `@psalm-suppress` comments from the code
6. fixes few type bugs detected by mago

`inflection` does not support mago currently, so we are unable to integrate it there :/


src/Psl/Str/replace_ci.php
```diff
@@ -19,20 +19,28 @@
  */
 function replace_ci(string $haystack, string $needle, string $replacement, Encoding $encoding = Encoding::Utf8): string
 {
-    if ('' === $needle || null === search_ci($haystack, $needle, 0, $encoding)) {
+    if ('' === $haystack || '' === $needle || null === namespace\search_ci($haystack, $needle, 0, $encoding)) {
         return $haystack;
     }
```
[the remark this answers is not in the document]
azjezz: PHP `||` will short circuit here, the function won't be called if the haystack is empty.

src/Psl/Json/encode.php
```diff
@@ -33,10 +33,13 @@ function encode(mixed $value, bool $pretty = false, int $flags = 0): string
...
 
     try {
-        /** @var non-empty-string $json */
         $json = json_encode($value, $flags);
     } catch (JsonException $e) {
-        throw new Exception\EncodeException(Str\format('%s.', $e->getMessage()), $e->getCode(), $e);
+        throw new Exception\EncodeException(Str\format('%s.', $e->getMessage()), (int) $e->getCode(), $e);
+    }
+
+    if ($json === false || $json === '') {
```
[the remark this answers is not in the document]
azjezz: No, it won't return an empty string. It would return `"\"\""`

Makefile
```diff
@@ -21,22 +24,11 @@ create-benchmark-reference:
 compare-benchmark-to-reference:                                                 ## run benchmarks, compare result to the "reference" run
 	./vendor/bin/phpbench run --config config/phpbench.json --ref=benchmark_reference
 
-static-analysis:                                                                ## run static analysis checks
-	./vendor/bin/psalm -c config/psalm.xml --show-info=true --no-cache
-	./vendor/bin/psalm -c config/psalm.xml tests/static-analysis --no-cache
```
veewee, line 26: Shouldn't we still run psalm on tests/static-analysis ? Given it checks the psalm plugin?
(not sure if mago already has the PSL specific type enhancements?)

.github/workflows/security-analysis.yml
```diff
@@ -1,28 +0,0 @@
...
-          ini-values: memory_limit=-1
-          tools: composer:v2, cs2pr
-          extensions: bcmath, mbstring, intl, sodium, json
-
-      - name: "installing dependencies"
-        run: |
-          make install -j10 -O
-
-      - name: "running security analysis ( psalm )"
-        run: make security-analysis
```
veewee, line 28: does mago have taint analysis as well?

config/infection.json.dist
```diff
@@ -21,8 +21,8 @@
         }
     },
     "timeout": 10,
-    "minMsi": 100,
-    "minCoveredMsi": 100,
+    "minMsi": 97,
```
veewee, line 24: would it make sense to keep it at 100 and add ignores instead?
Procentual coverage is kinda strange :)

config/mago.toml
```diff
@@ -1,65 +1,34 @@
...
+class-name = { psr = false }
+no-global = { enabled = false }
+literal-named-argument = { enabled = false }
+no-boolean-flag-parameter = { enabled = false }
+no-boolean-literal-comparison = { enabled = false }
+no-error-control-operator = { enabled = false }
+
+[analyzer]
+excludes = ["tests"]
+ignore = ["possibly-static-access-on-interface"]
```
veewee, line 34: should we add tests/static-analysis?

Makefile
azjezz: I think those tests should be in the plugin repo, not here.

.github/workflows/security-analysis.yml
azjezz: No, but we don't actually need it in Psl 🤷🏼‍♂️

config/infection.json.dist
azjezz: Nuh, i think it's better to lower it, and fix those issues later.

The problem is that infection no longer uses psalms to know if it should skip a mutation due to failing SA.

PR #546, opened 2025-10-23
feat: add Tree component for hierarchical data structures

azjezz:
Introduces a new `Tree` component providing immutable tree data structures and functional operations for hierarchical data manipulation.

Architecture:
- NodeInterface<T>: Base interface with getValue()
- LeafNode<T>: Immutable leaf nodes (no children)
- TreeNode<T>: Immutable tree nodes with getChildren()
- Pure constructor functions: tree() and leaf()

Features:
- Transformation operations: map, filter, reduce, fold
- Traversal algorithms: pre_order, post_order, level_order
- Search functions: find, any, all, contains
- Utility functions: count, depth, is_leaf, leaves
- Path operations: path_to, at_path
- Array conversion: from_array, to_array


src/Psl/Tree/at_path.php
```diff
@@ -0,0 +1,57 @@
...
+ * @template T
+ *
+ * @param NodeInterface<T>  $node
+ * @param list<int<0, max>> $path
+ *
+ * @return T|null null if the path is invalid
+ *
+ * @pure
+ */
+function at_path(NodeInterface $node, array $path): mixed
```
veewee: this would rather be `at_index` since it is not compatible with `to_path`
Maybe also introduce a `to_index` ?

src/Psl/Tree/from_array.php
```diff
@@ -0,0 +1,68 @@
...
+ *         }>
+ *       }>
+ *     }>
+ *   }>
+ * } $array
+ *
+ * @return TreeNode<T>
+ *
+ * @pure
+ */
```
veewee, line 44: What would be a nice addition is something like a `from_(vec|list)` with a grouping function.
For example to create a tree based on db records that are linked by using a parent_id. (if the parent_id = null, you have to root and can start filling up from there)

wdyt?

src/Psl/Tree/map.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @template T
+ * @template Tu
+ *
+ * @param NodeInterface<T> $node
+ * @param (Closure(T): Tu) $function
+ *
+ * @return NodeInterface<Tu>
+ */
+function map(NodeInterface $node, Closure $function): NodeInterface
```
veewee, line 31: Maybe another addition besides mapping, would be able to traverse / convert a tree in order to serialize it in the format you want?
For example:

```php
<?php
 Tree\traverse(
     Tree\tree(['id' => 1, 'label' => 'A'],
         [
             Tree\leaf(['id' => 2, 'label' => 'B']),
             Tree\leaf(['id' => 3, 'label' => 'C'])
         ]),
     fn($value, $traverse) => ['id' => $value['id'], 'label' => $value['label'], 'customChildrenProp' => $traverse()]
 );
```
```
['id' => 1, 'label' => 'A', 'customChildrenProp' => [
        ['id' => 2, 'label' => 'B', 'customChildrenProp' => []],
        ['id' => 3, 'label' => 'C', 'customChildrenProp' => []]
    ]]
```

Maybe the traverse name is more known in combination with 'visitor' functions to perform an operation on a node. That visitor could result in for example: add sibling below, remove node, remove parent, ... when a condition is met. (maybe a bridge to far right ?! :D)

veewee (commented):
Looks nice and will defenitely use this!

PR #547, opened 2025-10-23
feat: add Graph component with directed and undirected graph support

azjezz:
Add comprehensive Graph component for with immutable graph data structures and algorithms for working with directed and undirected graphs.

Features:
- DirectedGraph and UndirectedGraph classes with adjacency list representation
- Pure constructor functions: directed() and undirected()
- Weighted and unweighted edge support
- Graph operations: add_node(), add_edge(), nodes(), neighbors()
- Traversal algorithms: BFS (breadth-first search), DFS (depth-first search)
- Path algorithms: shortest_path(), shortest_path_by(), has_path()
- Topological algorithms: topological_sort(), has_cycle()


src/Psl/Graph/UndirectedGraph.php
```diff
@@ -0,0 +1,111 @@
...
+ * Immutable undirected graph using adjacency list representation.
+ *
+ * In an undirected graph, edges are bidirectional.
+ *
+ * Supports any node type (scalars, objects, arrays, resources, etc.).
+ *
+ * @template TNode
+ * @template TWeight
+ */
+final readonly class UndirectedGraph
```
veewee: would it make sense to have a common interface?

src/Psl/Graph/has_cycle.php
```diff
@@ -0,0 +1,120 @@
...
+ * @template TNode
+ * @template TWeight
+ *
+ * @param DirectedGraph<TNode, TWeight>|UndirectedGraph<TNode, TWeight> $graph
+ *
+ * @pure
+ */
+function has_cycle(DirectedGraph|UndirectedGraph $graph): bool
+{
+    if ($graph instanceof DirectedGraph) {
```
veewee: wouldnt it make sense to have this logic on the graph direclty, since those are 2 different implementations and just use this as a shortcut function here?

src/Psl/Graph/shortest_path_by.php
```diff
@@ -0,0 +1,175 @@
...
+ * @return list<TNode>|null
+ *
+ * @pure
+ */
+function shortest_path_by(
+    DirectedGraph|UndirectedGraph $graph,
+    mixed $from,
+    mixed $to,
+    Closure $weight_converter,
+): null|array {
```
veewee, line 51: would option be a better idea for places with null returns?

azjezz: we dont use options currently anywhere in psl. it is better, but, more complex, and inconsistent with the rest of library. unless we change nullable -> option everywhere else.

PR #549, opened 2025-10-24
Add support for PHP 8.5

veewee:


azjezz:
we need to fix deprecations, and maybe bump revolt too

veewee:
Revolt seems to have an open PHP 85 support PR. Not sure how good it is but I dont know the package either.
I've put this one on draft and will pick it up further soon.

veewee:
This one should be good to go.
I've kept infection on PHP 8.4 since it breaks on the revolt deprecations.

The only deprecation warnings left are the once from revolt, which have been fixed in https://github.com/revoltphp/event-loop/pull/110. Which hasn't been officially tagged yet.

src/Psl/Async/Internal/State.php
```diff
@@ -69,7 +69,8 @@ public function __destruct()
      */
     public function subscribe(Closure $callback): string
     {
-        $id = self::$nextId++;
+        $id = self::$nextId;
+        self::$nextId = str_increment(self::$nextId);
```
azjezz, line 75: missing import.

src/Psl/Str/Byte/ord.php
```diff
@@ -13,5 +13,5 @@
  */
 function ord(string $character): int
 {
-    return \ord($character);
+    return \ord($character[0]);
```
azjezz: wouldn't this fail on empty strings?

azjezz (changes requested):

azjezz (approved):

PR #552, opened 2025-11-09
fix(iter): do not narrow down seek($offset) type

azjezz:


src/Psl/Iter/Iterator.php
```diff
@@ -164,13 +165,15 @@ public function rewind(): void
     /**
      * Seek to the given position.
      *
-     * @param int<0, max> $offset
```
azjezz, line 167: detected by the upcoming version of mago.

we can't narrow down parameter types.

PR #557, opened 2025-12-01
feat(option,result): casting between `Option` and `Result`

devnix:


src/Psl/Result/Failure.php
```diff
@@ -171,4 +172,9 @@ public function always(Closure $always): ResultInterface
             throw $this->throwable;
         });
     }
+
+    public function toOption(): Psl\Option\Option
```
azjezz, line 176:
```suggestion
    public function ok(): Psl\Option\Option
```

azjezz (changes requested):
I think the `Result` to `Option` makes sense, as it would match Rust's [`Result::ok`](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok).

However, i don't think we should do `Option` -> `Result`, because the Psl exception thrown is definitly not the error the user wants to give out.

src/Psl/Result/Failure.php
devnix: Excellent!

Regarding this:

> However, i don't think we should do Option -> Result, because the Psl exception thrown is definitly not the error the user wants to give out.

There's indeed some functions in Rust to transform Option into Result:
https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or
https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else

would you be ok if I write these and remove the current `Option -> Result`?

azjezz: Sure, we can add `Option::okOr(Exception)` and `Option::okOrElse(fn() => Exception)`, but we shouldn't use the psl exception like the current implementation.

PR #567, opened 2026-01-29
chore: update dependencies

azjezz:


src/Psl/Shell/execute.php
```diff
@@ -85,7 +85,6 @@ static function (array $m) use (&$environment, &$variable_cache, &$variable_coun
                 }
 
                 if (isset($variable_cache[$m[0]])) {
-                    /** @var string */
```
azjezz, line 88: mago got smarter! this is not needed anymore

PR #569, opened 2026-02-06
chore: prepare for the next version of mago

azjezz:


azjezz:
Ignoring CI failure, its a mago bug, fixed in the upcoming version.

PR #571, opened 2026-02-17
Fix PHPDoc return type annotation

mitelg:
hey there :wave: 

This fixes a typo in the `@return` annotation. It is only a small change, but PHPStorm gathers all annotations from the whole project and puts them in the auto-complete suggestions, which is a bit annoying for this one :sweat_smile:


azjezz:
Thanks @mitelg !

mitelg:
thanks for merging that fast! :rocket:

PR #572, opened 2026-02-18
feat: introduce Either

### LINKED ISSUE #355: Result is too restrictive, introduce Either type?
I want to type a function/method representing action which may fail, but not from pure technical perspective, rather from the business logic one. For instance, "register new user" action may succeed, or may fail for reasons like "username is already taken" or "password is too short" etc. I don't want to split this action into two steps, a validation method and then the actual action method, because I don't like creating an intermediate state which is invalid (however others may want to do it in two steps — it's just style preference, I made this choice).

This lib has a Result type which looks like it's exactly what I need, but it's unfortunately not: its Failure branch is limited only to exceptions (more correctly, throwables). However this is not how I view the business logic: for me exceptions are, well, exceptional situations: e.g. "database connection failed" or "file could not be read", "remote API returned HTTP 500" etc. Purely technical ones. Or could be business-logic related, but unexpected, e.g. some entity's invariant was violated due to some logic error in the code, or maybe someone edited its state directly in the database etc. This is the other thing though, i don't want any exception-related fluff here: traces, codes etc. Such action "failures" are no more than just normal operation of my system.

Ideally I'd like to extend Result failure type to accept not only Exceptions. This is not possible due to possible backwards-compatibility break. Instead, a new type can be introduced: `Either`, with `Left` and `Right` variants as inspired by Haskell.

I have my own library for this, https://github.com/someniatko/php-result-type, which was inspired by this one: https://github.com/GrahamCampbell/Result-Type, but it has more robust Psalm typing.

### LINKED ISSUE #440: [RFC] Change `ResultInterface<T>` to `ResultInterface<T, E>`
I've been chatting with @veewee on Discord about adding a template to `ResultInterface` to return what kind of error it could contain.

I'm not sure if this could be considered a breaking change, as it only affects static analysis and can be easily ignored, baselined, or fixed in a moment by going to each error and changing every `@return ResultInterface<Foo>` to `@return ResultInterface<Foo, Throwable>` if you don't feel like studying the specific throwables you want to return from there.

Another option apart from waiting for a new major version or just letting static analysis break is to wait for some default generic type syntax, see:
 - https://github.com/phpstan/phpdoc-parser/pull/148
 - https://github.com/phpstan/phpstan/issues/4801
 - https://github.com/phpstan/phpstan-src/pull/1835

This could allow us to have a syntax that would behave the same way but without having any new errors:
```php
/**
 * @template T
 * @template E of Throwable = Throwable
 */
interface ResultInterface {}
```

If we are OK with baselining errors (regardless of whether the change is in a minor or major version), the following syntax would infer the template with `Throwable` as the default value:
```php
/**
 * @template T
 * @template E of Throwable
 */
interface ResultInterface {}
```
Psalm would only throw a [`MissingTemplateParalm`](https://psalm.dev/r/891520eca6). At the same time, PHPStan would report [two errors](https://phpstan.org/r/280792e3-a574-401f-ab1a-accad696ca81). Still, the return would be implicitly inferred from `ResultInterface<int> to ResultInterface<Throwable>` in the `not_documented()` case on both major static analysis tools.

---

Apart from this, there is another feature that would be nice to have in the static analysis, and that is to be able to infer the type of `@throws` with a generic, so we could type this:
```php
/**
 * @template T
 * @template E of Throwable = Throwable
 */
interface ResultInterface {
    /**
     * @return T
     * 
     * @throws E
     */
    public function getResult();
}
```
- Psalm:
  - [`UndefinedDocblockClass`](https://psalm.dev/r/e3257c26e9)
  - https://github.com/vimeo/psalm/issues/6098
- PHPStan
  - [No errors, but when using checked exceptions it will not evaluate `E` at `@throws`](https://phpstan.org/r/4dd65582-3f67-46f4-adbe-a6aed7c5a70a)
  - https://github.com/phpstan/phpstan/issues/6043

simPod:
Resolves https://github.com/azjezz/psl/issues/355, Resolves https://github.com/azjezz/psl/issues/440


azjezz:
Thanks @simPod !

PR #622, opened 2026-03-14
feat(async): introduce cancellation tokens, replacing Duration-based timeouts across all IO, networking, channel, and async suspension points

azjezz:


azjezz:
@veewee up to you if we merge this and release 6.0.0 :)

PR #638, opened 2026-03-15
bc(socks): introduce configuration object

azjezz:


evapatel1654 (commented):
task bc(socks): introduce configuration object is still pending

PR #692, opened 2026-03-19
feat: introduce `Compression` package

azjezz:


composer.json
```diff
@@ -24,7 +24,8 @@
         "infection/infection": "^0.32.6",
         "php-coveralls/php-coveralls": "^2.9.1",
         "phpbench/phpbench": "^1.5.1",
-        "phpunit/phpunit": "^13.0.5"
+        "phpunit/phpunit": "^13.0.5",
+        "ext-brotli": "*"
```
azjezz: @veewee - just dev dependency, used for testing compression with a real implementation :) but will be a hard deps in the future when we add `Psl\Brotli`.

veewee: I'm afraid adding this extension as a hard dependency will make installation of the main php-standard-library package quite hard - for a rather niche component. I assume mostly applications will want to have the full stuck minus the exotic extension parts. Maybe time to set-up some "composer pack" system at that point as well?

azjezz: yea, hesitant about adding brotli, and honestly, i'm not sure if its right.

The problem, if we want to support http server/client, we need brotli and other (e.g zstd ) compressions, i will thinkg more about how we can introduce them, there is the optional extension dependency option, but not sure about it... we will see, but not adding it now.

As for packs, i think yes, i will see how i can set them up, either updating the splitter to also handle it, or maybe manual setup.

PR #724, opened 2026-03-23
feat: add `bin/psl detect-packages` command

veewee:
## Summary

- Adds a `bin/psl` CLI tool shipped with the monorepo, with a `detect-packages` command that scans PHP source directories and reports which `php-standard-library/*` standalone packages are used.
- Supports `--dev` flag to separate `require` from `require-dev` dependencies — packages found in both are placed in `require` only.
- Supports `--composer` flag to output a JSON snippet ready to paste into `composer.json`, with version constraints derived from the installed package's branch alias.
- Centralizes the namespace-to-package map (previously inlined in the splitter's `verify.php`) into `Psl\Internal\package_namespace_map()` so both the splitter and the new binary share a single source of truth.

### Why this lives in the monorepo

This tool exists to help users **migrate away from the monorepo** (`php-standard-library/php-standard-library`) to individual standalone packages. It needs access to the full namespace-to-package map and uses several PSL packages internally (`File`, `Filesystem`, `Regex`, `Json`, `IO`, `Str`, `Vec`, `Type`). Placing it in `foundation` would force foundation to depend on all of those packages, bloating a package that currently only requires PHP itself. Since the tool's purpose is to facilitate the mono → packages transition, it naturally belongs in the monorepo — once users have migrated, they no longer need it.

### Usage

```bash
# Scan source directories
vendor/bin/psl detect-packages src/

# Separate require and require-dev
vendor/bin/psl detect-packages src/ --dev tests/

# Output JSON for composer.json
vendor/bin/psl detect-packages src/ --dev tests/ --composer
```

**Human-readable output:**

```
Detected PSL packages:

  require:
    php-standard-library/dict
    php-standard-library/file
    php-standard-library/foundation
    php-standard-library/fun
    php-standard-library/iter
    php-standard-library/math
    php-standard-library/regex
    php-standard-library/result
    php-standard-library/str
    php-standard-library/type
    php-standard-library/vec

  require-dev:
    php-standard-library/collection
    php-standard-library/filesystem
    php-standard-library/os

Run with --composer to get a JSON snippet for composer.json.
```

**JSON output (`--composer`):**

```json
{
    "require": {
        "php-standard-library/dict": "^6.1",
        "php-standard-library/file": "^6.1",
        "php-standard-library/foundation": "^6.1",
        "php-standard-library/fun": "^6.1",
        "php-standard-library/iter": "^6.1",
        "php-standard-library/math": "^6.1",
        "php-standard-library/regex": "^6.1",
        "php-standard-library/result": "^6.1",
        "php-standard-library/str": "^6.1",
        "php-standard-library/type": "^6.1",
        "php-standard-library/vec": "^6.1"
    },
    "require-dev": {
        "php-standard-library/collection": "^6.1",
        "php-standard-library/filesystem": "^6.1",
        "php-standard-library/os": "^6.1"
    }
}
```

### DIFF
--- a/bin/psl
+++ b/bin/psl
@@ -106,7 +106,8 @@ function detect_packages_command(array $arguments): never
 
             $content = File\read($entry);
             foreach ($nsToDir as $ns => $dir) {
-                if (Regex\matches($content, '/Psl\\\\' . $ns . '[\\\\\\s;,)]/')) {
+                $escapedNs = Str\replace($ns, '\\', '\\\\');
+                if (Regex\matches($content, '/Psl\\\\' . $escapedNs . '[\\\\\\s;,)]/')) {
                     $found[$dir] = true;
                 }
             }
--- a/bin/psl
+++ b/bin/psl
@@ -3,6 +3,7 @@
 
 declare(strict_types=1);
 
+use Psl\Dict;
 use Psl\Env;
 use Psl\File;
 use Psl\Filesystem;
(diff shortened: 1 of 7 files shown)


bin/psl
```diff
@@ -0,0 +1,167 @@
...
+    IO\write_error_line('Commands:');
+    IO\write_error_line('  detect-packages <path> [--composer]  Detect PSL packages used in a directory');
+    IO\write_error_line('  help                                 Show this help');
+    IO\write_error_line('');
+
+    exit(1);
+}
+
+(static function (): never {
+    $arguments = Vec\slice($GLOBALS['argv'], 1);
```
azjezz:
```suggestion
    $arguments = Vec\slice(Env\args(), 1);
```

:P

docs/generate.php
```diff
@@ -126,6 +126,7 @@
     'terminal' => 'Terminal',
     'security' => 'Security',
     'system' => 'System',
+    'tools' => 'Tools',
     'other' => 'Other',
```
azjezz: maybe have it after `Other`? since `Other` is still a category of components, `Tools` is just DX stuff? idk..

PR #735, opened 2026-03-26
feat(type): support 'true'/'false' string literals in bool coercion

veewee:
## Summary
- Add support for coercing `'true'` and `'false'` string literals to `bool` in `BoolType::coerce()`
- Uses `mb_strtolower` so that `'True'`, `'FALSE'`, etc. are also handled
- Updates `LiteralScalarBoolTypeTest` to reflect the new coercion behavior

## Why
JavaScript uses `'true'`/`'false'` in its URL encoding API, XSD uses these literals for boolean values. This is a very common pattern that warrants built-in coercion support.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### features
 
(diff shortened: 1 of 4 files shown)


azjezz:
`CHANGELOG.md` entry :D

PR #762, opened 2026-04-24
Add EitherOrBoth + Iter\merge_join_by / merge_join_by_key

veewee:
## Motivation

Three-way diff of two collections by key — insert / update / delete events — shows up constantly (DB sync, cache invalidation, UI reconciliation, ETL) and is easy to get wrong by hand. Every major language except PHP exposes some form of it:

| Language / library | API |
|---|---|
| Rust `itertools` | `merge_join_by(a, b, cmp) -> Iterator<EitherOrBoth<L, R>>` |
| Haskell | `Data.These` + `align` |
| Elm | `Dict.merge onLeft onBoth onRight` |
| SQL | `MERGE … WHEN MATCHED / NOT MATCHED` |
| Hack / HSL | `Dict\diff_by_key` (set difference only) — **no three-way primitive** |

Psl can lead where HSL has a gap, reusing `Psl\Either`'s disjoint-union conventions.

### `Psl\EitherOrBoth` (new package)

Three-variant disjoint union: `Left`, `Right`, or `Both`. Unlike `Either`, no side is privileged — the three are equal citizens.

- Interface + three final readonly variant classes + two exceptions
- `left()` / `right()` / `both()` free constructors
- Method surface matches `Psl\Either` (`map`, `mapLeft`, `mapRight`, `proceed`, `apply`, `swap`, `unwrapLeft`, `unwrapRight`, `containsLeft`, `containsRight`), plus `mapAny` for independent per-side transforms and the `isLeft`/`isRight`/`isBoth` + `hasLeft`/`hasRight` predicate split.

### `Psl\Iter\merge_join_by` and `Psl\Iter\merge_join_by_key`

- `merge_join_by(left, right, compare)` — lazy two-cursor merge of sorted inputs, O(1) memory. Comparator returns `Psl\Comparison\Order`.
- `merge_join_by_key(left, right, key_by)` — hash-based variant for keyed inputs that need not be pre-sorted, O(|right|) memory. Psl-only convenience; Rust itertools requires pre-sorted inputs, but PHP rarely has them and hashing is cheap.

Both yield `EitherOrBoth` events — callers compose with `proceed`, `Iter\apply`, `Vec\map`, `Vec\filter`.

## Design decisions worth surfacing

- **No `Comparable`.** Rust's `EitherOrBoth` derives only `PartialEq, Eq, Hash` — no total order, since the three variants have no canonical ordering. `Equable` only.
- **Positional argument order** for `proceed`, `apply`, and `mapAny` (`left, right[, both]`). `Either::proceed` puts `right` first as a happy-path convention; that convention doesn't carry here since all three variants are equal citizens.
- **`apply` takes a single closure** matching `Either::apply`. On `Both` it runs twice, mirroring `map`.
- **`merge_join_by_key` duplicate-key semantics:** last-write-wins on the right (consistent with `Dict\reindex`); left-side duplicates each yield their own event, with only the first pairing consuming the right match.

## `functions.php` vs one-file-per-function

Free constructors live in a single `src/Psl/EitherOrBoth/functions.php` rather than the usual one-function-per-file convention (`Option/some.php`, `Option/none.php`). On case-insensitive filesystems (default macOS APFS, Windows NTFS), lowercase `left.php` collides with the existing PascalCase `Left.php`. `Psl\Option` escapes this because it has no sibling `Some`/`None` class files. Happy to split into per-function files if reviewers prefer the convention and are OK assuming a case-sensitive working tree.

## Open for reviewers: name and vocabulary

Went with **`EitherOrBoth`** (Rust itertools) over **`These`** (Haskell). The variants `Left`/`Right`/`Both` reuse the "side" vocabulary from `Psl\Either`, reading as a natural superset. Haskell's `This`/`That`/`These` is awkward in PHP: `This` collides mentally with `$this`, `That` is nondescript, and the outer type shares its name with one of its variants.

The tradeoff: `Psl\Either\Left` and `Psl\EitherOrBoth\Left` now exist in parallel namespaces — no code collision, but readers need to pay attention in `use` statements. Open to the Haskell naming or a third option (e.g. renaming the variants to `LeftOnly`/`RightOnly`/`Both` for full disambiguation) if reviewers prefer.

Closes relevant use-case gap relative to Hack/HSL.


packages/either-or-both/src/Psl/EitherOrBoth/functions.php
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * @param TRight $value
+ *
+ * @return Right<TRight>
+ *
+ * @pure
+ *
+ * @api
+ */
+function right(mixed $value): Right
```
azjezz, line 51: one per file? :P

packages/iter/src/Psl/Iter/merge_join_by.php
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @param iterable<L> $left
+ * @param iterable<R> $right
+ * @param (Closure(L, R): Order) $compare
+ *
+ * @return iterable<EitherOrBoth\EitherOrBoth<L, R>>
+ *
+ * @api
+ */
+function merge_join_by(iterable $left, iterable $right, Closure $compare): iterable
```
azjezz: should we return `Iter\Iterator`? as its rewindable, as user might try to iterate over the result twice, which won't work as they expect, espically since this function says it returns `iterable` not `Generator`.

If we are keeping generator, the signature must be updated to `Generator` ( most specific )

packages/iter/src/Psl/Iter/merge_join_by_key.php
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @param iterable<T> $left
+ * @param iterable<T> $right
+ * @param (Closure(T): I) $key_by
+ *
+ * @return iterable<EitherOrBoth\EitherOrBoth<T, T>>
+ *
+ * @api
+ */
+function merge_join_by_key(iterable $left, iterable $right, Closure $key_by): iterable
```
azjezz: same as about iterable vs Generator vs Iter\Iterator

packages/either-or-both/src/Psl/EitherOrBoth/functions.php
azjezz: Oh, just saw description for this... i'm pretty sure we are doing this in other places.. wonder why it never hit us there..

veewee: I don't think this is true:

- I couldn't find any case-only filename duplicate within any single directory
- I Inspected each related package:
    - Psl\Either — has Left.php/Right.php, no factory functions (callers do new Left(...))
    - Psl\Option — has some()/none() factories, but no Some.php/None.php class files (factory methods on abstract Option)
    - Psl\Result — has Failure.php/Success.php, no factory functions

azjezz: lets keep it then. its fine.

azjezz (approved):

PR #768, opened 2026-04-28
feat(string): Minor return type refinements

gsteel:
Adds conditional return types to refine to `non-empty-whatever` in `Str\format`, `Str\lowercase` and `Str\uppercase`

### DIFF
--- a/packages/str/src/Psl/Str/lowercase.php
+++ b/packages/str/src/Psl/Str/lowercase.php
@@ -25,7 +25,7 @@ use function mb_strtolower;
  *      Str\lowercase('سيف')
  *      => Str('سيف')
  *
- * @return lowercase-string
+ * @return ($string is non-empty-string ? non-empty-lowercase-string : lowercase-string)
  *
  * @pure
  *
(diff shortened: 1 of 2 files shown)


packages/str/src/Psl/Str/format.php
```diff
@@ -21,7 +21,7 @@
  *
  * @pure
  *
- * @return string a string produced according to the $format string.
```
azjezz, line 24: This is wrong, e.g `format('%s', '')` gives an empty string.

Mago already has a special return type handler for `sprintf`/`format`, so if you are using it, that should just work.

If you are using PHPStan or Psalm, we have plugins for both, and we can improve typing there.

azjezz (changes requested):
`format` change should be reverted. others LGTM.

packages/str/src/Psl/Str/format.php
gsteel: OK, Psalm complaining about `format` is what brought me here 👍

azjezz (approved):

PR #769, opened 2026-05-12
feat(type): Allow null argument to `Type\class_string` to assert or coerce bare `class-string`

gsteel:
I considered adding another type for this, but what would you call it?


azjezz:
Mago issues need to be fixed.

Note: mutation tests will not pass. ignore them.

gsteel:
Thanks @azjezz 

It'd be handy if CI failed on warnings - Can mago be made to exit 1 on warnings? Couldn't see an option in the docs…

azjezz:
@gsteel Mago has `minimum-fail-level` setting for controlling this. we could/should add it later. probably add it to CI using `--minimum-fail-level` arg tho instead of mago.toml, because we use same `mago.toml` in the CI of mago itself to ensure we don't break PSL. and warnings come up there as we improve things ( e.g redundant var docblock due to enhanced type inference ).

azjezz:
@gsteel can you please add a CHANGLOG entry for this?

PR #770, opened 2026-05-12
chore(ci): Fail CI on warning for lint and analyze

gsteel:


.github/workflows/coding-standards.yml
```diff
@@ -29,4 +29,4 @@ jobs:
         run: just fmt-check
 
       - name: "lint"
-        run: just lint
+        run: ./vendor/bin/mago --config config/mago.toml lint --minimum-fail-level warning
```
azjezz: we can just ( pun intended ) add this to `Justfile`.

gsteel: 🤦‍♂️ Gimme a sec!

PR #784, opened 2026-06-01
feat(docs): add permalink anchors to section headings

erikgaal:
## Summary

Section headings on the docs site now render with a clickable permalink anchor, so individual sections can be linked directly via URL fragments.

This enables league/commonmark's `HeadingPermalink` extension (already available via `league/commonmark`), configured minimally:
- `min_heading_level => 2` — the `h1` page title is excluded.
- `insert => after` — the anchor sits after the heading text.

Everything else uses the extension defaults (the `¶` symbol, `content-` id/fragment prefix, `heading-permalink` class).

## Styling

- The anchor is muted and revealed on heading hover / keyboard focus.
- `scroll-margin-top` is added to section headings so fragment navigation doesn't clip the heading under the viewport edge.

## Notes

- Only the rendered HTML changes; the raw `index.md` / `llms*.txt` outputs are generated from the unmodified markdown and stay clean.
- The existing install-box insertion (which keys off `</h1>`) is unaffected, since the `h1` carries no anchor.

### DIFF
--- a/docs/generate.php
+++ b/docs/generate.php
@@ -7,6 +7,7 @@ require_once __DIR__ . '/vendor/autoload.php';
 use League\CommonMark\Environment\Environment;
 use League\CommonMark\Extension\Autolink\AutolinkExtension;
 use League\CommonMark\Extension\CommonMark\CommonMarkCoreExtension;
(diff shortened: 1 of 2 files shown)


veewee:
Thanks

PR #788, opened 2026-06-19
reified generics

henderkes:


packages/async/src/Psl/Async/Awaitable.php
```diff
@@ -25,13 +25,9 @@
...
  * Copyright (c) 2015-2021 Amphp ( https://amphp.org )
  *
- * @template-covariant T
- *
- * @implements PromiseInterface<T>
- *
  * @api
  */
-final readonly class Awaitable implements PromiseInterface
+final readonly class Awaitable<T = mixed> implements PromiseInterface<T>
```
azjezz: Should be marked covariant, and default should be removed

azjezz:
see #789

PR #793, opened 2026-06-25
feat(type): add `numeric` type

### LINKED ISSUE #791: [RFC] Add `Psl\Type\numeric()`
There is currently `Psl\Type\num()` representing `int|float` and `Psl\Type\numeric_string()` for `numeric-string`…

Would the addition of `numeric()` to represent `int|float|numeric-string` be an acceptable addition?

If so, I'd be happy to send in a patch for it :)

It would be a nice shortcut for `union(num(), numeric_string())`.

gsteel:
Closes #791

### DIFF
--- a/packages/type/src/Psl/Type/Internal/NumericType.php
+++ b/packages/type/src/Psl/Type/Internal/NumericType.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Psl\Type\Internal;
+
+use Psl\Type\Exception\AssertException;
+use Psl\Type\Exception\CoercionException;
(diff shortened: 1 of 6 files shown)


packages/type/src/Psl/Type/Internal/NumericType.php
```diff
@@ -0,0 +1,53 @@
...
+ * @extends Type<numeric>
+ *
+ * @internal
+ */
+final readonly class NumericType extends Type
+{
+    public function matches(mixed $value): bool
+    {
+        return is_numeric($value);
+    }
```
azjezz, line 24: Much better than a union!

PR #796, opened 2026-07-22
fix(dict): limit diff values to string-castable values only

azjezz:


packages/dict/src/Psl/Dict/diff.php
```diff
@@ -4,14 +4,16 @@
...
+
 use function array_diff;
 use function array_map;
 
 /**
  * Computes the difference of iterables.
  *
  * @template Tk of array-key
- * @template Tv
+ * @template Tv of scalar|null|resource|Stringable
```
azjezz, line 16: A non-string-castable value would result in a runtime error.

See https://github.com/carthage-software/mago/issues/2116

PR #798, opened 2026-07-24
chore(deps): update mago to the latest version

azjezz:


packages/io/src/Psl/IO/MemoryHandle.php
```diff
@@ -77,7 +77,7 @@ public function tryRead(null|int $maxBytes = null): string
         $length -= $this->offset;
         $length = $length > $maxBytes ? $maxBytes : $length;
         $result = substr($this->buffer, $this->offset, $length);
-        $this->offset = ($offset = $this->offset + $length) >= 0 ? $offset : 0;
+        $this->offset += $length;
```
azjezz, line 80: I thought mago was having a false positive here, it took me like 10 minutes of reading this method to figure out that mago was indeed correct..

PR #801, opened 2026-08-19
fix(docs): Update header formatting in compression.md

94noni:
👋🏻 
on https://php-standard-library.dev/6.2.1/compression/ (and on the left navbar)
the `compression` seems different in the display, perhaps related to this PR fix

cheers!


veewee:
Thanks!

